PR #33, opened 2016-06-29
cmake .. does not refer to chemfiles folder

maxlevesque:
Falls in "obvious" but ... :)


Luthaf:
That is what append when you don't test your examples ... Thanks!

PR #44, opened 2016-10-08
Fix label support in PDB.cpp

pgbarletta:
I forgot to add `trim()` function when reading label.


pgbarletta:
Some spacing was missing in the PDB output, neither Pymol or VMD could read coordinates properly. I also added a default resName "RES" to each atom. 
Checking each atom for its residue seems quite inefficient, but since `Residue` objects are so optional (topologies may completely lack residues, have all its atoms grouped in residues or they may even have only some of its atoms in residues), I couldn't figure an elegant workaround.

I overextended in the initial intention of this pull request, so please let me know if you find these changes appropriate.

Luthaf:
>  Checking each atom for its residue seems quite inefficient, but since Residue objects are so optional (topologies may completely lack residues, have all its atoms grouped in residues or they may even have only some of its atoms in residues), I couldn't figure an elegant workaround.

Checking atom for residue should not be too expensive: it is only a [lookup](https://github.com/chemfiles/chemfiles/blob/40b8daa8e93d081a7bb15fa99029ae8a360fb136/src/Topology.cpp#L110) in a [map](https://github.com/chemfiles/chemfiles/blob/40b8daa8e93d081a7bb15fa99029ae8a360fb136/include/chemfiles/Topology.hpp#L127).

I usually prefer making the software correct first, and fast later. In this case, I do not think the performance penalty is too high, and I prefer having a better PDB output.

> Some spacing was missing in the PDB output, neither Pymol or VMD could read coordinates properly.

Let me just check the standard about that, and then let's merge this PR!

Luthaf:
> Let me just check the standard about that, and then let's merge this PR!

I misread the standard when writing this code, sorry about that.

Could you modify the code to use the real residue name/id when available? I feel like this is the better and less surprising thing to do.

pgbarletta:
> I misread the standard when writing this code, sorry about that.

It's quite annoying to read and understand, it took me a while to get it righ

>Could you modify the code to use the real residue name/id when available? I feel like this is the better and less surprising thing to do.

Yes, this chunk of code handles that.
```
 +        std::string resname;
 +        size_t resid;
 +        if (auto resopt = frame.topology().residue(i)) {
 +            resname = resopt->name();
 +            resid = resopt->id();
 +        }
 +        else {
 +            resname = "RES";
 +            resid = i;
 +        }
```

This is a line I got with a residue I named "ANA" and an atom named "N3"
`HETATM   11   N3 ANA X   3     -21.423  27.921 -23.949  0.00  0.00           N`

Luthaf:
Sorry, I missed that.

PR #51, opened 2016-10-11
Check for unit cell presence in NetCDF format

### LINKED ISSUE #45: Allowing Netcdf Amber trajectories without box information (Unit Cell)
I think the [current](https://github.com/chemfiles/chemfiles/blob/b7c23309d31d36d484b1ac8bd306bdbae97d201b/src/formats/NcFormat.cpp#L93) way to allow .nc trajectories without box info is not working as expected.

When `NCFormat::read_cell()` calls `dimension`, if the trajectory has no box information, an error gets thrown by `nc::check` (called by `dimension`) before `dimension` can return, say, `size=0`

Maybe the call to `dimenson` in `NCFormat::read_cell()` could be enclosed in a `try{...} catch{}`?

Luthaf:
This should fix #45. 

@pgbarletta, is it possible for you to send me a file without Unit Cell information, so that I can add a test here?

### DIFF
--- a/include/chemfiles/files/NcFile.hpp
+++ b/include/chemfiles/files/NcFile.hpp
@@ -132,6 +132,10 @@ public:
 
     //! Get the value of a specific dimmension
     size_t dimension(const std::string& name) const;
+    //! Get the value of an optional dimmension, or the default `value` if the
+    //! dimmension is not in the file
+    size_t optional_dimension(const std::string& name, size_t value) const;
+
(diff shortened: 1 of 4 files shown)


pgbarletta:
Great! Just mailed it to you. Let me know if it's OK.

Luthaf:
Yes, thank you.

Luthaf:
You may need to run `git submodule update` on the next git update. I do not know if there is a way to do that automatically.

PR #55, opened 2016-10-27
Residues, part n+1

Luthaf:
Some work to make the atoms/residues more usable
- b828926 removes the `Atom::type`, which was not used anywhere;
- 778cf74 rename `label => name` and `element => type`. @pgbarletta, what do you think of this? This match the naming of PDB and TNG formats, and looks like it is also the names used in most software.
- 1831d4a and  0e2c384 add `resname` and `resid` selections.

Also @pgbarletta if/when you update to the this version, take care that the order of parameters of `Atom` constructor where inverted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   used by Gromacs.
 * All the integers at C boundary have a fixed size, most of the time using
   `uint64_t`.
+* Remove the `Atom::type` value from C and C++ API.
 
 ## 0.6 (1 July 2016)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 * Add a public `Residue` class to C++ and C API to represent residue data.
   Residues are groups of atoms bonded together, which may or may not correspond
   to molecules; and are often used for bio-molecules.
+* Add the `resname` and `resid` selector, to select atoms based on their
+  residue.
 * Remove the `Trajectory::sync` and the `chfl_trajectory_sync` functions.
   To ensure that all content of a file is written to the disk, the user need to
   close it.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,12 +16,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   (and velocity) data to a frame, and the C API `chfl_frame_add_atom` function.
 * Rename `UnitCell::type` to `UnitCell::shape`. This also affect
   `chfl_cell_shape_t`, `chfl_cell_shape`, and `chfl_cell_set_shape`.
-* Add the atomic label in addition to the atomic element. This introduces the
-  `Atom::{label,element}` method and remove the `Atom::name` method. This also
-  remove `chfl_atom_{name,set_name}` and add `chfl_atom_{element,set_element}`
-  and `chfl_atom_{label,set_label}`.
-* The `name` selector now select atoms based on the atomic label. To select on
-  the element, use the new `element` selector.
+* Account for the difference between the atom name and atom type in some
+  formats (PDB, TNG, ...). This introduces the `Atom::type` member function
+  and the `chfl_atom_type` C API function.
+* Add the `type` selector, to select atoms based on their type.
 * All the floating point data uses doubles instead of floats. This concerns
   atomic data, positions and velocities.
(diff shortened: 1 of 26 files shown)


pgbarletta:
Thanks for the heads up. It's been a while since I updated, since I'm using a slightly mod version.

Agree with the `label => name`, but I think `element` was OK. 
From the PDB format specification:

```
COLUMNS          DATA TYPE           FIELD                   DEFINITION
-------------------------------------------------------------------------------------

77 - 78    |    LString(2)    |     element     |       Element symbol, right-justified.
```

Luthaf:
Would it be better if we ensured that `Atom::type` or `Atom::element` is effectively the name of an element from the periodic table?

I think an atom type could be `"CH3"` and the atom name `"CH3a"`, so I prefer using `type`. But I'd like to hear your opinion on this.

pgbarletta:
Right.
`Atom::name` could be anything, `Atom::element` would be an element from the table and `Atom::type` would specify how the atom is bonded. Like `CT` for sp3 carbons, `OH` for oxygens from hydroxy groups,  and so on.

I don't know if there's a standard set of types of atoms. `CT` and `OH` come from AMBER prep files, but I don't know if this is universal or not. Or if it even makes sense to have a list of the possible types of atoms for chemistry areas other than biochemistry.

Luthaf:
>  Atom::name could be anything, Atom::element would be an element from the table and Atom::type would specify how the atom is bonded. Like CT for sp3 carbons, OH for oxygens from hydroxy groups, and so on.

As this information is not present in most of the files formats, this would be hard to get. Do you think a member function `bool Atom::is_element()` would be useful here? It could check if `type` or `name` are in the periodic table.

> I don't know if there's a standard set of types of atoms. CT and OH come from AMBER prep files, but I don't know if this is universal or not.

I have never meet them.

Luthaf:
>  Atom::name could be anything, Atom::element would be an element from the table and Atom::type would specify how the atom is bonded. Like CT for sp3 carbons, OH for oxygens from hydroxy groups, and so on.

As this information is not present in most of the files formats, this would be hard to get. Do you think a member function `bool Atom::is_element()` would be useful here? It could check if `type` or `name` are in the periodic table.

> I don't know if there's a standard set of types of atoms. CT and OH come from AMBER prep files, but I don't know if this is universal or not.

I have never meet them in other file formats, so I would say they are not.

pgbarletta:
So `Atom::type` could be an element, a CG atom or a dummy atom, right?  `Atom::is_element()` could be useful.
Appart from that, there 're infinite ways to coarse grain molecules, so there's no way to differentiate CG atoms from dummy ones, right?

Luthaf:
> So Atom::type could be an element, a CG atom or a dummy atom, right?

Yep.

> Appart from that, there 're infinite ways to coarse grain molecules, so there's no way to differentiate CG atoms from dummy ones, right?

We could have a small list of usual CG atoms (ALA, CH3, H2O, ...) and try to classify using this. But I do not see any general solution here.

Luthaf:
Merging this, `Atom::is_element()` can be added later.

pgbarletta:
Looked up how Martini force field (I believe is the most popular CG force field) names its CG atoms in the PDB format.

------------------------------
ATOM    123  BB   TYR A  59      19.425  22.304  32.329  1.00  2.00           B
ATOM    124  SC1 TYR A  59      21.925  22.840  34.025  1.00  2.00           S
ATOM    125  SC2 TYR A  59      21.644  22.956  36.317  1.00  2.00           S
ATOM    126  SC3 TYR A  59      23.794  22.271  36.806  1.00  2.00           S

BB: Backbone ; SC: Sidechain

------------------------------

So a `Atom::is_CG()` would look like this:
```
// Check if the atom is a CG atom
static bool is_CG() {
    const auto CG_LIST = { "BB" ,  "SC1",  "SC2", "SC3", "H2O"};
    // clang-format on
    for (auto& each_CG : CG_LIST) {
        if (element == name_) {
            return true;
        }
    }
    return false;
}
```
and the new `Atom::is_element()` would be your old `Atom::is_element()` function, but as a member function that compares against `type_`

Do you think this is worth a PR, or are am I making too much fuss about something none will ever use?
Personally, I've never used the `Atom::is_element()` function, but I don't work with CG or dummy atoms, so I know all my atoms are real ones.

Luthaf:
> Do you think this is worth a PR, or are am I making too much fuss about something none will ever use?

I would use the YAGNI principle here: while there is no need for it, let's wait before adding more functionality. The code for these two functions is trivial, and adding more functions means supporting them in the C++, C, Fortran, Python, Julia and Rust API.

PR #74, opened 2017-04-06
Switch to BSD license

Luthaf:
Fix #48.

This changes the license of the code to BSD, which only requires attribution of copyright. I believe that a more permissive license can help adoption of the code.

@pgbarletta and @pelsa, you contributed code to this repository under the previous license. Do you agree to distribute your contributions under the BSD license? If so, please leave a comment here!

PR #77, opened 2017-04-14
Write name in XYZ format rather than type

### LINKED ISSUE #76: Write names instead of types in XYZ format
Change in XYZ write function

lscalfi:
Fix #76


Luthaf:
Can you add a test for this? You should change [this file](https://github.com/chemfiles/chemfiles/blob/b3c150c50ae282c7c3f5d679260fc4aa817dfef8/tests/formats/xyz.cpp#L101) to add an atom with different name and type, and then check that the name is picked instead of the type. This will prevent regressions in the future.

Luthaf:
The appveyor failure is not from your code, and the same as in #74: a change in the environment (some DLL is not found). I have a fix coming for this.

Luthaf:
Appveyor failure should be fixed now, if you can rebase this on top of master the tests should pass.

PR #78, opened 2017-04-18
Set types in configuration file

### LINKED ISSUE #75: Allow to override element data in a configuration file
This is a proposal to add *configuration file* to chemfiles. These configuration files could contain two kind of data:

- element mapping: allow the user to specify that Oh is an oxygen, that C_h should be a carbon, ...
- element data: give mass, charge, radius for pseudo-elements: CH3, NH2, ALA, ... This could also be used to override pre-existing values: setting the charge for all oxygens for example.

## User interface

The configuration file should be automatically loaded from `~/.chemfilesrc` if such a file exists on OS X and Linux. On Windows, I'll have to look where to put such files.

The user could also provide a specific file to be used, changing the behaviour of Atom constructor.

## File format and syntax

I propose to use the [TOML](https://github.com/toml-lang/toml) format. There are multiple C++ implementation for this, I like the [cpptoml](https://github.com/skystrife/cpptoml) one, but I am open to other suggestions.

The file could look like this:

```toml
# Set the atomic type for the given atomic names 
[types]
O_h = "O"
"234" = "Rd"

# Override data by atom type
[atoms]
O = {charge = -1.0}
CH3 = {radius = 4.2}
```

## Code changes

To implement this, we would need to change how the atomic data are accessed to check first for anything in a configuration file.

We also need to add a C++ function to set the path to a new configuration file, and mirror this function in the C API.

---

What do you think about this proposal? Is it useful? What could be missing? 

Ping @lscalfi who is interested in this.

lscalfi:
Fix #75 

Allows to read in the file `~/.chemfilesrc` and set the types in the simulation. The configuration file must be in TOML:
```
[types]
oh = "O"
Hw = "H"
12 = "Ar"
```

### DIFF
--- a/src/Configuration.cpp
+++ b/src/Configuration.cpp
@@ -1,0 +1,38 @@
+// Chemfiles, a modern library for chemistry file reading and writing
+// Copyright (C) 2015-2016 Guillaume Fraux and contributors
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+#include "chemfiles/Configuration.hpp"
+#include "chemfiles/Error.hpp"
+#include "cpptoml/cpptoml.h"
+#include <fstream>
+#include <unordered_map>
+using namespace chemfiles;
+
+const char PATH[] = "~/.chemfilesrc";
+
+Configuration Configuration::get() {
+    Configuration configuration;
+    if (std::ifstream(PATH)) {
+        try {
+            auto toml = cpptoml::parse_file(PATH);
+
+            for (auto& entry: *toml) {
+                auto name = entry.first;
+                auto type = entry.second->as<std::string>()->get();
+                configuration.rename_.insert({name, type});
+            }
+        } catch (const cpptoml::parse_exception& e) {
+            throw ConfigurationError("Failed to parse " + std::string(PATH) + ": " + e.what());
+        }
+    }
+    return configuration;
+}
+
+bool Configuration::find(std::string name) {
+    auto search = rename_.find(name);
+    return (search != rename_.end());
+}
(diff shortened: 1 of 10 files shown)


src/Configuration.cpp
```diff
@@ -10,12 +10,20 @@
...
 #include <fstream>
 #include <unordered_map>
+#include <cstdlib>
 using namespace chemfiles;
 
-const char PATH[] = "~/.chemfilesrc";
+static std::string find_file() {
+    std::string PATH;
+    if (const char* env_home = std::getenv("HOME")) {
+        PATH = std::string(env_home) + "/.chemfilesrc";
```
Luthaf: On Windows, this will need to use `\` as path separator.

src/Configuration.cpp
```diff
@@ -10,12 +10,20 @@
 #include "cpptoml/cpptoml.h"
 #include <fstream>
 #include <unordered_map>
+#include <cstdlib>
 using namespace chemfiles;
 
-const char PATH[] = "~/.chemfilesrc";
+static std::string find_file() {
+    std::string PATH;
+    if (const char* env_home = std::getenv("HOME")) {
```
Luthaf: This will only work on Linux. Windows does not define `HOME`, and OS X may not. See this for a possible resolution: http://stackoverflow.com/a/3733955/4692076

But cross-platform path handling is hard =/

Luthaf:
On Windows, we also may want to use `C:\Users\%USERNAME%\AppData\Local\chemfilesrc` instead of `~/.chemfilesrc`, it looks like the standard way of doing it.

lscalfi:
We should have: 
   - a default file (in ~/.chemfilesrc) to find for each platform
   - an optional user defined path
I will work on this :-)

Luthaf:
This PR work is now part of #86!

Luthaf:
Closing in favor of #86

PR #86, opened 2017-06-18
Configuration files for atomic type renaming

Luthaf:
Part of #75 

This PR allow to store configuration data in `.chemfilesrc` files, with the following formatting:

```toml
[types]
old_type = "new_type"

# For example
Ow = "O"
Hw = "H"
C3 = "C"
```

I am open to discussion concerning the name of the TOML table: `rename` feels a bit off, as we are not changing the atom name, but only the atom type.


Luthaf:
Code coverage is disturbed by the `[ci skip]` commit on master. Let's merge this, I'll fix coverage later if it is needed.

PR #88, opened 2017-07-19
Feedmereadmes edit of README

LappleApple:
Pretty good job already. I just moved some things around a bit. Some questions/tips:
- use this product template (https://github.com/zalando/zalando-howto-open-source/blob/master/producttemplate.md) or something similar to assess your audience and then bring some of the user more into the README. Especially:
  - why would I use this?
  - is this different/better/newer/simpler than similar tools out there? Or are there no similar tools like it (which could be a good thing to mention)?
  - how has this helped you and/or others work faster/more effectively/more simply?
 - how do you quantify the words you use to "sell" Chemfiles: "Modern" (compared to what?); "easy-to-use" (because of its object-oriented nature? what does that actually *look* like, concretely, such that it's easy to use?)
- why five classes? 
- why the languages you've chosen? because they're most popular among researchers at the moment?
- for contributors, I wonder what the labels in the Issues Tracker mean? Maybe you can identify some "quick starters" items for brand-new contributors to work on?
- wondering why you chose BSD license

Thank you for submitting to FMRM!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 ## Chemfiles: a library for reading and writing chemistry files
 
-[![Documentation](https://img.shields.io/badge/docs-latest-brightgreen.svg)](http://chemfiles.github.io/chemfiles/)
+[![Documentation](https://img.shields.io/badge/docs-latest-brightgreen.svg)](http://chemfiles.org/chemfiles/)
 [![Build Status -- Linux & OSX](https://img.shields.io/travis/chemfiles/chemfiles/master.svg)](https://travis-ci.org/chemfiles/chemfiles)
 [![Build Status -- Windows](https://ci.appveyor.com/api/projects/status/dvn6nr3lsssd23lo/branch/master?svg=true)](https://ci.appveyor.com/project/Luthaf/chemfiles/branch/master)
 [![Code Coverage](http://codecov.io/github/chemfiles/chemfiles/coverage.svg?branch=master)](http://codecov.io/github/chemfiles/chemfiles?branch=master)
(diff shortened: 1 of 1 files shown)


Luthaf:
Who, thank you very much! This is very appreciated.

> - use this product template (https://github.com/zalando/zalando-howto-open-source/blob/master/producttemplate.md) or something similar to assess your audience and then bring some of the user more into the README.

I'll try to think about this, looks interesting!

>  - is this different/better/newer/simpler than similar tools out there? Or are there no similar tools like it (which could be a good thing to mention)?

I have a bit about this in the [documentation](http://chemfiles.org/chemfiles/latest/others.html), do you think I should move it to the README ?

> - how do you quantify the words you use to "sell" Chemfiles: "Modern" (compared to what?); "easy-to-use" (because of its object-oriented nature? what does that actually *look* like, concretely, such that it's easy to use?)

Modern is more about the usage of C++11 vs older standards. You make me realize it might not be the best word to appeal scientists. The easy-to-use part is a goal of the code, which I try to illustrate in the example. Do you think I should extend this part a bit? I am a bit afraid that a README too long might be worst.

> - why five classes? 

This is actually left-over from previous version. The "five classes" was an emphasis on the easy-to-use side, as the most known competitor ([openbabel](http://openbabel.org/wiki/Main_Page)) contains a lot of classes and functions.

> - why the languages you've chosen? because they're most popular among researchers at the moment?

This, and because they are the one I know enough to write C bindings. I am thinking to extend the list with JavaScript (either with emscripten or node) and Java (actually all JVM languages), but this takes quite a bit of time.

> - for contributors, I wonder what the labels in the Issues Tracker mean? Maybe you can identify some "quick starters" items for brand-new contributors to work on?

We do have the [E-easy](https://github.com/chemfiles/chemfiles/issues?q=is%3Aissue+is%3Aopen+label%3AE-easy) label, maybe I should link to it more. I'll add explanation for the other labels on the Contributing.md file.

> - wondering why you chose BSD license

Asking the hard questions here =). Most of the competitors are GPL licensed, and I wanted something that could be used by companies in commercials products as well as in open source work. I already talked to one company that showed interest in using chemfiles because it is BSD. On the BSD vs MIT, I still like the citation/attribution clause of BSD.

> Thank you for submitting to FMRM!

Thank you for taking the time to do this !

LappleApple:
>>I have a bit about this in the documentation, do you think I should move it to the README ?
Yes, possibly in a briefer form. But definitely something. :)

>>Do you think I should extend this part a bit? I am a bit afraid that a README too long might be worst.
I wouldn't extend it so much as just make the point clear. For example, quantify it like "modern (C++11)".

I would include the plan for JS. That could get you some contributors. If it's already listed in the issues tracker, you can point to the relevant issues. :)

>> license stuff
You might add the point about getting commercial products/companies in your "why use this?" section near the top, to point out your intended users.

You're asking cool questions and thinking about this in a great way!

Luthaf:
Thank you again for your help here @LappleApple

LappleApple:
:1st_place_medal:

PR #89, opened 2017-07-29
Altered elements.py to fetch default element colors for viewers from BODR

filipecosta90:
Altered elements.py to fetch default element colours for viewers from the the Blue Obelisk Data Repository. 
The header file periodic.hpp is produced containing now RGB info regarding every atom type.
The atom class is altered do contain the referred information.


Luthaf:
Thank you for the pull request ! I am currently off grid until the 15th of august, I'll take some time when I am back to look at your changes.

include/chemfiles/Atom.hpp
```diff
@@ -66,6 +66,16 @@ class CHFL_EXPORT Atom {
     /// found.
     int atomic_number() const;
 
+    /// Try to get the Default Red(from RGB) element colors for viewers, if defined. Returns -1.0f if it can not be
+    /// found.
+    float color_r() const;
```
Luthaf: I would rather have a single method returning the `(R, G, B)` vector at once. Something like `Vector3D rgb_color() const`.

include/chemfiles/periodic.hpp
```diff
@@ -14,140 +13,144 @@
...
+    //! Mass in atomic units
     const double mass;
-    /// Covalent radius in Angstrom
+    //! Covalent radius in Angstrom
     const double colvalent_radius;
-    /// Van der Waals radius in Angstrom
+    //! Van der Waals radius in Angstrom
     const double vdw_radius;
+    //! Default element colors for viewers
+    const float elementColor_r;
```
Luthaf: Following the same idea of using a single vector for the three components of the color, this could be `const double[3] rgb_color`.

Luthaf:
As this adds a new method to the public C++ interface, it should also add a new public C function, to be able to expose this to all the other languages bindings. Do you want to do this too ?

Luthaf:
Ping @filipecosta90, do you still want to work on this ? If you don't have time now, that's completely fine !

filipecosta90:
hi there, 
I will make the changes tomorrow. 
regards,
Filipe

Luthaf:
I'm closing this Pull request due to inactivity. Feel free to resubmit it with the requested changes!

Issue #95, opened 2017-09-24
PDB format: ATOM / HETATM selection


pgbarletta:
Hi!
   I need to distinguish between atoms that came from ATOM records from those that came from HETATM records, and for what I have seen, chemfiles doesn't do this yet. 
My planed changes would be:

1_ Modify `Atom.hpp`: add a private member variable to the `Atom` class. Something like `bool hetatm_ = false`.

2_ Modify `Atom.cpp`: to print this bool. Eg, `const bool hetatm() const { return hetatm_; }`. 

3_ Modify `PDB.cpp`: from this:
```
case Record::ATOM:
case Record::HETATM:
         read_ATOM(frame, line);
         break;
```
to this:

```
case Record::ATOM:
         read_ATOM(frame, line);
         break;
case Record::HETATM:
         read_HETATM(frame, line);
         break;
```
or, ---since the difference between the 2 functinos (`read_ATOM` and `read_HETATM`) would only lie in the setting of the putative `hetatom_` variable---, something like this:

```
case Record::ATOM:
         read_xyz(frame, line, false);
         break;
case Record::HETATM:
         read_xyz(frame, line, true);
         break;
```
4_ Maybe add some test?

Hope I haven't overlooked anything.

Luthaf:
Hey !

This looks like an interesting addition. Do you know other formats that also do this ATOM/HETATOM distinction ? Maybe mmCIF ? 

I am a bit worried about this that it is adding a public API function only for one format. Maybe it would be better as "additional data" in the frame, like a `is_hetatom = std::vector<bool>` in the frame (see #85 for the idea on storing additional data on the frame).

This would be more work, but if you need it, I can do the foundation work for this soon. Then you could use this additional data storage facility to store and then retrieve the information you need in PDB files ?

pgbarletta:
Not that I know of. Every other format just reads "atoms". [mmCIF](http://mmcif.wwpdb.org/docs/pdb_to_pdbx_correspondences.html#ATOMP)  groups ATOMs and HETATMs in the Data Item `_atom_site`.

Right, it's a very specific use. How about this:

Since `Frame` objects are created for each frame and they all map to the same `Atom` object, I think this variable would correspond to the `Atom` class, since it would be a constant for each `Atom` object. I mean, it shouldn't change along a trajectory.

I'm thinking in a template class inside the `Atom` class, that is, a new member variable. It could be called just `info` and the user would access it by the `info()` method, which would return a reference to it. 
[...]

PR #96, opened 2017-09-25
1st implementation of Property struct and PBD format with ATOM/HETATM…

pgbarletta:
… markup

from https://github.com/chemfiles/chemfiles/issues/95

OK, I learned some C++ today. Had never seen tagged unions before (I didn't even know that was a thing). 
The code still needs some work, but at least is functional (tested it in my application and it worked).  I should also update  the `write` function of `PDB.cpp` to check if the `Atom` object has `is_hetatm` set.

1. Didn't know where to put the new `Property` struct, so I chose `types.hpp`.
2. I'm using the base `Error()` function. 
3. Couldn't make it work with `std::string` and had to go with `char`. It should be possible though, since C+11 to use strings inside tagged unions, declaring constructors and destructors.
4. Couldn't
5. It seems that my `clang-format` package changed your code, I thought you used that package too. Once I'm done with this code I'll edit the original code from VIM, to avoid formatting changes.
6. Had to initialize a temporary variable `  struct Property tmp {is_hetatm};` to make it work. List initialization was the only way to avoid: `error: cannot convert ‘bool’ to ‘chemfiles::Property::<anonymous enum>`. Same thing would happen if either `kind_ enum` or the anonymous `union` were private, so I had to go with `Struct`

### DIFF
--- a/include/chemfiles/Atom.hpp
+++ b/include/chemfiles/Atom.hpp
@@ -4,10 +4,14 @@
 #ifndef CHEMFILES_ATOM_HPP
 #define CHEMFILES_ATOM_HPP
 
+#include <map>
 #include <string>
 
 #include "chemfiles/exports.hpp"
 #include "chemfiles/optional.hpp"
+#include "chemfiles/types.hpp"
+
+using property_map = std::map<std::string, chemfiles::Property>;
 
 namespace chemfiles {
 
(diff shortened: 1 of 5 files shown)


Luthaf:
This looks nice, thanks !

> It seems that my clang-format package changed your code, I thought you used that package too. Once I'm done with this code I'll edit the original code from VIM, to avoid formatting changes.

This repository used to have a `.clang-format` file, but I removed it in https://github.com/chemfiles/chemfiles/commit/cb808effcd89081c5b4da4a41e55e7833699b764 because I was not using it. But maybe it is time to bring it back ^^

Maybe you can use the previous .clang-format file locally to reformat the code ? Currently the diff is very hard to read.

> I'm using the base Error() function.

You can add a new `PropertyError` class for this.

> Didn't know where to put the new Property struct, so I chose types.hpp.

You can start a new file, the code is complex enough for this. 

> Had to initialize a temporary variable struct Property tmp = {is_hetatm}; to make it work. List initialization was the only way to avoid: error: cannot convert ‘bool’ to ‘chemfiles::Property::<anonymous enum>

You should not need to do that, I'll give it a more in-depth look after you reformat the code =)

FYI I also started working on this in these commits (b78ee3958264a2c792fb85b86b1c331f84c6d90d, 0733b8ed39b1cd2973684d0ab20d75076521a46b, b9854efb53135f844ca3a5ec6a442e02c8aa3260). Maybe you can draw some inspiration from it, concerning the use of std::string for example. I am fine with using your code instead!

pgbarletta:
Oh, that pretty much solves it.
Correct me if I'm wrong, but with those commits, all is left for me to do is to modify the PDB format. Right?

Luthaf:
>  Correct me if I'm wrong, but with those commits, all is left for me to do is to modify the PDB format. Right?

This, and adding a `CHFL_PROPERTY` struct on the C API side (I have a WIP for this locally 😄).

pgbarletta:
Closed and replaced by https://github.com/chemfiles/chemfiles/pull/97

PR #97, opened 2017-09-26
Add arbitrary properties to atoms and frames

### LINKED ISSUE #85: Store additional data in the frames
Some formats defines more information than what is used inside chemfiles. We may want to keep this data around, in case someone would like to use it.

I propose to add a generic data storage inside a `Frame`, to be used for additional data from the file format (think to occupancy in PDB or CIF files, forces, ...). This storage should be able to store per atom and global data.

On the C++ side, data can be stored in a `std::map<std::string, value>`, where `value` is a tagged union of the allowed data types. I think the allowed data types should be scalars (as doubles), 3D vectors and strings.

On the C side, we should wrap the union into an opaque struct, and make it a bit more type safe

```c
/// Setter functions
chfl_status chfl_frame_store_scalar(const CHFL_FRAME* frame, const char* name, double value);
chfl_status chfl_frame_store_string(const CHFL_FRAME* frame, const char* name, const char* value);
chfl_status chfl_frame_store_vector(const CHFL_FRAME* frame, const char* name, chfl_vector value);

enum chfl_data_type {
    CHFL_SCALAR,
    CHFL_STRING,
    CHFL_VECTOR,
}

/// Get the type of stored data, or an error
chfl_status chfl_frame_load_type(const CHFL_FRAME* frame, const char* name, chfl_data_type* type);

/// Extract the value by calling the right function
chfl_status chfl_frame_load_scalar(const CHFL_FRAME* frame, const char* name, double* value);
chfl_status chfl_frame_load_string(const CHFL_FRAME* frame, const char* name, const char** value);
chfl_status chfl_frame_load_vector(const CHFL_FRAME* frame, const char* name, chfl_vector* value);
```

Luthaf:
Fix #85.

@pgbarletta, this can replace your implementation from #96 if you are ok with it. You can then add the PDB part of this =)

If you have some time and you could give this a quick look to tell me if anything looks weird,  it would be great! Compared to our last discussion on the topic, I changed to return type of `Frame::get` and `Atom::get` to `optional<const Property&>`, because the other functions on Atom already use this.

The alternative would be to return `const Property&` and throw an exception on missing data.

### DIFF
--- a/tests/capi/doc/chfl_property/vector3d.c
+++ b/tests/capi/doc/chfl_property/vector3d.c
@@ -1,0 +1,21 @@
+// Chemfiles, a modern library for chemistry file reading and writing
+// Copyright (C) Guillaume Fraux and contributors -- BSD license
+
+#include <chemfiles.h>
+#include <assert.h>
+#include <math.h>
+
+int main() {
+    // [example]
+    CHFL_PROPERTY* property = chfl_property_vector3d((chfl_vector3d){2, 3.2, -1});
+
+    chfl_vector3d value = {0};
+    chfl_property_get_vector3d(property, value);
+    assert(fabs(value[0] - 2) < 1e-12);
+    assert(fabs(value[1] - 3.2) < 1e-12);
+    assert(fabs(value[2] - -1) < 1e-12);
+
+    chfl_property_free(property);
+    // [example]
+    return 0;
+}
--- a/doc/src/capi/chfl_atom.rst
+++ b/doc/src/capi/chfl_atom.rst
@@ -37,4 +37,8 @@ Function manipulating ``CHFL_ATOM``
 
 .. doxygenfunction:: chfl_atom_atomic_number
 
+.. doxygenfunction:: chfl_atom_set_property
+
+.. doxygenfunction:: chfl_atom_get_property
+
 .. doxygenfunction:: chfl_atom_free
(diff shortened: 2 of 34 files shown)


tests/capi/doc/chfl_property/vector3d.c
```diff
@@ -0,0 +1,20 @@
...
+// Copyright (C) Guillaume Fraux and contributors -- BSD license
+
+#include <chemfiles.h>
+#include <assert.h>
+
+int main() {
+    // [example]
+    CHFL_PROPERTY* property = chfl_property_vector3d((chfl_vector3d){0, 3.2, -1});
+
+    chfl_vector3d value = {0};
```
pgbarletta: Should be
`chfl_vector3d value = {0., 0., 0.};`
right?

pgbarletta:
Yeah, I was about to cancel that PR. 
Good, once you merge this I'll cancel my PR, fork again and add the PDB support for HETATM/ATOM using `Property`.

tests/capi/doc/chfl_property/vector3d.c
Luthaf: Missing value in the brace are replaced by 0 automatically: see http://en.cppreference.com/w/c/language/array_initialization, " Initialization from brace-enclosed lists" section. This is really neat when working with a lot of arrays =)

Luthaf:
> Yeah, I was about to cancel that PR.

Sorry about this, I'll try to check first before starting an implementation next time. Maybe you can also leave a comment on the issue when you start working on something ? I'll try to do the same.

pgbarletta:
No worries, I learnt quite a bit!

PR #99, opened 2017-09-28
Also normalize 1-letter atomic type

Luthaf:
We already normalize 2 letters atomic types when looking for atomic properties: `Atom("ZN")` and `Atom("zN")` use the Zinc atomic data.

This introduce the same normalization for 1 letter atomic type, so that `Atom("C")` and `Atom("c")` are the same thing.

@pgbarletta, do you think we should do this? It feels more coherent to me.

### DIFF
--- a/src/Atom.cpp
+++ b/src/Atom.cpp
@@ -18,8 +18,14 @@ static char to_lower(char c) {
 
 static optional<const ElementData&> find_element(const std::string& name) {
     std::map<std::string, ElementData>::const_iterator periodic;
-    if (name.length() == 2) {
-        auto normalized = name;
+    std::string normalized;
+    if (name.length() == 1) {
+        normalized = name;
+        normalized[0] = to_upper(normalized[0]);
+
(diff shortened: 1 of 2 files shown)


pgbarletta:
Mmm..., the lowercase/uppercase problem arose on PDB files; I think I've never seen a lowercase letter in the atom name column. It's always uppercase letters on all fields, although you never know what surprise crystallographers have for you.

I'm not sure of this. Maybe you've found this problem on other formats?

Luthaf:
> I'm not sure of this. Maybe you've found this problem on other formats?

No, not really. I found this while writing example for the Atom class. 

On the plus side, this change makes things a bit more consistent and less surprising. But at the same time this is introducing more special cases.

pgbarletta:
>  I found this while writing example for the Atom class.

Right, it does prevent Chemfiles users from writing ill-formatted files. Yeah, I think it's useful.

Luthaf:
Ok, let's do this then.

PR #101, opened 2017-10-02
Various changes, again

Luthaf:
@pgbarletta, I'd love to have your input on 3592daa6aab7b03c24fe5c415be4fe38212dd8df. This commit removes the assigned "RES" name for atoms without a residue. It feels nicer to me, because we are not making up some information. But I am not sure this is creating a valid PDB file then. The standard is not clear whether residues name is optional or not. What is your point of view on this ?


pgbarletta:
Your original implementation was just fine. Technically, all fields are mandatory, except for the Occupancy and Temperature Factor fields since that data is not available in RMN structures.

Luthaf:
OK, thanks for the clarification. I still don't like `RES`, it looks like it could be an actual residue name. Maybe `XXX` would do it?

pgbarletta:
Sure, it even has the virtue of being funny.

Luthaf:
Did not thought of that ^^ well it should still be OK to use

PR #102, opened 2017-10-02
Added is_hetatm Property when reading&writing PDB files. Added left j…

pgbarletta:
…ustification for atom name field.
From: https://github.com/chemfiles/chemfiles/issues/95

Sorry, had busy weekend.

You'll see that I also changed the PDB write format. I'm attaching image from the PDB format where you can see that the Atom Name Field is the only one with left justification. I know, annoying, but this is coming from the people who thought that we would never have a molecule with a residue number larger than 4 digits

pdb

### DIFF
--- a/tests/capi/doc/chfl_last_error.c
+++ b/tests/capi/doc/chfl_last_error.c
@@ -11,7 +11,7 @@ int main() {
     chfl_trajectory_open("noformat", 'r');
 
     const char* error = chfl_last_error();
-    assert(strcmp(error, "file at 'noformat' do not have an extension, provide a format name to read it") == 0);
+    assert(strcmp(error, "file at 'noformat' does not have an extension, provide a format name to read it") == 0);
     // [example]
     return 0;
 }
--- a/tests/capi/doc/chfl_set_warning_callback.c
+++ b/tests/capi/doc/chfl_set_warning_callback.c
@@ -24,7 +24,7 @@ void warning_callback(const char* message);
         // Generate a warning event
         chfl_trajectory_open("noformat", 'r');
 
-        assert(strcmp(LAST_WARNING, "file at 'noformat' do not have an extension, provide a format name to read it") == 0);
+        assert(strcmp(LAST_WARNING, "file at 'noformat' does not have an extension, provide a format name to read it") == 0);
         return 0;
     }
 
--- a/tests/capi/misc.cpp
+++ b/tests/capi/misc.cpp
@@ -12,7 +12,7 @@
 static std::string generate_chemfiles_error() {
     // Generate a log event
     CHECK_FALSE(chfl_trajectory_open("noformat", 'r'));
-    return "file at 'noformat' do not have an extension, provide a format name to read it";
+    return "file at 'noformat' does not have an extension, provide a format name to read it";
 }
 
 TEST_CASE("Errors") {


include/chemfiles/formats/PDB.hpp
```diff
@@ -30,7 +30,7 @@ class PDBFormat final: public Format {
     // Read CRYST1 record
     void read_CRYST1(Frame& frame, const std::string& line);
     // Read ATOM and HETATM records
-    void read_ATOM(Frame& frame, const std::string& line);
+    void read_ATOM(Frame& frame, const std::string& line, const bool is_hetatm);
```
Luthaf: const is not needed here: bool is a primitive type passed by value

src/formats/PDB.cpp
```diff
@@ -131,7 +133,8 @@ void PDBFormat::read_CRYST1(Frame& frame, const std::string& line) {
     }
 }
 
-void PDBFormat::read_ATOM(Frame& frame, const std::string& line) {
+void PDBFormat::read_ATOM(Frame& frame, const std::string& line,
+  const bool is_hetatm) {
```
Luthaf: Same, const not needed here

src/formats/PDB.cpp
```diff
@@ -299,6 +304,21 @@ void PDBFormat::write(const Frame& frame) {
...
+        auto atom_property = frame.topology()[i].get("is_hetatm");
+        if (atom_property) {
+          try {
+            if (atom_property->as_bool()) {
+              atom_hetatm = "HETATM";
+            } else{
+              atom_hetatm = "ATOM  ";
+            }
+          }
+          catch (...) {
```
Luthaf: why the catch(...) ? Is it to catch the fact that the property might not be a boolean value ? In this case, you can catch `PropertyError`, and I would send a warning about the atom having a non-boolean `is_hetatm` property.

Luthaf:
This looks good to me, thanks for implementing it!

It is missing a test to check that if the property is set, `ATOM` gets written instead of `HETATM`, and another one that check that the property is set upon reading. The [water.pdb](https://github.com/chemfiles/tests-data/blob/3c188339993f81e37cb391c0b2fd7b00cdaed599/pdb/water.pdb) file does contains ATOM record, and the [MOF-5.pdb](https://github.com/chemfiles/tests-data/blob/3c188339993f81e37cb391c0b2fd7b00cdaed599/pdb/MOF-5.pdb) contains HETATM.

Also, which properties are set by a given format should be documented somewhere. I don't know where it could fit, maybe put it in `doc/src/formats.rst` for now.

src/formats/PDB.cpp
pgbarletta: Like this?
```
catch (const PropertyError &e) {
            warning("\'is_hetatm\' property set to non-bool variable.
            Defaulting to HETATM");
            atom_hetatm = "HETATM";
          }
```

Luthaf: Yes =)

pgbarletta:
Added a new small pdb file for the new test, since I wanted both ATOM and HETATM records in the same file.

pgbarletta:
New PDB file @ https://github.com/chemfiles/tests-data/pull/1

Luthaf:
@pgbarletta, there seems to be a conflict with #101 here. You'll need to rebase your pull-request. If you don't know how to do this, ask me and I'll be happy to help you!

pgbarletta:
supplanted by https://github.com/chemfiles/chemfiles/pull/106

Issue #103, opened 2017-10-02
PDB format: write as models


pgbarletta:
I've been using a modified write function for a while. It came handy to let visualization programs know they were reading a trajectory:
`void PDBFormat::write_model(const Frame& frame, const size_t model_n);`

The only difference being that each time the first and last lines would be:
`    fmt::print(*file_, "MODEL     {:4d}\n", model_n);`
`    fmt::print(*file_, "ENDMDL\n\n");`

The result is a PDB format without an `END` line, but that doesn't cause any trouble, and if necessary, it's easy to fix.
I know it's a scrappy implementation, but I think it's a useful feature.  To make it nicer, I was thinking of overloading the `write` function for the PDB format:

`void PDBFormat::writel(const Trajectory& trajectory);`

What do you think @Luthaf?

Luthaf:
> It came handy to let visualization programs know they were reading a trajectory:

I initially used the `END` record, because this is what VMD does for trajectories in PDB format. I don't know  about the MODEL record, is it intended for use with trajectories ?

Do you have an example of a visualization program supporting the MODEL record, but not multiple END record? As long as this change does not break VMD (which is widely used in the biomolecular community), I am fine with it!

If we implement this for writing PDB, we should also add support for it when reading PDB files.

--- 

The idea would be to replace

```
CRYST1 ...
ATOM ...
END
CRYST1 ...
ATOM ...
END
CRYST1 ...
ATOM ...
END
```

by something like this:

```
MODEL 1
CRYST1 ...
ATOM ...
ENDMDL
MODEL 2
CRYST1 ...
ATOM ...
ENDMDL
MODEL 3
CRYST1 ...
ATOM ...
ENDMDL
```

Am I right ? 

If so, we don't need to pass an additional model_n parameter: we can store it in the PDBFormat class itself (it is called step_ in some other formats), or use the `Frame::step` value for this.

> The result is a PDB format without an END line, but that doesn't cause any trouble, and if necessary, it's easy to fix.

Yeah, we could add a `Format::finalize` function and reflect it to `Trajectory` if needed. Or just call it from the Trajectory destructor.

pgbarletta:
> I initially used the END record, because this is what VMD does for trajectories in PDB format. I don't know about the MODEL record, is it intended for use with trajectories ?

It was meant for RMN experiments that sampled different conformations. That's why they call it `MODEL` instead of, for example `STEP`. They see it as alternative structures instead of dynamics.

[...]

PR #108, opened 2017-10-05
1st step in MODEL/ENDMDL support for PDB format. Added step_ variable…

pgbarletta:
… support in the PDB format and added MODEL record as an IGNORED record.

From: https://github.com/chemfiles/chemfiles/issues/103

OK, I'm gonna need some help with this. I couldn't find the code to check for the last frame of the trajectory. That is, how do I know, inside the `PDBFormat::Write` function, that the current `Frame` is the last of the trajectory I'm writing?
If write a 10 frames trajectory, this is how my PDB file is ending:
```
MODEL   11
CRYST1    0.000    0.000    0.000  90.00  90.00  90.00 P 1           1
ENDMDL
```
I could fix this and add the `END` statement if I could detect the last frame. 
Or maybe I'm not using `step_` and `steps_positions_`  properly. If this is the case, then I could fix the bad ending, but I would still need a place where to add the `END` record (which by the way, is not mandatory). Maybe inside the `~Trajectory` destructor like you said? But I would have to check, ---for the `Trajectory` object being deleted---, that:
`format_ == std::unique_ptr<PDBFormat>` is `True`
and that is was written to a file, which I wouldn't know how to do. 
Maybe this `END` statement should come just before closing a PDB format file? And where does that happen? 
I don't know, I'm getting hazy.

### DIFF
--- a/src/formats/PDB.cpp
+++ b/src/formats/PDB.cpp
@@ -45,7 +45,7 @@ enum class Record {
 static Record get_record(const std::string& line);
 
 PDBFormat::PDBFormat(const std::string& path, File::Mode mode)
-  : file_(TextFile::create(path, mode)), step_(0) {
+  : file_(TextFile::create(path, mode)), models_(0) {
     while (!file_->eof()) {
         auto position = file_->tellg();
         if (!file_ || position == std::streampos(-1)) {
@@ -89,7 +89,7 @@ void PDBFormat::read(Frame& frame) {
             read_CONECT(frame, line);
             break;
         case Record::MODEL:
-            ++step_;
+            models_++;
             break;
         case Record::END:
             goto end; // We have read a frame!
@@ -297,7 +297,7 @@ static std::string to_pdb_index(uint64_t i) {
 
 void PDBFormat::write(const Frame& frame) {
     written_ = true;
-    fmt::print(*file_, "MODEL {:>4}\n", step_+1);
+    fmt::print(*file_, "MODEL {:>4}\n", models_ + 1);
 
     auto& cell = frame.cell();
     check_values_size(Vector3D(cell.a(), cell.b(), cell.c()), 9, "cell lengths");
@@ -410,7 +410,7 @@ void PDBFormat::write(const Frame& frame) {
 
     fmt::print(*file_, "ENDMDL\n");
 
-    ++step_;
+    models_++;
     steps_positions_.push_back(file_->tellg());
 }
 
--- a/src/formats/PDB.cpp
+++ b/src/formats/PDB.cpp
@@ -245,6 +245,7 @@ bool forward(TextFile& file) {
 
             if (line.substr(0, 6) == "ENDMDL" &&
             file.readline().substr(0, 3) == "END" ) {
+              // An ENDMDL was followed and an END. This means `END`.
               return false;
             }
 
@@ -425,7 +426,7 @@ void check_values_size(const Vector3D& values, unsigned width, const std::string
 }
 
 PDBFormat::~PDBFormat() {
-  if (written_) {
-    fmt::print(*file_, "END\n");
-  }
+    if (written_) {
+      fmt::print(*file_, "END\n");
+    }
 }
--- a/src/formats/PDB.cpp
+++ b/src/formats/PDB.cpp
@@ -264,7 +264,7 @@ Record get_record(const std::string& line) {
                rec == "CAVEAT" || rec == "COMPND" || rec == "EXPDTA" ||
                rec == "KEYWDS" || rec == "OBSLTE" || rec == "SOURCE" ||
                rec == "SPLIT " || rec == "SPRSDE" || rec == "TITLE " ||
-               rec == "JRNL  " || rec.substr(0, 5) == "MODEL" ) {
+               rec == "JRNL  " || rec == "MODEL " ) {
         return Record::IGNORED_;
     } else {
         return Record::UNKNOWN_;
@@ -393,7 +393,7 @@ void PDBFormat::write(const Frame& frame) {
             fmt::print(*file_, "\n");
         }
     }
-    
+
     fmt::print(*file_, "ENDMDL\n");
 
     ++step_;
--- a/src/formats/PDB.cpp
+++ b/src/formats/PDB.cpp
@@ -242,9 +242,15 @@ bool forward(TextFile& file) {
     while (true) {
         try {
             auto line = file.readline();
-            // Applies to 'END' and 'ENDMDL' records.
+
+            if (line.substr(0, 6) == "ENDMDL" &&
+            file.readline().substr(0, 3) == "END" ) {
+              return false;
+            }
+
             if (line.substr(0, 3) == "END") {
                 return true;
+
             }
         } catch (const FileError&) {
             return false;
@@ -270,7 +276,7 @@ Record get_record(const std::string& line) {
                rec == "CAVEAT" || rec == "COMPND" || rec == "EXPDTA" ||
                rec == "KEYWDS" || rec == "OBSLTE" || rec == "SOURCE" ||
                rec == "SPLIT " || rec == "SPRSDE" || rec == "TITLE " ||
-               rec == "JRNL  ") {
+               rec == "JRNL  " || rec == "TER   ") {
         return Record::IGNORED_;
     } else {
         return Record::UNKNOWN_;
--- a/src/formats/PDB.cpp
+++ b/src/formats/PDB.cpp
@@ -31,6 +31,8 @@ enum class Record {
     ATOM,
     HETATM,
     CONECT,
+    // Beggining of model.
+    MODEL,
     // End of file
     END,
     // Ignored records
@@ -86,6 +88,9 @@ void PDBFormat::read(Frame& frame) {
         case Record::CONECT:
             read_CONECT(frame, line);
             break;
+        case Record::MODEL:
+            ++step_;
+            break;
         case Record::END:
             goto end; // We have read a frame!
         case Record::IGNORED_:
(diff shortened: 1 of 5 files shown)


src/formats/PDB.cpp
```diff
@@ -261,7 +264,7 @@ Record get_record(const std::string& line) {
                rec == "CAVEAT" || rec == "COMPND" || rec == "EXPDTA" ||
                rec == "KEYWDS" || rec == "OBSLTE" || rec == "SOURCE" ||
                rec == "SPLIT " || rec == "SPRSDE" || rec == "TITLE " ||
-               rec == "JRNL  ") {
+               rec == "JRNL  " || rec.substr(0, 5) == "MODEL" ) {
```
Luthaf: You should be able to test for `"MODEL "` (with a space) directly here

Luthaf:
> If write a 10 frames trajectory, this is how my PDB file is ending:

I am not sure I understand your problem here. Could you show the code you use for writing the trajectory?

The issue is that there is no END record at the end of the file ?

src/formats/PDB.cpp
pgbarletta: yeah, I don't know why I did that.

pgbarletta:
> The issue is that there is no END record at the end of the file ?
Yeap. And the fact that because of the way `done()` is implemented, I end up iterating an extra step:
```
#include "chemfiles.hpp"
#include  <iostream>
#include  <fstream>


int main(int argc, char *argv[]) {

    std::string filename = "mtn.pdb";

    chemfiles::Trajectory input_pdb_traj(filename);

    chemfiles::Trajectory salida("sal.pdb", 'w');

    while(!input_pdb_traj.done()) {
        chemfiles::Frame input_pdb_frame = input_pdb_traj.read();
        salida.write(input_pdb_frame);
    }
    return 0;
}
```

Luthaf:
And how to does the `mtn.pdb` file looks like? You can send it to me by email if you prefer not putting it on github. Else, [gists](http://gist.github.com/) are a nice way to send big-ish files through github.

Luthaf:
> Yeap. And the fact that because of the way done() is implemented, I end up iterating an extra step:

OK, I get the issue. I think you need to somehow check in [forward](https://github.com/pgbarletta/chemfiles/blob/a6d7b5c736c0461ce4eb59de5ef8f12c138c945d/src/formats/PDB.cpp#L236) that you don't have a `END` record just after an `ENDMDL` record. Else each of these will be matched as the end of a frame, thus creating an empty frame in between.

pgbarletta:
Here:

https://gist.github.com/pgbarletta/1d953df2a399a740ed661f5e34f2e0e6

Luthaf:
OK, I get the issue. I think you need to somehow check in [forward](https://github.com/pgbarletta/chemfiles/blob/a6d7b5c736c0461ce4eb59de5ef8f12c138c945d/src/formats/PDB.cpp#L236) that you don't have a `END` record just after an `ENDMDL` record. Else each of these will be matched as the end of a frame, thus creating an empty frame in between.

pgbarletta:
Thanks! Now I've got to add the `END` statement at the end.

Luthaf:
Just though of it, but you can actually add the END statemented at the end in the PDBFormat destructor, not the Trajectory destructor. You might need to check to only add it in files opened in write/append mode, and only if a frame was actually written to the file.

pgbarletta:
Did the easy thing and added a flag called `written_` . Let me know if this is OK with you.

src/formats/PDB.cpp
```diff
@@ -236,8 +242,15 @@ bool forward(TextFile& file) {
     while (true) {
         try {
             auto line = file.readline();
+
+            if (line.substr(0, 6) == "ENDMDL" &&
```
Luthaf: Maybe you can add a comment here to tell what this condition is doing ?

src/formats/PDB.cpp
```diff
@@ -403,3 +423,9 @@ void check_values_size(const Vector3D& values, unsigned width, const std::string
         );
     }
 }
+
+PDBFormat::~PDBFormat() {
+  if (written_) {
```
Luthaf: Could you please use 4 spaces for indentation?

tests/formats/pdb.cpp
```diff
@@ -127,14 +127,17 @@ TEST_CASE("Read files in PDB format"){
...
+    "MODEL    1\n"
     "CRYST1   22.000   22.000   22.000  90.00  90.00  90.00 P 1           1\n"
     "ATOM      1 A    XXX X   1       1.000   2.000   3.000  1.00  0.00           A\n"
     "HETATM    2 B    XXX X   2       1.000   2.000   3.000  1.00  0.00           B\n"
     "ATOM      3 C    XXX X   3       1.000   2.000   3.000  1.00  0.00           C\n"
     "HETATM    4 D    XXX X   4       1.000   2.000   3.000  1.00  0.00           D\n"
     "CONECT    1    2\n"
     "CONECT    2    1\n"
+    "ENDMDL\n"
     "END\n"
```
Luthaf: Why are there two END in this file ?

pgbarletta: Because of how is written:
```
    {
        auto file = Trajectory(tmpfile, 'w');
        file.write(frame);
    }
```
`file` is going out of scope twice.

src/formats/PDB.cpp
pgbarletta: done.

src/formats/PDB.cpp
pgbarletta: done.

Luthaf:
Ok, this looks good to me. Let's merge it when tests passes !

src/formats/PDB.cpp
```diff
@@ -85,6 +88,9 @@ void PDBFormat::read(Frame& frame) {
         case Record::CONECT:
             read_CONECT(frame, line);
             break;
+        case Record::MODEL:
+            ++step_;
+            break;
         case Record::END:
             goto end; // We have read a frame!
```
Luthaf: Just reading everything again, and this is bugging me: what happens if we have a PDB file like this:

```
MODEL 1
...
ENDMDL
END
MODEL 2
...
ENDMDL
END
```

Because how `Record::END` is matched, does this means that we will read two frames, the real one and an empty one ?

src/formats/PDB.cpp
```diff
@@ -85,6 +88,9 @@ void PDBFormat::read(Frame& frame) {
         case Record::CONECT:
             read_CONECT(frame, line);
             break;
+        case Record::MODEL:
+            ++step_;
```
Luthaf: Also, why are you incrementing the step here instead of after the `end:` label? Does this means that the step will not be incremented in case of file without `MODEL` record?

tests/formats/pdb.cpp
Luthaf: Ok, thanks!

src/formats/PDB.cpp
pgbarletta: Yeap. `step_` is just to keep track of the `MODEL` records.

src/formats/PDB.cpp
pgbarletta: The thing is, there should be only 1 `END` record. I was actually surprised to see one of the examples (`wat.pdb` I think?) that used the `END` record more like as an `ENDMDL`. It seems that is common practice in some chemistry fields? I've always worked with biomolecules so I wouldn't know.
I think that, if Chemfiles supports PDBs with more than 1 `END` record, fine. But if it doesn't, then that's fine too, because that's poor formatting. But, again, I'm biased by my own personal experience on my field.

Luthaf: Yes, this is an issue with file formats in chemistry: there might be a standard (and we are already lucky when there is one), and then there are the implementations and different way of doing of all the existing codes.

My usual point of view on that is that chemfiles should write files conforming to the standard as much as possible, and accept as many files as possible when reading, even non-conforming ones if possible. I think I'll try to fix this! Give me some time and I'll add a commit on this PR.

Luthaf: 869a0629b7c4b9b432708f2978aa3055bb789e73 should do the trick here. Tell me what you think.

FYI, I also changed code in the forward function, because file.readline() changes the position of the cursor in the file stream, so we need to store it and restore it afterwards using `tellg` and `seekg`.

pgbarletta: > FYI, I also changed code in the forward function, because file.readline() changes the position of the cursor in the file stream, so we need to store it and restore it afterwards using tellg and seekg.

Yeap, sorry for introducing a bug. I thought of that and then forgot to fix it.

So, correct me if I'm wrong, but now files like this one:
```
MODEL 1    // well formatted file
...
ENDMDL
MODEL 2
...
```
or this one:

```
ATOM    1 ....     // a typical chemistry PDB
...
END
ATOM    1 ....
...
```
or this one:
```
MODEL 656      // a horrendous quimera of the above
...
ENDMDL
END
MODEL 123
...
```
would result in the same `Trajectory` object. Right?
I think is great to have such a robust library. Do you want me to add warnings for the above 2 cases? Or would they become too annoying when working with PDB chemistry files?

Luthaf: >would result in the same Trajectory object. Right?

Yes, you are right.

> Do you want me to add warnings for the above 2 cases? Or would they become too annoying when working with PDB chemistry files?

We could add a warning for the last case, which is really a weird one. For the second case it might become very annoying quickly.

pgbarletta: Ah, if its too rare then I wouldn't worry about it. The second case was the one that I cared about the most, but you're right, chemistry people would hate such a warning. I think we're done!

PR #110, opened 2017-10-10
Changed max resSeq check and added same check when residue info is not avilable

pgbarletta:
Hi, me again.
Wrote a +999-atoms PDB file without residue info and noticed some strange coordinates, so I noticed that the check was not being done for these cases.
Also, resSeq has 4 characters, so max allowed number should be 9999.


Luthaf:
Looks good to me. Could you add a test, at least for the case with an explicit residue id ?

tests/formats/pdb.cpp
```diff
@@ -232,14 +235,84 @@ TEST_CASE("Write files in PDB format") {
...
+        frame.set_cell(UnitCell(1234567890));
+        CHECK_THROWS_AS(trajectory.write(frame), FormatError);
+
+        frame.set_cell(UnitCell(12));
+        frame.positions()[0] = Vector3D(123456789, 2, 3);
+        CHECK_THROWS_AS(trajectory.write(frame), FormatError);
+    }
+
+    SECTION("Default residues") {
+        auto tmpfile = "data/pdb/tmp.pdb";
```
Luthaf: You can use `auto tmpfile = NamedTempPath(".pdb");` instead, it will automatically do the cleanup for you, and will create a file in /tmp instead.

tests/formats/pdb.cpp
```diff
@@ -232,14 +235,84 @@ TEST_CASE("Write files in PDB format") {
...
+        // correctly
+        CHECK(positions[998] == Vector3D(1., 2., 3.));
+        CHECK(positions[9998] == Vector3D(4., 5., 6.));
+        CHECK(positions[9999] == Vector3D(7., 8., 9.));
+
+        remove("data/pdb/tmp.pdb");
+    }
+
+    SECTION("User specified residues") {
+        auto tmpfile = "data/pdb/tmp.pdb";
```
Luthaf: Same here

tests/formats/pdb.cpp
```diff
@@ -2,6 +2,9 @@
 // Copyright (C) Guillaume Fraux and contributors -- BSD license
 
 #include <fstream>
+
+#include <iostream>
```
Luthaf: You should not need this input file

pgbarletta: Yeah, I was just using it for checking. I deleted this line in the last commit.

tests/formats/pdb.cpp
pgbarletta: You know I tried that, and I could read the file doing:
```
    std::ifstream checking(tmpfile);
    std::string content((std::istreambuf_iterator<char>(checking)),
                         std::istreambuf_iterator<char>());
```
but, for some reason, the last lines would always be missing. And if I tried to read it through chemfiles, it would complain that the file had 0 frames in it.
I'll try again and let you know.

tests/formats/pdb.cpp
Luthaf: Ok, sorry, I did not see your last commit 😄

tests/formats/pdb.cpp
Luthaf: You need to make sure the `Trajectory` object goes out of scope before opening it with either chemfiles or `std::ifstream`: the file is synced when the destructor runs.

pgbarletta: aaaahhh, I thought that the file would be deleted if `Trajectory` goes out of scope, but that happens when the `NamedTempPath` object goes out of scope. Alright, I'll fix it

pgbarletta:
I guess the [error](https://travis-ci.org/chemfiles/chemfiles/jobs/287255323#L1129) is due to the large temporary PDB files I'm constructing?

Luthaf:
Yes, the test is trying to allocate too much memory for the emscripten runtime. I think the simplest fix is to compile with `-s ALLOW_MEMORY_GROWTH=1`

PR #124, opened 2017-10-25
Fix pdb MODEL reading

### LINKED ISSUE #103: PDB format: write as models

Luthaf:
Fix #103 

The issue was that the `line` variable was overwritten in forward when checking for consecutive `ENDMDL/END` lines. This meant that the number of steps as seen by the Trajectory class was off, thus creating the errors.

@pgbarletta, if this looks good to you, I'll merge it

### DIFF
--- a/src/formats/PDB.cpp
+++ b/src/formats/PDB.cpp
@@ -262,9 +262,9 @@ bool forward(TextFile& file) {
 
             if (line.substr(0, 6) == "ENDMDL") {
                 auto position = file.tellg();
-                line = file.readline();
+                auto next = file.readline();
                 file.seekg(position);
-                if (line.substr(0, 3) == "END") {
+                if (next.substr(0, 3) == "END") {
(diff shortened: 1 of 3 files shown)


pgbarletta:
Great. Thanks for taking care of this.

Issue #38, opened 2016-09-22
Heuristic to guess residues in a topology


Luthaf:
In the same spirit as the `guess_bonds` function, we could try to guess the residues in a topology. The simplest version could simply assume residues == molecule, and just follow the bonds graph. A more elaborated version could try to match sub-graphs to known residues like amino-acids.

Luthaf:
I now think this is out of scope for chemfiles itself. It can be implemented on top of it if someone is interested!

PR #141, opened 2018-01-11
Guess molecule ids from bonds in LAMMPS Data

Luthaf:


Luthaf:
Ref #38, the same algorithm could be used to generate residues as the list of molecules

PR #142, opened 2018-02-13
MMTF Support

frodofine:
I've implemented support the MMTF file format (https://mmtf.rcsb.org/) using the mmtf-c parser. This adds an additional two dependencies, mmtf-c and msgpack-c.  Unfortunately, mmtf-c does not support writing to MMTF files, so I've only implemented reading support.


src/formats/MMTF.cpp
```diff
@@ -0,0 +1,118 @@
...
+    atomSkip_(0) {
+}
+
+size_t MMTFFormat::nsteps() {
+    return file_->numModels;
+}
+
+void MMTFFormat::read_step(const size_t step, Frame& frame) {
+
+    if (step >= nsteps()) {
```
Luthaf: This is already checked by the [Trajectory](https://github.com/chemfiles/chemfiles/blob/deb490b7aaef9c7fddf77bb3ee371b7d71f4d223/src/Trajectory.cpp#L64-L69) class

src/formats/MMTF.cpp
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    modelIndex_ = 0;
+    chainIndex_ = 0;
+    groupIndex_ = 0;
+    atomIndex_  = 0;
+    atomSkip_   = 0;
+
+    // Fast-forward, keeping all indecies updated
+    while(modelIndex_ != step) {
```
Luthaf: Does this work if I call `read_step(10)` and then `read_step(2)`? Or should the code reset the position in the file ?

Luthaf:
Concerning Appveyor and Travis failures, you should be able to update the test data SHA now that https://github.com/chemfiles/tests-data/pull/2 is merged. This should fix Travis.

On Windows (Appveyor), the code does not link due to `unresolved external symbol __imp_ntohl referenced in function assign_bigendian_4` errors. Is there some kind of dll_export define symbol that is needed to build mmtf-c on windows ?

src/formats/MMTF.cpp
frodofine: The MMTF-c parser reads the entire file at once, so there's no position to reset. If you have suggestions, I'm all ears.

src/formats/MMTF.cpp
frodofine: I've removed the check.

src/files/MMTFFile.cpp
```diff
@@ -0,0 +1,32 @@
...
+#include "chemfiles/utils.hpp"
+#include "chemfiles/files/MMTFFile.hpp"
+
+using namespace chemfiles;
+
+MMTFFile::MMTFFile(std::string filename, File::Mode mode):
+  File(std::move(filename), mode), handle_(MMTF_container_new()) {
+
+    if (handle_ == nullptr) {
+        throw memory_error("Could not allocate a MMTF contain");
```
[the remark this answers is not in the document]
frodofine: Whoops, fixed.

include/chemfiles/formats/MMTF.hpp
```diff
@@ -0,0 +1,48 @@
...
+#include "chemfiles/File.hpp"
+#include "chemfiles/Residue.hpp"
+#include "chemfiles/files/MMTFFile.hpp"
+
+namespace chemfiles {
+
+/// [MMTF][MMTF] file format reader and writer.
+///
+/// For multi-frame trajectories, we follow the convention of VMD to use multiple
+/// `END` records, separating the steps.
```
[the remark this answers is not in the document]
frodofine: I've updated the comment. The convention comes from PyMOL.

include/chemfiles/formats/MMTF.hpp
```diff
@@ -0,0 +1,48 @@
...
+///f
```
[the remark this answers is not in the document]
frodofine: Fixed

external/CMakeLists.txt
```diff
@@ -135,6 +135,30 @@ set(VMD_MOLFILE_PLUGINS
...
+    MSGPACK_ENABLE_CXX
+    MSGPACK_ENABLE_SHARED
+)
+
+# ==========
+# mmtf-c format: https://github.com/chemfiles/molfiles
+# ==========
+unpack_library(mmtf-c)
+add_library(mmtf-c OBJECT ${CMAKE_CURRENT_BINARY_DIR}/mmtf-c/src/mmtf_parser.c)
+target_include_directories(mmtf-c SYSTEM BEFORE PRIVATE ${EXTERNAL_INCLUDES})
```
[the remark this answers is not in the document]
frodofine: Changed.

src/formats/MMTF.cpp
Luthaf: So the whole file is always in memory ? In that case I'm fine with this 😃

frodofine: From my understanding, yes. The files are quite small so the parser loads the whole file at once.

frodofine:
Looks like the asm.jl build is not working properly. I may just submit a patch tomorrow that removes support for MTFF when building for asm.js. RCSB offers reference code for JavaScript anyway.

src/files/MMTFFile.cpp
```diff
@@ -11,6 +11,10 @@ using namespace chemfiles;
 MMTFFile::MMTFFile(std::string filename, File::Mode mode):
   File(std::move(filename), mode), handle_(MMTF_container_new()) {
 
+    #ifdef __EMSCRIPTEN__
```
Luthaf: I would rather put the registration in `TrajectoryFactory` inside ifdef, it feels clearer.

Luthaf:
The error with asm.js looks like a memory corruption to me:

```
due to unexpected exception with message:
  out of bounds atomic index in `Topology::add_bond`: we have 70 atoms, but the
  bond indexes are 1668509099 and 70
Error in MMTF_parser_decode_apply_strategy: 1543503872 does not refer to any strategy.
Error in MMTF_parser_decode_apply_strategy: 1959025664 does not refer to any strategy.
Error in MMTF_parser_decode_apply_strategy: 975962112 does not refer to any strategy.
Error in MMTF_parser_decode_apply_strategy: -989376512 does not refer to any strategy.
Error in MMTF_parser_decode_apply_strategy: -617349120 does not refer to any strategy.
Error in MMTF_parser_decode_apply_strategy: -975496192 does not refer to any strategy.
Error in MMTF_parser_decode_apply_strategy: 335544320 does not refer to any strategy.
Error in MMTF_parser_decode_apply_strategy: 1959007232 does not refer to any strategy.
Error in MMTF_parser_decode_apply_strategy: -989772800 does not refer to any strategy.
Error in MMTF_parser_decode_apply_strategy: 1275068416 does not refer to any strategy.
```

But the valgrind build looks fine, so I think we can merge this and have a more detailed look on the asm.js build later.

Concerning your `Special version of msgpack-c` commit, is the corresponding source available somewhere ? I would like to put it in a repository as `chemfiles/msgpack-c`, to ensure the source is always available.

Before I merge this, could you also add a bit of doc in `doc/src/formats.rst` for MMTF ? And if you want to rebase and squash commits, feel free to do so =)

frodofine:
Please see https://github.com/frodofine/msgpack-c for my modified version of msgpack-c. I just removed the parts that conflict with having it embed in chemfiles (not well mind you, more can be removed). The version of mmtf-c I used is vanilla.

I think the issues with asm.js come from msgpack-c or mmtf-c. These libraries do some bit level magic that sometimes involves asm instructions (the ntohl functions from before are actually Intel x86 instructions for converting big endian to little endian). Unfortunately, I am not familiar enough with asm.js to know what exactly is going wrong, but I think the issue lies somewhere in this region of code. I will be attending a conference next week, so I will not have time to look deeper into it until later this month (this is something I am doing in my free time anyway).

Luthaf:
Thank you again for the PR, this is very appreciated =)

frodofine:
Thank you for the well written code to work off of!

PR #143, opened 2018-02-21
MMTF follow-up

Luthaf:
Ping @frodofine, are you OK with appearing in AUTHORS? I used your git email, is it fine ?


frodofine:
This is fine. Thank b you for the honor!

PR #145, opened 2018-02-27
Initial SDF Support

frodofine:
Still need to add support for atomic properties, which are supported by this format.

I'll update the test URL once it's been merged.

### DIFF
--- a/src/formats/SDF.cpp
+++ b/src/formats/SDF.cpp
@@ -1,0 +1,258 @@
+// Chemfiles, a modern library for chemistry file reading and writing
+// Copyright (C) Guillaume Fraux and contributors -- BSD license
+
+#include <cassert>
+#include <sstream>
+
+#include <fmt/format.h>
+#include <fmt/ostream.h>
+
+#include "chemfiles/formats/SDF.hpp"
+
+#include "chemfiles/ErrorFmt.hpp"
+#include "chemfiles/File.hpp"
+#include "chemfiles/Frame.hpp"
+#include "chemfiles/utils.hpp"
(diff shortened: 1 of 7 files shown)


src/formats/SDF.cpp
```diff
@@ -0,0 +1,258 @@
...
+    } catch (const FileError&) {
+        // We could not read the lines from the file
+        throw format_error(
+            "not enough lines in '{}' for SDF format", file.filename()
+        );
+    }
+
+    // Search for ending character, updating the position in the file only
+    while (!file.eof()) {
+        if (file.readline() == "$$$$") break;
```
Luthaf: Should this condition also check for `M END`, as `read`?

frodofine: No, the only delimiter between compounds is `$$$$` (this is the difference between SDF and MOL as MOL is single compound only).

 `M  END` denotes the end of the atom property block (which I don't use). Although the format guarantees the presence of `M  END`, some older programs don't write it, so I am being safe and only checking for `$$$$`, which MUST be present if the program supports the SDF format, otherwise, something went really wrong with writing the file.

Luthaf: OK, thanks! Does this means that the SDF reader can open MOL files ?

frodofine: Technically yes, that's why the `forward` function returns true at the end. A `MOL` file would just end, but `SDF` requires the presence of `$$$$` at the end of the file. In fact (to my understanding), `SDF` is just a wrapper around `MOL`/`MDL` to support more properties and multiple compounds.

Issue #83, opened 2017-05-05
Mathematical expressions in selections


lscalfi:
Similarly to issue #49, we could add mathematical expressions in the selection.
For example, if I want all the O atoms in the cylinder of radius 5, I could write something like:
`atoms: name O and math(x * x + y * y) < 25 `

Luthaf:
implemented in #146

Issue #73, opened 2017-04-03
Selections within a selection


lscalfi:
It is for now possible to select atoms, bonds, angles or dihedrals on the base of their components' name, type, coordinates...
It could be interesting to select atoms based on their bonded neighbours. We could create a selection inside a selection, for example:
"atoms: name O and bonded to {name H and x<5}"

Luthaf:
Why is this needed exactly? Or putted an other way, why do we need this?

In cfiles, it is currently not possible to use selections with more than 2 atoms in rdf for examples, but nothing prevent us from allowing selections with 3 or 4 atoms, and then only using the first two. This would allow to write selections like: `three: name(#1) O and name(#2) H and name(#3) Si and bonded(#1, #3) and bonded(#1, #2)`.

Do you see other use case for this?

Adding a bonded selector would be nice in all the cases.

Issue #72, opened 2017-03-21
Accept more selections strings 


Luthaf:
Currently, selection string like `pairs:name(#1)==H` are not accepted because the code require spaces as separators: `pairs: name(#1) == H`. But the form without space is not ambiguous to parse, so we should accept it.

This can be a good first bug to work with the selections code. Implementing this means modifying the [`split_selection`](https://github.com/chemfiles/chemfiles/blob/09d067a1841ede7aeb14925bf1b8921696d4feca/src/selections/lexer.cpp#L97) function to make it split string before and after punctuation: `:()!=<>`.

The only gotcha is that `name==H` should not parse as `[name, =, =, H] ` but as [name, ==, H].

I am willing to mentor anyone wanting to work on this. If you are interested, please comment here or on [gitter](https://gitter.im/chemfiles/chemfiles).

Luthaf:
This can be implemented by reading the selection string one char at the time, and then switching on the char. Differentiation between `<` and `<=` can be achieved by checking the next char.

PR #146, opened 2018-03-08
Selections v2

### LINKED ISSUE #136: Double not in selection does not parse
There is a bug in the shunting yard algorithm implementation that makes it fail on double negation (`not not name H`). Adding parenthesis makes is work (`not (not name H)`).

I've traced this to `not not index == 1` coming out of the shunting-yard algorithm as `not == 1 index not` instead of the expected `not not == 1 index`.

Luthaf:
This has been maturing for quite a while now, and I think it is ready to be merged. The core of this PR is a rewrite of the selections parser as a recursive descent parser, which should allow more liberty in the selections language.

This PR implement #83, fix #136, and open the road for #73.

All previously parsing selections are still accepted, and two more cases are available

- mathematical expressions (`x(#1) / 4 < sin(3 ^ 8)`)
- checking multiple values for equality (`name O H S` as a shorthand for `name O or name H or name S`)

TODO before merging:
 
- [x] documentation
- [x] improve lexer w.r.t. whitespaces (#72)

For the record, here is the EBNF-like grammar I used for the parser

```ebnf
selection = [context ":"] expression.
context = "atoms"
        | "one"
        | "pairs"
        | "bonds"
        | "two"
        | "angles"
        | "three"
        | "dihedrals"
        | "four".

expression = selector [{('or' | 'and') selector}].

selector = "all" | "none" | str_selector | math_selector | bool_selector | "not" expression | "(" expression ")".

str_selector = str_property ?string? [{?string?}]
             | str_property ("==" | "!=") ?string?.

math_selector = num_property ?number? [{?number?}]
              | sum binop sum.

binop = "==" | "!=" | "<" | "<=" | ">" | ">=".
sum = product [{('+' | '-') product}].
product = power [{('/' | '*') power }].
power = value ['^' power].
value = function | num_property | "(" sum ")" | ?number? | "+" value | "-" value .
function = ("cos" | "sin" | "sqrt") "(" sum ")".

(* unimplemented *)
bool_selector = bool_propname ["(" var [{"," var}] ")"].
bool_propname = "isbond" | "isangle" | "isdihedral".

num_property = num_propname ["(" var ")"].
num_propname = "mass" | "index" | "resid" | "x" | "y" | "z" | "vx" | "vy" | "vz".

str_property = str_propname ["(" var ")"].
str_propname = "name" | "type" | "resname".
var = "#" (1 | 2 | 3 | 4).
```


Luthaf:
@frodofine, do you have some time to give this PR a look ? The diff is pretty huge, but I tried to separated into smaller(-ish) commits. I would specifically appreciate comments on the documentation side of things. 

If you don't have time, this is fine =) The previous tests still passes, and new ones have been added!

frodofine:
I'll give a look later this week, I'm busy at the moment though.

doc/src/selections.rst
```diff
@@ -27,15 +27,22 @@ of three bonded atoms forming an angle such that the name of the second atom is
...
-(more or equals).
+equals); ``<`` (less than), ``<=`` (less or equals); ``>`` (more than); and
+``>=`` (more or equals).
+
+Numeric values produced either by numeric selectors (``x``; ``mass``, ...) or
+literal values (``5.2``, ``22.21e-2``) can also be combined together using
+mathematical operations: the usual ``+``, ``-``, ``*`` and ``/`` operators are
+supported, as well as ``^`` for power raising. Some mathematical functions are
+also supported. This functionality can be used to check if a given atom is
+inside a given region of space. For example, for a sphere of radius 15 A:
```
frodofine: I'd be sure to mention units more explicitly. If everything is in Angstroms for distance operations, then you should let the user know. GROMACS is very specific in that it uses nm for example.

doc/src/selections.rst
```diff
@@ -27,15 +27,22 @@ of three bonded atoms forming an angle such that the name of the second atom is
...
 ``mass(#2) < 4``. The comparison operators are ``==`` (equals); ``!=`` (not
-equals); ``<`` (less than), ``<=`` (less or equals); ``>`` (more than); and ``>=``
-(more or equals).
+equals); ``<`` (less than), ``<=`` (less or equals); ``>`` (more than); and
+``>=`` (more or equals).
+
+Numeric values produced either by numeric selectors (``x``; ``mass``, ...) or
+literal values (``5.2``, ``22.21e-2``) can also be combined together using
+mathematical operations: the usual ``+``, ``-``, ``*`` and ``/`` operators are
+supported, as well as ``^`` for power raising. Some mathematical functions are
```
frodofine: I prefer "exponents" over "power raising," but that's a personal preference.

doc/src/selections.rst
```diff
@@ -72,6 +79,11 @@ Additional selectors ideas are welcome if they cannot be created easily from a
 combinaison of the above. If you don't know if your selector idea can be
 created from a combinaison, just ask!
 
+Supported functions in mathematical expressions are: ``sin``, ``cos``, ``tan``
+for the trigonometric functions; ``asin`` and ``acos`` reciprocal trigonometric
```
frodofine: I'm confused. By reciprocal do you mean CSC, SEC, and COT? You mention the Arc trig functions by name, so I'm a bit lost.

frodofine: You also don't mention atan. This function is a bit tricky as there are two implementations of it in C++ ( tan() and tan2()) that behave differently. It may be good to mention which one you've used specifically.

doc/src/selections.rst
```diff
@@ -27,15 +27,22 @@ of three bonded atoms forming an angle such that the name of the second atom is
...
 ``#4`` variables: ``name(#3)`` will give the name of the third atom, and so on.
 Then we compare the selector result for a given atom with a value, either from
 another selector --- ``name(#1) != name(#2)`` --- or a literal value ---
 ``mass(#2) < 4``. The comparison operators are ``==`` (equals); ``!=`` (not
-equals); ``<`` (less than), ``<=`` (less or equals); ``>`` (more than); and ``>=``
-(more or equals).
+equals); ``<`` (less than), ``<=`` (less or equals); ``>`` (more than); and
+``>=`` (more or equals).
+
+Numeric values produced either by numeric selectors (``x``; ``mass``, ...) or
```
frodofine: Do you have a section that mentions all the numeric selectors? It could be useful to add a table with all the possible selectors.

doc/src/selections.rst
```diff
@@ -83,9 +95,14 @@ equivalent to ``atoms: name == H``.
 Then, if no context is given, the ``atoms`` context is used. This make ``atoms:
 name == H`` equivalent to ``name == H``.
 
-And finally, the ``==`` comparison operator is the default one if no operator is
+Then, the ``==`` comparison operator is the default one if no operator is
```
frodofine: May I suggest:

"If no comparison operator is given, the ``==`` operator is used by default."

frodofine: Also, does this mean I can use other relations like `>`? How does this work for strings?

doc/src/selections.rst
```diff
@@ -83,9 +95,14 @@ equivalent to ``atoms: name == H``.
...
 precised. This means that we can write ``name H`` instead of ``name == H``.
 
+Finally, multiple values are interpreted as multiple choices. A selection like
```
frodofine: Remove "finally" as you did previously. You can place it at the beginning of the next paragraph instead of 'At the end' because this paragraph is a culmination of all the selection rules you've discussed previously.

doc/src/selections.rst
frodofine: Maybe also mention the order of operations here. Is 2+3*3 evaluate to 11 or 18?

Luthaf: 2+3*3 evaluate to the expected 11 (and not 15, nor 18 ^^) here, but yes I can mention it 😃

doc/src/selections.rst
Luthaf: > By reciprocal do you mean CSC, SEC, and COT? You mention the Arc trig functions by name, so I'm a bit lost.

For me, reciprocal means the one going from the sin value to the angle. How would you call them?

> You also don't mention atan. This function is a bit tricky as there are two implementations of it in C++ ( tan() and tan2()) that behave differently. It may be good to mention which one you've used specifically.

Yeah, `atan` is tricky, so I left it out 😃

doc/src/selections.rst
Luthaf: It is bellow, but not separated in string/numeric selectors. I'll do this.

doc/src/selections.rst
frodofine: To me a reciprocal trig function is defined like so:

`sec(a) = 1 / cos(a)` or `csc(a) = 1/csc(a)`. See (https://www.khanacademy.org/math/geometry-home/right-triangles-topic/reciprocal-trig-ratios-geo/a/reciprocal-trig-ratios) for details (Khan academy can explain it better than I can).

What you describe is an inverse trig function (sometimes called an Arcus trig function, hence the 'a' in front of the name). They are defined as the graph of `x = sin(y)`, or `y=arcsin(x)` - whichever you prefer. See (https://en.wikipedia.org/wiki/Inverse_trigonometric_functions) for details.

Luthaf:
I updated the last commit, hopefully the documentation is easier to understand now. Let me know what you think !

doc/src/selections.rst
```diff
@@ -20,22 +20,32 @@ one, two, three or four independent atoms; and ``bonds``, ``angles`` and
...
-coordinate of this atom. Other selectors can be applied to two or more atoms,
-like ``distance`` or ``angle``. When using a selection with more than one atom,
-selectors refers to the different atoms with the ``#1``, ``#2``, ``#3`` or
-``#4`` variables: ``name(#3)`` will give the name of the third atom, and so on.
-Then we compare the selector result for a given atom with a value, either from
-another selector --- ``name(#1) != name(#2)`` --- or a literal value ---
-``mass(#2) < 4``. The comparison operators are ``==`` (equals); ``!=`` (not
-equals); ``<`` (less than), ``<=`` (less or equals); ``>`` (more than); and ``>=``
-(more or equals).
+that are evaluated for each atom, and return either ``true`` if the atom match,
```
frodofine: matches, not match

doc/src/selections.rst
```diff
@@ -20,22 +20,32 @@ one, two, three or four independent atoms; and ``bonds``, ``angles`` and
...
+or ``false`` if it does not. There are three kinds of selectors:
+
+- global selectors are ``all`` that match all atoms; and ``none`` that match no
+  atom;
+- string selectors compare string values with one of ``==`` (equal) or ``!=``
+  (not equal). One can either compare two atomic properties (``name(#1) ==
+  type(#2)``) or atomic properties to literal strings (``name(#1) != He``);
+- numeric selector compare two numeric values with either ``==``, ``!=``, ``<``
```
frodofine: numeric selector**s**

doc/src/selections.rst
```diff
@@ -20,22 +20,32 @@ one, two, three or four independent atoms; and ``bonds``, ``angles`` and
...
+  (less than), ``<=`` (less or equal), ``>`` (more than), and ``>=`` (more or
+  equal)
+
+Numeric values are produced by numeric selectors (``x``; ``mass``, ...) or
+literal values (``5.2``, ``22.21e-2``). They can also be combined together using
+mathematical operations: the usual ``+``, ``-``, ``*`` and ``/`` operators are
+supported, as well as ``^`` for exponentiation. These operations follow the
+usual priority rules: ``1 + 2 * 3`` is 7, not 9.
```
frodofine: Great!

doc/src/selections.rst
```diff
@@ -20,22 +20,32 @@ one, two, three or four independent atoms; and ``bonds``, ``angles`` and
...
+
+When using a selection with more than one atom, selectors must refers to the
```
frodofine: selectors must refer (no s).

doc/src/selections.rst
```diff
@@ -44,48 +54,55 @@ false and false to true. ``name(#1) == H and not x(#1) < 5.0`` and ``(z(#2) < 45
...
+- Then if no comparison operator is given, ``==`` is used by default. This means
+  that we can write ``name H`` instead of ``name == H``.
+- Then, multiple values are interpreted as multiple choices. A selection like
+  ``name H O C`` is expanded into ``name H or name O or name C``.
 
 At the end, using all these elisions rules, ``atom: name(#1) == H or name(#1) ==
-O`` is equivalent to ``name H or name O``; and ``bond: name(#1) == C and
-mass(#2) == 4.5`` is equivalent to ``bond: name(#1) C and mass(#2) 4.5``.
+O`` is equivalent to ``name H O``. A more complex example can be ``bonds:
+name(#1) O C and index(#2) 23 55 69``, which is equivalent to ``bonds:
```
frodofine: Is there are a property like 'numbonds' for the number of bonds a given atom has?
(Sorry, this is an awkward place for this - just thought of it while reading).

doc/src/selections.rst
```diff
@@ -44,48 +54,55 @@ false and false to true. ``name(#1) == H and not x(#1) < 5.0`` and ``(z(#2) < 45
...
-- ``resid``: select atoms based on their residue name. An atom without residue
-  information will never be selected.
-- ``mass``: select atoms based on their mass;
-- ``x``, ``y`` and ``z``: select atoms based on their position cartesian components;
-- ``vx``, ``vy`` and ``vz``: select atoms based on their velocity cartesian components;
-
-Additional selectors ideas are welcome if they cannot be created easily from a
-combinaison of the above. If you don't know if your selector idea can be
-created from a combinaison, just ask!
+List of implemented properties
```
frodofine: I might open a separate issue for this - but maybe check the atom properties if the selector is unknown. For example, I could select all atoms with a property I've assigned specifically.

doc/src/selections.rst
Luthaf: No, not yet. But this could be added =) It might be a bit expensive to compute due to the way bonds are stored currently.

doc/src/selections.rst
Luthaf: Yes, this could be nice ! I would rather have a specific function for this:  `property("is_hetatm", #1)` or `prop("is_hetatm", #1)`; as property names might contain spaces.

Luthaf:
I'll merge this tomorrow if everyone is OK with it!

PR #147, opened 2018-03-11
Residue connectivty

### LINKED ISSUE #144: Residue connectivy not reflected in topology
The connectivity of residues to one another (for example the peptide bond between two amino acids in a protein) is not assigned for valid PDBs. Currently, bonds are assigned based on a distance criterion which may disrupt ligand connectivity (especially with metals). Information in some crystallographic file formats, such as CIF, provide this information by specifying a bond type (like  'L-PEPTIDE LINKING' for a protein).

One solution for assignment of residue connectivity is to leverage this information, along with the atom name, to assign interresidue bonds. For example, the atom name 'C' is always bond to atom name 'N' in a protein PDB. Implementing this should not be difficult, assuming it is desirable.

frodofine:
closes #144


include/chemfiles/Topology.hpp
```diff
@@ -255,6 +255,18 @@ class CHFL_EXPORT Topology final {
...
+    ///
+    /// The connectivity is assigned based on the maps in pdb_connectivity.hpp,
+    /// you may edit this file for your own means, but by default this table
+    /// contains the 20 'natural' L-Amino acids and the nucleic acids commonly
+    /// found in DNA and RNA (dA, T, dG, dC, A, U, G, C).
+    ///
+    /// @throw Error if an atom name is not in the table.
+    ///
+    /// @example{tests/doc/frame/link_standard_residue_bonds.cpp}
+    void link_standard_residue_bonds();
```
Luthaf: I don't know if adding a public methods is the best solution here. The standard residue connectivity only concern PDB related formats, right ? (MMTF, PDB, mmcif) Then, why not have it as an helper function for these formats, and automatically add these bonds?

scripts/pdb_connectivity.py
```diff
@@ -0,0 +1,107 @@
+#!/usr/bin/env python
+# -* coding: utf-8 -*
+
+"""
+This script reads the standard connectivty table provided by
+the PDB and produces a C++ header file that maps residue names
+to their connectivity by atom name.
+
+WARNING: The entire PDB dictionary is quite large, so
```
Luthaf: Missing the end of the sentence ?

scripts/pdb_connectivity.py
```diff
@@ -0,0 +1,107 @@
...
+    if len(sys.argv) < 3:
+        usage()
+        sys.exit(0)
+
+    if len(sys.argv) >= 4:
+        residues = read_residues(sys.argv[1], sys.argv[3:])
+    else:
+        residues = read_residues(sys.argv[1])
+
+write_elements(sys.argv[2], residues)
```
Luthaf: This should be in the `if __name__ == "__main__` block.

scripts/pdb_connectivity.py
```diff
@@ -0,0 +1,107 @@
...
+    f.write(MAP)
+    for residue in residues:
+        f.write("    " + str(residue) + ",\n")
+    f.write("};\n\n")  # closing the map.
+    f.write("} // namespace chemfiles\n\n")  # closing the namespace.
+    f.write("#endif\n")  # closing the header guard
+
+
+def usage():
+    print(sys.argv[0] + " path/to/het_dictionary.txt pdb_connectivity.hpp [Residue names]")
```
Luthaf: Could you add which residues where used to generate a `pdb_connectivity.hpp` file in a comment in the file ?

src/Topology.cpp
```diff
@@ -134,3 +135,85 @@ optional<const Residue&> Topology::residue_for_atom(size_t index) const {
...
+            const auto& entry = residue_table->second.find(key_val.first);
+            if ( entry == residue_table->second.end() ) {
+
+                // A special case missed by the standards committee????
+                if ( key_val.first == "HO5'" ) {
+                    add_bond(key_val.second, atom_name_to_index["O5'"]);
+                    continue;
+                }
+
+                throw error("{}_{} contains an invalid name: {}", res.name(), *(res.id()), key_val.first);
```
Luthaf: Are we sure we have a res.id() here ? Else this is undefined behaviour.

Luthaf:
The emscripten failure on Travis is weird, if it stays here I'll give it a look locally.

scripts/pdb_connectivity.py
frodofine: Whoops

include/chemfiles/Topology.hpp
frodofine: I thought about that as that was my original solution, but decided it might more helpful to allow the software built around Chemfiles to decide if this is necessary.  It allows one to skip this step if necessary or use it for other means like peptide building.

Luthaf: But then the software in question must know about this function and decide when to call it. Concerning peptide building, I would imagine that people/software using chemfiles would add the bonds "manually" nevertheless.

I prefer to keep the API as simple as possible (partly because this makes chemfiles easier to use, and partly because more functions means more work, as they have to be mirrored to the C API and all the other languages bindings).

Do you see any other use case that would justify adding a new public function?

frodofine: That's true, but I see this function as somewhat of an analog to Frame::guess_topology (I don't need position data though, so it's in topology), in the sense that it assigns the topology of a residue automatically.

In the software I am developing (that uses Chemfiles for parsing), the logic flow is like this:

Decide if we need to do bond assignment, if so - try to use the standard connectivity and check again. If all else fails, use guess topology. I'm not using the CAPI, and you shouldn't let my one use case dictate decisions, but this is what is guiding my decisions.

Luthaf: I would still prefer this functionality to be part of the PDB/MMTF parsing. That way, you can still get good(-ish) topology to work with in your code (and call guess_topology if it is needed).

If there is more demand on exposing this functionality (we could open an issue to discuss this and ask feedback), it should be easy to extract the code into a separated method in the future.

---

Sorry for being picky here, I try to balance functionality, usefulness to existing users and ease of learning and usage for new users, and I dont' always see every possible use case. Please tell me if I'm getting to annoying!

frodofine: It's not required for MMTF parsing as this format lists bonds explicitly. Thus, it's only required for PDB (and CIF?). I can move it to the PDB format sometime this week in any case.

I'm contributing to YOUR project :smile:, so it's your decision in the end.

Luthaf: Well I created the project, but I would love to see it grow enough to no longer be MY project :smiley:

frodofine:
Interesting, converting the `std::map< std::string, std::vector<std::string> >` to a `std::multimap` fixed Travis.

Luthaf:
> Interesting, converting the std::map< std::string, std::vector<std::string> > to a std::multimap fixed Travis.

That's weird ... But good idea using std::multimap here!

---

BTW, you don't have to squash all your commits together all the time. It can make the history nicer to squash "review commits" with the base one (`fixed foo bar`), but there is no need to squash everything. Ideally, I try to stick with one commit = one change; but there are no hard rules here :smile:

Luthaf:
Could you add the corresponding entry in the CHANGELOG ?

frodofine:
I actually didn't intentionally squash them! Since I was moving all the code from `Topology.cpp` to 'PDB.cpp`, I started over from the master branch and copied all my code into that. I changed the algorithm significantly, so I don't know how much the old code would have been useful anyway.

Ideally, I would have treated MMTF differently - but I'm lazy.

PR #148, opened 2018-03-12
Residue properties

frodofine:
Maybe this is better as an atom property, but the chainid is generally regarded as a residue property (for example, how the MMTF data structure is ordered).

### DIFF
--- a/include/chemfiles/capi/atom.h
+++ b/include/chemfiles/capi/atom.h
@@ -188,7 +188,7 @@ CHFL_EXPORT chfl_status chfl_atom_set_property(
 
 /// Get a property with the given `name` in this `atom`.
 ///
-/// This function returns `NULL` is no property exist with the given name.
+/// This function returns `NULL` if no property exists with the given name.
 ///
 /// The user of this function is responsible to deallocate memory using the
 /// `chfl_property_free` function.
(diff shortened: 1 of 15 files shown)


Luthaf:
This looks good to me! Having Residues properties as well as atomic and frame properties make complete sense.

Could you add the corresponding C API functions (`chfl_residue_set_property`, `chfl_residue_get_property`) ?

frodofine:
Sure thing.

include/chemfiles/capi/atom.h
```diff
@@ -188,7 +188,7 @@ CHFL_EXPORT chfl_status chfl_atom_set_property(
 
 /// Get a property with the given `name` in this `atom`.
 ///
-/// This function returns `NULL` is no property exist with the given name.
+/// This function returns `NULL` if no property exists with the given name.
```
Luthaf: thanks! I am not a native speaker, so there might be more of them hiding around

Luthaf:
Just though of this: could you also add you changes to the CHANGELOG.md ? Both new formats (SDF and MMTF) and this ?

include/chemfiles/capi/atom.h
frodofine: Not a problem, I'll fix 'em as I see 'em.

Luthaf:
Thanks! I'll wait for Travis and Appveyor tests to pass, and then I'll merge this.

Issue #150, opened 2018-03-14
Choose a name and document which properties are written/read by formats


Luthaf:
Properties are being used more and more, and the names of existing properties have no consistency right now. On the top of my head, we have 

- Atomic properties: is_hetatm, 
- Frame properties: name (for molecular name), various properties from the SDF format
- Residue properties (#148): chainid, chainname, chainindex

We should pick a consistent scheme for naming properties (maybe something like `<FORMAT>_property`, i.e. `PDB_is_hetatm`, `MMTF_chainid`). Do you have any idea for this?

And we should document which properties are used (set/read) by each format.

Pinging @frodofine and @pgbarletta on this, as you are the main users of properties !

pgbarletta:
Assigning formats to properties could be problematic since formats share some properties (eg: chainid between MMTF and PDB). 

The `Property` class is used by both the Chemfiles library and the users, something which is not rare, but I think the problem lies in the fact that conventional users (those that haven't added a new `Property` object to her/his atoms or frames) are also obliged to know about the `Property` class to get basic info like, whether an atom is a heteroatom or not.

Given the abstract and vague nature of the "Property" class, I don't think the conventional user needs to know that these unrelated pieces of information (if an atom is a heteroatom or the ID of a polypeptide chain) are actually implemented in the same way. If I recall correctly, before calling this class `Property`, we were debating if we should call it `Information` or something else; so, it's just that, info.

I think that, given the current state of the code, the easiest way to do this would be to enclose the `get()` function within a function with a more informative name. And then, that function would call the `get()` function with the appropriate string that specifies which piece of information is being queried.
Would be something like this:
```
auto atom = traj("sample.pdb").read().topology()[0];
// Is this an hetero atom?
bool het = atom.is_hetatm()
```
And on the Chemfiles side, inside `Atom.cpp`:
```
bool is_hetatm() {
    return properties_.get("is_hetatm").value_or(false).as_bool())
}
```

Therefore, only users that have constructed their own properties will have to deal with a naming scheme, but that naming scheme would be their naming scheme!
Hope this makes sense.

Luthaf:
> Assigning formats to properties could be problematic since formats share some properties (eg: chainid between MMTF and PDB).

True, I did not thought of this.

> I think that, given the current state of the code, the easiest way to do this would be to enclose the get() function within a function with a more informative name. And then, that function would call the get() function with the appropriate string that specifies which piece of information is being queried.

[...]

PR #151, opened 2018-04-10
Iterate over all properties

pgbarletta:
From #150 

Need help with proper allocation of the `char*** names` variable that will hold the properties names. 
I can't figure out how to properly (de)allocate memory for this pointer.
I'm guessing this is the expected usage:
```
    uint64_t n;
    char const** names;
    chfl_atom_properties_names(pter_to_some_atom, &n, &names);

// do stuff w/ n and names

   free(n)
   free(names)
```

### DIFF
--- a/include/chemfiles/capi/atom.h
+++ b/include/chemfiles/capi/atom.h
@@ -194,12 +194,13 @@ CHFL_EXPORT chfl_status chfl_atom_set_property(
 ///         You can use `chfl_last_error` to learn about the error.
 CHFL_EXPORT CHFL_PROPERTY* chfl_atom_get_property(const CHFL_ATOM* const atom, const char* name);
 
-// Get a pointer to the properties names
-CHFL_EXPORT chfl_status chfl_atom_properties_names(const CHFL_ATOM* const atom, uint64_t* count, const char*** names);
-
-// Get a pointer to the properties names
-CHFL_EXPORT chfl_status chfl_atom_properties_names(const CHFL_ATOM* atom, uint64_t* count, const char*** names);
-
+/// Get a pointer to the properties names
+/// This function returns `nullptr` if there're no properties.
+///
+/// @example{tests/capi/doc/chfl_atom/property.c}
+/// @return The operation status code. You can use `chfl_last_error` to learn
+///         about the error if the status code is not `CHFL_SUCCESS`.
+CHFL_EXPORT chfl_status chfl_atom_properties_names(const CHFL_ATOM* const atom, uint64_t* count, char*** names);
 
 /// Free the memory associated with an `atom`.
 ///
--- a/src/capi/atom.cpp
+++ b/src/capi/atom.cpp
@@ -193,26 +193,31 @@ error:
     return nullptr;
 }
 
-extern "C" chfl_status chfl_atom_properties_names(const CHFL_ATOM* atom, uint64_t* count, const char*** names) {
+extern "C" chfl_status chfl_atom_properties_names(const CHFL_ATOM* const atom, uint64_t* count, char*** names) {
     CHECK_POINTER(atom);
-    auto opt_atom_properties_begin = atom->properties_begin();
+    CHECK_POINTER(count);
+    CHECK_POINTER(names);
 
-    if (opt_atom_properties_begin) {
-        auto atom_properties_begin = opt_atom_properties_begin.value();
-        auto atom_properties_end = atom->properties_end().value();
-        CHFL_ERROR_CATCH(*count = (uint64_t) std::distance(atom_properties_begin,
-            atom_properties_end););
-
-        for (size_t i = 0; i < *count; i++) {
-            (*names)[i] = atom_properties_begin->first.c_str();
-            ++atom_properties_begin;
-        }
-    }
-    else {
-        *count = 0;
-        names = nullptr;
-        return CHFL_SUCCESS;
-    }
+    CHFL_ERROR_CATCH(
+        auto property = atom->properties_begin();
+        auto properties_end = atom->properties_end();
+        auto size = static_cast<size_t>(std::distance(property, properties_end));
+        if (size == 0) {
+            *count = 0;
+            *names = nullptr;
+            return CHFL_SUCCESS;
+        } else {
+            *count = size;
+            *names = (const char**) malloc(sizeof(char**) * size);
+            if (*names == nullptr) {
+                set_last_error(
+                    "nullputr in properties in function 'chfl_atom_properties_names'.");
+                return CHFL_MEMORY_ERROR;
+            }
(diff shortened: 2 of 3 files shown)


include/chemfiles/Atom.hpp
```diff
@@ -178,6 +178,14 @@ class CHFL_EXPORT Atom final {
     /// @example{tests/doc/atom/property.cpp}
     optional<const Property&> get(const std::string& name) const;
 
+    /// Get the constant beginning iterator for the atom's properties.
+    optional<std::unordered_map<std::string, Property>::const_iterator> properties_begin() const {
```
Luthaf: We can use a `typedef/using` to define a `properties_iterator` for clarity instead of using `std::unordered_map<std::string, Property>::const_iterator`

include/chemfiles/capi/atom.h
```diff
@@ -183,7 +183,7 @@ CHFL_EXPORT chfl_status chfl_atom_atomic_number(const CHFL_ATOM* atom, uint64_t*
 /// @return The operation status code. You can use `chfl_last_error` to learn
 ///         about the error if the status code is not `CHFL_SUCCESS`.
 CHFL_EXPORT chfl_status chfl_atom_set_property(
-    CHFL_ATOM* atom, const char* name, const CHFL_PROPERTY* property
+    CHFL_ATOM* const atom, const char* name, const CHFL_PROPERTY* property
```
Luthaf: having a const here (in the declaration) is actually useless, as it will not be checked by the compiler.

include/chemfiles/capi/atom.h
```diff
@@ -200,6 +200,10 @@ CHFL_EXPORT CHFL_PROPERTY* chfl_atom_get_property(
     const CHFL_ATOM* atom, const char* name
 );
 
+// Get a pointer to the properties names
```
Luthaf: This will need an example and a bit more documentation :smiley:

src/Property.cpp
```diff
@@ -78,3 +78,21 @@ optional<const Property&> property_map::get(const std::string& name) const {
         return nullopt;
     }
 }
+
+optional<std::unordered_map<std::string, Property>::const_iterator> property_map::cbegin() const {
```
Luthaf: I don't think having an optional is really worth it here: it goes against the usual pattern in the STL

include/chemfiles/Atom.hpp
Luthaf: Also, all the function will need documentation and examples

src/capi/atom.cpp
```diff
@@ -182,7 +182,7 @@ extern "C" chfl_status chfl_atom_set_property(CHFL_ATOM* const atom, const char*
     )
 }
 
-extern "C" CHFL_PROPERTY* chfl_atom_get_property(const CHFL_ATOM* const atom, const char* name) {
+extern "C" CHFL_PROPERTY* chfl_atom_get_property(const CHFL_ATOM* atom, const char* name) {
```
Luthaf: But the const is useful (-ish) here: the compiler will check that we don't reset the pointer to another location.

src/capi/atom.cpp
```diff
@@ -200,6 +200,28 @@ extern "C" CHFL_PROPERTY* chfl_atom_get_property(const CHFL_ATOM* const atom, co
     return nullptr;
 }
 
+extern "C" chfl_status chfl_atom_properties_names(const CHFL_ATOM* atom, uint64_t* count, const char*** names) {
+    CHECK_POINTER(atom);
+    auto opt_atom_properties_begin = atom->properties_begin();
```
Luthaf: The whole function body needs to be wrapped in a `CHFL_ERROR_CATCH` block as any line might throw an exception.

src/capi/atom.cpp
```diff
@@ -200,6 +200,28 @@ extern "C" CHFL_PROPERTY* chfl_atom_get_property(const CHFL_ATOM* const atom, co
...
+
+    if (opt_atom_properties_begin) {
+        auto atom_properties_begin = opt_atom_properties_begin.value();
+        auto atom_properties_end = atom->properties_end().value();
+        CHFL_ERROR_CATCH(*count = (uint64_t) std::distance(atom_properties_begin,
+            atom_properties_end););
+
+        for (size_t i = 0; i < *count; i++) {
+            (*names)[i] = atom_properties_begin->first.c_str();
```
Luthaf: This is going to be a segfault: there is no memory allocated in `*names`

src/capi/atom.cpp
```diff
@@ -200,6 +200,28 @@ extern "C" CHFL_PROPERTY* chfl_atom_get_property(const CHFL_ATOM* const atom, co
...
+            ++atom_properties_begin;
+        }
+    }
+    else {
+        *count = 0;
+        names = nullptr;
```
Luthaf: This needs to be `*names = nullptr` to change the value outside of the function

Luthaf:
>  I'm guessing this is the expected usage:

```
    uint64_t n;
    char const** names;
    chfl_atom_properties_names(pter_to_some_atom, &n, &names);

// do stuff w/ n and names

   free(n)
   free(names)
```

We can not `free` n, as it was not created with `malloc` but allocated on the stack. And we need to `malloc` names in the function for the call to free to work.

------

The more I think about it, the more giving access to the `const PropertyMap&` feels like the right thing to do. It would allow for a way simpler C API function:

```cpp
extern "C" chfl_status chfl_atom_properties_names(const CHFL_ATOM* atom, uint64_t* count, const char*** names) {
    CHECK_POINTER(atom);
    CHECK_POINTER(count);
    CHECK_POINTER(names);
    CHFL_ERROR_CATCH(
        auto& properties = atom.properties();
        auto size = properties.size();
        if (size == 0) {
            *count = 0;
            *names = nullptr;
            return CHFL_SUCCESS;
        } else {
            *count = size;
            *names = malloc(sizeof(char**) * size);
            if (*names == nullptr) {
                // TODO: set message
                return CHFL_MEMORY_ERROR;
            }

            auto property = properties.begin();        
            for (size_t i = 0; i < properties.size(), property != properties.end(); i++, property++) {
                (*names)[i] = property->first.c_str();    
            }
        }
    )
}
```

Luthaf:
BTW, the appveyor failure is unrelated to this PR, and fixed in 3ca42536fd16490997641318d35f820a90b32db9. You can rebase on top of master to get it.

Luthaf:
Great to see you got some time to work on this! :smiley: 

The change set seems confusing, maybe because you merged chemfiles/master in your own branch. Could you rebase this on top of master and maybe create a specific branch for it ? If you need any help doing this, ask me and I'll be happy to get this going!

PR #160, opened 2018-05-31
Updated tarball

frodofine:
I've updated the version of fmt. The older version does not support GCC7 with C++11, but they have fixed the bug upstream.

See: [this issue](https://github.com/fmtlib/fmt/issues/734).


Luthaf:
Let's wait for all the CI to go green, and then merge this!

Issue #149, opened 2018-03-13
Switch to MMTF-CPP for reading/writing MMTF files


frodofine:
It looks like some work has been put into the MMTF C++ decoder/encoder (the last update was over a year ago). Upstream has yet to accept the pull request,  but the PR includes writing to MMTF files (MMTF-C does not support write support at all). 

Other benefits include a pure C++ API, which fits better with Chemfiles and the ability to include support as a header-only library (no need to include tarballs for msgpack-c and mmtf-c, the heads can be placed in external).

The largest change will take place in MMTFFile.cpp as this abstracts the encoder, but traversal will remain largely the same.

I'll work on updating Chemfiles once upstream (RCSB) has accepted the pending PR.

 - [x] Switch from mmtf-c to mmtf-cpp once Windows support and Emscripten support are merged upstream.

 - [x] Incorporate initial write support into Chemfiles. This will just write coordinates, atom types and simple group information.

 - [x] Add complete group information and full connectivity information.

PR #161, opened 2018-06-01
Initial port to MMTF-CPP

frodofine:
This addresses the first part of #149.

The only major difference is now Emscripten works. We can merge this now or when all issues are addressed.


src/FormatFactory.cpp
```diff
@@ -56,10 +56,8 @@ FormatFactory::FormatFactory() {
     this->add_format<Molfile<LAMMPS>>();
     this->add_format<Molfile<MOLDEN>>();
 
-    // The MMTF format does not work well with asm.js
-    #ifndef __EMSCRIPTEN__
-        this->add_format<MMTFFormat>();
-    #endif
+    // MMTF Format
+    this->add_format<MMTFFormat>();
```
Luthaf: Could you move this on the top together with the other non-molfiles formats ?

external/CMakeLists.txt
```diff
@@ -61,7 +61,7 @@ set(EXTERNAL_INCLUDES "")
 # fmtlib: https://github.com/chemfiles/fmt
 # ==========
 external_library(fmt)
-list(APPEND EXTERNAL_INCLUDES ${CMAKE_CURRENT_BINARY_DIR}/fmt)
+list(APPEND EXTERNAL_INCLUDES ${CMAKE_CURRENT_BINARY_DIR}/fmt/include/)
```
Luthaf: It looks like you got #160 intermixed in this PR. Could you rebase on top of master?

src/files/MMTFFile.cpp
```diff
@@ -6,18 +6,16 @@
...
 
 MMTFFile::MMTFFile(std::string filename, File::Mode mode):
-  File(std::move(filename), mode), handle_(MMTF_container_new()) {
-
-    if (handle_ == nullptr) {
-        throw memory_error("Could not allocate a MMTF container");
-    }
+  File(std::move(filename), mode), handle_() {
 
     if (mode == File::READ) {
```
Luthaf: Since write/append mode are not supported, could you throw an error in these cases ?

src/files/MMTFFile.cpp
```diff
@@ -6,18 +6,16 @@
...
-        bool result = MMTF_unpack_from_file(this->filename().c_str(), handle_);
-        if (!result) {
+        decodeFromFile(handle_, this->filename());
```
Luthaf: mmtf function are not in a namespace ?

src/formats/MMTF.cpp
```diff
@@ -64,10 +67,12 @@ void MMTFFormat::read_step(const size_t step, Frame& frame) {
 void MMTFFormat::read(Frame& frame) {
     frame.resize(0);
 
-    UnitCell cell (file_->unitCell[0], file_->unitCell[1], file_->unitCell[2],
-                   file_->unitCell[3], file_->unitCell[4], file_->unitCell[5]);
+    if (file_->unitCell.size() >= 6) {
```
Luthaf: Do you know in which cases could unitCell.size() be more than 6?

src/files/MMTFFile.cpp
frodofine: I plan on adding these eventually, but in the meantime, if we are merging this PR I can.

Luthaf: That's up to you! I am fine with merging this as it is, or waiting for write support

src/formats/MMTF.cpp
frodofine: It's not possible, I'll change it.

Luthaf: Can it be less than 6 though ? I think 1 and 3 could make sense, but I don't know what the MTTF format says on this

frodofine: It says it must be 6 or zero, it is based on the CRYST1 line from PDB files I think.  The C++ code actually checks for this [here](https://github.com/rcsb/mmtf-cpp/blob/18fa57ceee1332da4f69f2998731013db9febd50/include/mmtf/structure_data.hpp#L460) so we know if the vector is size 0 (not given) or 6.

Luthaf: ok, thanks!

src/files/MMTFFile.cpp
Luthaf: So, what are we doing? Merging this as it is or waiting for write support?

Luthaf: Answering to myself: I think I'll merge this (I just rebased on top of master) because the previous implementation is blocking me on another branch =)

PR #163, opened 2018-06-05
Iterate over properties

pgbarletta:
From #150 
Alright, this should be it. Sorry for taking so long. I've been doing this in bunches of 5 min.s Not the best way to work.

### DIFF
--- a/tests/doc/atom/property.cpp
+++ b/tests/doc/atom/property.cpp
@@ -22,5 +22,26 @@ TEST_CASE() {
     // Override the "first" property
     atom.set("first", false);
     assert(atom.get("first")->as_bool() == false);
+
+    // Iterate over all properties
+    atom.set("third", "string");
+    atom.set("fourth", Vector3D(1., 2., 3.));
+    for(auto prop_ite = atom.properties_begin(); prop_ite != atom.properties_end() ; ++prop_ite) {
+        auto kind = prop_ite->second.get_kind();
+        switch(kind) {
+            case Property::BOOL:
+                assert(prop_ite->second.as_bool() == false);
+                break;
+            case Property::DOUBLE:
+                assert(prop_ite->second.as_double() == 42.5);
+                break;
+            case Property::STRING:
+                assert(prop_ite->second.as_string() == "string");
+                break;
+            case Property::VECTOR3D:
+                assert(prop_ite->second.as_vector3d() == Vector3D(1., 2., 3.));
+                break;
+        }
+    }
     // [example]
 }


include/chemfiles/Property.hpp
```diff
@@ -200,6 +200,8 @@ inline bool operator!=(const Property& lhs, const Property& rhs) {
     return !(lhs == rhs);
 }
 
+using properties_iterator = std::unordered_map<std::string, Property>::const_iterator;
```
Luthaf: Should this be define inside the `property_map` class, as `property_map::iterator` ?

include/chemfiles/capi/atom.h
```diff
@@ -200,6 +200,15 @@ CHFL_EXPORT CHFL_PROPERTY* chfl_atom_get_property(
     const CHFL_ATOM* atom, const char* name
 );
 
+
+/// Get a pointer to an array of properties names and the number of properties.
+/// This function returns `nullptr` if there're no properties.
```
Luthaf: I would say something like

> Set `names` to point to an array of properties names, and `count` to the number of properties in the array. If there are no properties, `names` is set to `NULL`. 
>
> The `names` array is allocated with `malloc` by this function, and should be deallocated by the user with `free`.

include/chemfiles/capi/atom.h
```diff
@@ -200,6 +200,15 @@ CHFL_EXPORT CHFL_PROPERTY* chfl_atom_get_property(
...
+///
+/// @example{tests/capi/doc/chfl_atom/property.c}
+/// @return The operation status code. You can use `chfl_last_error` to learn
+///         about the error if the status code is not `CHFL_SUCCESS`.
+CHFL_EXPORT chfl_status chfl_atom_properties_names(const CHFL_ATOM* const atom, uint64_t* count, const char*** names);
```
Luthaf: Usually, the `count` parameter comes after the array parameter:

```
CHFL_EXPORT chfl_status chfl_atom_properties_names(const CHFL_ATOM* const atom, const char*** names,  uint64_t* count); 
```

src/capi/atom.cpp
```diff
@@ -200,6 +200,34 @@ extern "C" CHFL_PROPERTY* chfl_atom_get_property(const CHFL_ATOM* const atom, co
...
+        if (size == 0) {
+            *count = 0;
+            *names = nullptr;
+            return CHFL_SUCCESS;
+        } else {
+            *count = size;
+            *names = (const char**) malloc(sizeof(char**) * size);
+            if (*names == nullptr) {
+                set_last_error(
+                    "nullputr in properties in function 'chfl_atom_properties_names'.");
```
Luthaf: How about 

> Could not allocate memory for properties names in chfl_atom_properties_names

as error message ?

tests/capi/doc/chfl_atom/property.c
```diff
@@ -4,22 +4,38 @@
...
     chfl_property_free(property);
 
+    // Get property
     property = chfl_atom_get_property(atom, "this");
 
     double value = 0;
     chfl_property_get_double(property, &value);
     assert(value == -23);
 
+    // Get all properties
```
Luthaf: I would create a separated example for this function, as it is more complex than the two other ones

tests/capi/doc/chfl_atom/property.c
```diff
@@ -4,22 +4,38 @@
...
+    uint64_t count = 0;
+    const char** names;
+    chfl_atom_properties_names(atom, &count, &names);
+    CHFL_PROPERTY *same_property = chfl_atom_get_property(atom, names[0]);
+
+    value = 0;
+    chfl_property_get_double(same_property, &value);
+
+    assert(value == -23);
+    assert(!strcmp(names[0], "this"));
```
Luthaf: We need a call to `free(names)` here

tests/doc/atom/property.cpp
```diff
@@ -22,5 +22,26 @@ TEST_CASE() {
     // Override the "first" property
     atom.set("first", false);
     assert(atom.get("first")->as_bool() == false);
+
+    // Iterate over all properties
+    atom.set("third", "string");
+    atom.set("fourth", Vector3D(1., 2., 3.));
+    for(auto prop_ite = atom.properties_begin(); prop_ite != atom.properties_end() ; ++prop_ite) {
```
Luthaf: Do you think it is possible to have an API that looks like this instead ?

```cpp
for (auto& property: atom.properties()) {
    // ...
} 
```

Luthaf:
> Sorry for taking so long. I've been doing this in bunches of 5 min.s Not the best way to work.

Don't worry, nobody is in a hurry =) Don't feel obligated to work, we are doing this for fun!

Luthaf:
Hey @pgbarletta ! Do you think you'll have some time to work on this ? If you want I can take over this PR and implement it =)

pgbarletta:
Ugh, I completely forgot about this. Yes, please do, my hands are full
right now. Sorry for this.

PR #164, opened 2018-06-05
Matrix update

### LINKED ISSUE #162: More matrix operation support
While writing the GRO parser, I've come to realize there should be two features relating to `Matrix3D`.

- [x] Support for scalar operations such as `*` and `/`.

- [ ] Support for compound operators `*=`, `+=`, etc.

- [ ] Support for setting the unit cell via a matrix.

frodofine:
I decided not to implement `Matrix3D operator*=(Matrix3D)` as this requires a copy of the matrix to be made and I feel compound assignment operators should be free of coping. Other than that, I feel this closes #162.

### DIFF
--- a/include/chemfiles/types.hpp
+++ b/include/chemfiles/types.hpp
@@ -83,6 +83,11 @@ inline bool operator!=(const Vector3D& lhs, const Vector3D& rhs) {
     return !(lhs == rhs);
 }
 
+/// Negate vector
+inline Vector3D operator-(const Vector3D& lhs) {
+    return {-lhs[0], -lhs[1], -lhs[2]};
+}
+
 /// Add two vectors
 inline Vector3D operator+(const Vector3D& lhs, const Vector3D& rhs) {
     return {lhs[0] + rhs[0], lhs[1] + rhs[1], lhs[2] + rhs[2]};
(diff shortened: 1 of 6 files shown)


include/chemfiles/types.hpp
```diff
@@ -173,6 +178,57 @@ inline bool operator!=(const Matrix3D& lhs, const Matrix3D& rhs) {
     return !(lhs == rhs);
 }
 
+/// Negate matrix
+inline Matrix3D operator-(const Matrix3D& lhs) {
+    Matrix3D res;
```
Luthaf: Because the default constructor initialize the memory to zero, this might be silghtly slower than calling the constructor directly as 

```cpp
return {
   {-lhs[0][0], -lhs[0][1], -lhs[0][2]},
   {-lhs[1][0], -lhs[1][1], -lhs[1][2]},
   {-lhs[2][0], -lhs[2][1], -lhs[2][2]},
};
```

But the optimizer might be able to figure this.

frodofine: I was not aware of the ability to do that, I just followed the other operators which always created a new Matrix to return.  In any case, I'll merge and close #162 .

Luthaf: Yeah, that's fine for now =). I'll try some benchmarks if I get the time

frodofine: That's the only sure fire way to know!

Issue #154, opened 2018-04-12
Rewrite molfile plugins


Luthaf:
The [`Molfile`](https://github.com/chemfiles/chemfiles/blob/2ecb0d2f5db6270e7d88462e68c6bf24b8d9ed37/include/chemfiles/formats/Molfile.hpp#L34) readers have several issues:

- no support for read_step (see #153 for a workaround)
- no write support
- writing to stdout/stderr instead of using chemfiles::warning
- the code is rather hard to read and modify

I think it can be worth it to rewrite the corresponding formats directly in chemfiles

**Binary formats**

- [x] GROMACS .trr #265 #275
- [x] GROMACS .xtc #262 #274 
- [ ] LAMMPS/NAMD .dcd

There is a C library for reading xtc and trr files: http://www.gromacs.org/Developer_Zone/Programming_Guide/XTC_Library. Version 1.1.4 is BSD licensed.

For DCD and GRO files, the code from molfiles is BSD licensed and could be adapted and improved.

**Text formats**

- [x] GROMACS .gro
- [ ] LAMMPS trajectory
- [x] MOL2 files
- [ ] Molden files

frodofine:
I can take care of the MOL2 support, I've written a parser/writer for this format in the past.

marvinbernhardt:
This would be super cool. Especially since at the moment chemfiles does not read velocities from .trr files, which for me was the criterion not to use it.

marvinbernhardt:
There is https://github.com/wesbarnett/libxdrfile, but it is gpl, so i guess it can not be used

Luthaf:
Yes, as I was saying in gitter, GPL version of the library are out =/.

The version provided by GROMACS (http://www.gromacs.org/Developer_Zone/Programming_Guide/XTC_Library) is BSD though (the website says LGPL, but the COPYING file says is 2-clauses BSD). It is not available on github, but this is not really an issue =)

Luthaf:
> This would be super cool. Especially since at the moment chemfiles does not read velocities from .trr files, which for me was the criterion not to use it.

@marvinbernhardt The code to read TRR file, including velocity, has just been added by @ezavod, if that is still of interest to you!

marvinbernhardt:
It is! And It works for me, thanks for the implementation.

Luthaf:
Closing in favor of separate issues for each format

PR #165, opened 2018-06-06
Gro format rewrite

frodofine:
Closes part of #154.


src/FormatFactory.cpp
```diff
@@ -45,10 +45,10 @@ FormatFactory::FormatFactory() {
     this->add_format<LAMMPSDataFormat>();
     this->add_format<SDFFormat>();
     this->add_format<CSSRFormat>();
+    this->add_format<GROFormat>();
 
     // VMD molfile plugins
     this->add_format<Molfile<DCD>>();
-    this->add_format<Molfile<GRO>>();
```
Luthaf: Can you also remove the template instantiation in Molfile.cpp ?

src/formats/GRO.cpp
```diff
@@ -0,0 +1,324 @@
...
+
+    size_t natoms = static_cast<size_t>(count);
+
+    residues_.clear();
+    frame.add_velocities();
+    frame.reserve(natoms);
+    frame.resize(0);
+
+    for (const auto& line: file_->readlines(natoms)) {
+        auto resid = std::stoul(line.substr(0,5));
```
Luthaf: could you add a check that the line is long enough?

src/formats/GRO.cpp
```diff
@@ -0,0 +1,324 @@
...
+        std::string resname = line.substr(5, 5);
+
+        std::string atomname= line.substr(10,5);
+
+        // GRO files store atoms in NM, we need to convert to Angstroms
+        auto x = std::stof(line.substr(20,8)) * 10;
+        auto y = std::stof(line.substr(28,8)) * 10;
+        auto z = std::stof(line.substr(36,8)) * 10;
+
+        if (line.length() > 44) {
```
Luthaf: Same here, could you check that the line is long enough to read all velocity data

src/formats/GRO.cpp
```diff
@@ -0,0 +1,324 @@
...
+
+        auto a = std::stof(box_values[0]) * 10;
+        auto b = std::stof(box_values[1]) * 10;
+        auto c = std::stof(box_values[2]) * 10;
+
+        auto cell = UnitCell(a, b, c);
+        frame.set_cell(cell);
+    }
+
+    if (box_values.size() == 9) {
```
Luthaf: Could this be a `else if` clause instead ? Both tests are related

src/formats/GRO.cpp
```diff
@@ -0,0 +1,324 @@
...
+        }
+
+    }
+
+    std::string box = file_->readline();
+    auto box_values = split(box, ' ');
+
+    if (box_values.size() == 3) {
+
+        auto a = std::stof(box_values[0]) * 10;
```
Luthaf: Could you use `string2double` from utils.hpp instead of `std::stof`? The former check if the value is indeed a number and throw an error if not.

src/formats/GRO.cpp
```diff
@@ -0,0 +1,324 @@
...
+        auto cell = UnitCell(a, b, c);
+        frame.set_cell(cell);
+    }
+
+    if (box_values.size() == 9) {
+        auto v1_x = std::stof(box_values[0]) * 10;
+        auto v2_y = std::stof(box_values[1]) * 10;
+        auto v3_z = std::stof(box_values[2]) * 10;
+
+        assert(std::stof(box_values[3]) == 0);
```
Luthaf: Is this because of the restriction that cell matrix must be upper triangular ?

src/formats/GRO.cpp
```diff
@@ -0,0 +1,324 @@
...
+
+    for (auto& residue: residues_) {
+        frame.add_residue(residue.second);
+    }
+}
+
+static std::string to_gro_index(uint64_t i) {
+    auto id = i + 1;
+
+    if (id >= 100000) {
```
Luthaf: you could just check `if (i + 1 >= 100000)`

src/formats/GRO.cpp
```diff
@@ -0,0 +1,324 @@
...
+        warning("Too many atoms for GRO format, removing atomic id");
+        return "*****";
+    } else {
+        return std::to_string(i + 1);
+    }
+}
+
+void GROFormat::write(const Frame& frame) {
+    if (frame.get("name")) {
+        fmt::print(*file_, frame.get("name")->as_string() + "\n");
```
Luthaf: Do we want to check that `frame.get("name")` is indeed a string property ?

src/formats/GRO.cpp
frodofine: No, it comes from the format itself. For some reason there must be zeros in those locations, see [this](http://manual.gromacs.org/current/online/gro.html):

> values: v1(x) v2(y) v3(z) v1(y) v1(z) v2(x) v2(z) v3(x) v3(y), the last 6 values may be omitted (they will be set to zero). Gromacs only supports boxes with v1(y)=v1(z)=v2(z)=0. >

src/formats/GRO.cpp
frodofine: Fair enough!

src/formats/GRO.cpp
frodofine: Meant to use 68, thanks!

src/formats/GRO.cpp
frodofine: Sure thing.

src/formats/GRO.cpp
frodofine: Sure

src/formats/GRO.cpp
Luthaf: Ok, thanks!

src/formats/GRO.cpp
frodofine: Sure

.travis.yml
```diff
@@ -81,7 +81,7 @@ script:
   - mkdir build && cd build
   - $CMAKE_CONFIGURE cmake $CMAKE_ARGS $CMAKE_EXTRA ..
   - make -j2
-  - ctest -j2 --output-on-failure
+  - travis_wait 30 ctest -j2 --output-on-failure
```
Luthaf: why is this needed ?

frodofine: Travis during the Valgrind test is timing out because no output is received for 10 minutes since the PDB and GRO tests take longer than this to complete (and they take up both cores). This is to stop Travis from killing the job. I'm happy to look into other fixes for this. For example, is `--track-origins=yes` needed? I know this can increase Valgrind runtime significantly.

On my machine, both PDB and GRO tests complete in seconds, not 10 minutes - so I'm not sure why Travis takes so long.

Luthaf: Ok! I did not know what the travis_wait command did :smile: 

`--track-origins=yes` might not be needed here, and I have a local branch to improve this a bit too by not running the longest test with Travis + valgrind. I'll push it as soon as possible.

> On my machine, both PDB and GRO tests complete in seconds, not 10 minutes - so I'm not sure why Travis takes so long.

By default, the tests are not run under valgrind. And Travis machines are not particularly powerful ones.

frodofine: I meant with `-DVALGRIND=ON` used on my local machine :smiley: 

I think the issue is that Travis does not provide powerful machines, but its free so I can't complain :roll_eyes: .

`travis_wait 30` seems to have solved the issue (looking at the timing I could have used a number smaller than 30, I just wanted to be sure).

src/FormatFactory.cpp
frodofine: Whoops forgot about this.

Issue #71, opened 2017-03-15
Implement reading gzipped/bzip/bzip2 files 


Luthaf:
The TextFile class is here to abstract the used interface, so hopefully this will be as easy as implementing the TextFile interface for a GZipFile class. 

https://github.com/mateidavid/zstr looks like a nice library that provides `std::fstream`-like objects to read and write files compressed with zlib.

Luthaf:
Now that this is partly implemented, I think that we want a way to request the use of compressed files even if the file extension does not match. A simple way to do this would be to pass the compression method in the `format` argument to Trajectory constructor: `format = "LAMMPS Data/XZ"`; `format = "Tinker/GZ"`.

frodofine:
And do a string slit on the `/` character to get compression method and `format`?

One concern I have for binary formats is one would need to know if that binary format supports reading from a compressed format or not. Since we've decided this would be implemented in a per format manner, maybe the constructor for the corresponding classes should throw an error if it detects the format is compressed. Thoughts?

Luthaf:
Yeah, we might want to only support this for text formats for now. So maybe add another field in FormatInfo for text formats and checking on this.

Luthaf:
Bzip2 support was added in #284, so this is done!

PR #166, opened 2018-06-06
Implement reading and writting compressed files (.gz, .xz)

Luthaf:
This PR implement two new kinds of `TextFile`: `GzFile` uses zlib to read and write gz compressed files; and `XzFile` uses lzma to read and write xz compressed files.

The choice of which File implementation to use is made based on file extension, and extension recognition in Trajectory was modified accordingly. This means that reading `whatever.xzy.gz` will automatically use GzFile with XYZFormat, and writing to `file.pdb.xz` will use XzFile and PDBFormat.

This strategy allow to use gz/xz compression with any text format calling `TextFile::create` without any modification of the Format code.

TODO:
   - [x] Internal documentation of the File interface and implementations
   - [x] User documentation for this feature (docs + readme + changelog)
   - [x] Decide whether there should be a way to use compressed files not having the expected extension. (Could we read/write `path.pgz` as gzipped PDB ?) (EDIT: see https://github.com/chemfiles/chemfiles/issues/71#issuecomment-396976158)

Part of #71 (missing bzip/bzip2)


src/formats/SDF.cpp
```diff
@@ -122,7 +121,7 @@ void SDFFormat::read(Frame& frame) {
                 // Proper end of block
                 break;
             } // TODO: Add actual ATOM property parsing here.....
-        } catch (const FileError& e) {
+        } catch (const FileError&) {
```
frodofine: Whoops, my bad.

CMakeLists.txt
```diff
@@ -23,6 +23,8 @@ option(CHFL_CODE_COVERAGE "Enable code coverage" OFF)
 option(CHFL_USE_WARNINGS "Compile the code with warnings (default in debug mode)" OFF)
 option(CHFL_DEBUG_GLIBCXX "Use the debug version of GNU libstdcxx" OFF)
 option(CHFL_SYSTEM_NETCDF "Use the system NetCDF instead of the internal one" OFF)
+option(CHFL_SYSTEM_ZLIB "Use the system zlib instead of the internal one" OFF)
```
frodofine: Do we need CI for this? Probably not, but worth a discussion.

CMakeLists.txt
```diff
@@ -95,9 +101,10 @@ set(CHEMFILES_OBJECTS
     $<TARGET_OBJECTS:molfiles>
     $<TARGET_OBJECTS:fmt::object>
     $<TARGET_OBJECTS:tng_io>
-    $<TARGET_OBJECTS:tng_io_zlib>
     $<TARGET_OBJECTS:msgpackc-object>
     $<TARGET_OBJECTS:mmtf-c>
+    ${ZLIB_OBJECTS}
```
frodofine: This is slightly different than the other `$<TARGET_OBJECTS>`, is this intentional as `${ZLIB_OBJECTS}` can be empty?

include/chemfiles/files/BasicFile.hpp
```diff
@@ -18,12 +18,8 @@ class BasicFile final: public TextFile {
     /// `'r'` or `'a'` mode.
     BasicFile(const std::string& filename, File::Mode mode);
```
frodofine: Maybe rename to "uncompressed"/"flat" text file or the like. I'm not sure what Basic means anymore.

include/chemfiles/files/GzFile.hpp
```diff
@@ -0,0 +1,53 @@
...
+#include <streambuf>
+#include <vector>
+
+#include <zlib.h>
+
+#include "chemfiles/File.hpp"
+
+namespace chemfiles {
+
+class gzstreambuf: public std::streambuf {
```
frodofine: Is this used exclusively by `GzFile`? Can `GzFile` only support TextFiles? I'm asking as binary formats can be compressed as well so I want to make sure the naming conventions are appropriate here.

Most compressed binary formats are read by reading the uncompressed data into the buffer, as is done here, and then passing that buffered data to the appropriate library for reading. I am trying to suggest a way to make this streamlined for our purposes, or the binary formats can handle compressed data however they wish and a unified interface only exists for text data.

Maybe it would be good to make a single buffer class that holds both text and binary data and other classes can read this buffer directly or line by line in the case of a text file. This is probably above the needs of this PR though - just brain storming.

include/chemfiles/files/XzFile.hpp
```diff
@@ -0,0 +1,75 @@
...
+#include <array>
+#include <cstdio>
+
+#include <lzma.h>
+
+#include "chemfiles/File.hpp"
+
+namespace chemfiles {
+
+class xzstreambuf : public std::streambuf {
```
frodofine: See above.

CMakeLists.txt
Luthaf: Yes, this is the same strategy we already use for netcdf

include/chemfiles/files/BasicFile.hpp
Luthaf: Good idea =)

CMakeLists.txt
frodofine: Fair enough.

include/chemfiles/files/BasicFile.hpp
frodofine: I like 'Flat', that's at least my lab uses to describe files that are not compressed and not binary.

CMakeLists.txt
Luthaf: I don't know. We have one builder on Travis testing with the system NetCDF, we could use it with the system zlib and lzma too.

frodofine:
I think the CI issues are with how Windows and Mac OS X handle `char`s. I think they handle then as `signed` by default. Unfortunately, I do not have the time to investigate further at the moment.

include/chemfiles/files/GzFile.hpp
Luthaf: > Is this used exclusively by GzFile? Can GzFile only support TextFiles? 

Yes and yes, for now.

This is an implementation of a std::streambuf for reading gz files. The whole interface is built for text-like files, and only decompress parts of the file as needed. This works using the same interface as usual C++ stream objects.

What you mean is about compressed binary formats, like .xtc.gz ? In this case we would need to decompress the whole file before reading it from memory. I don't think there is a generic solution for this. Most binary formats I know already use some kind of internal compression, and in this case there is not much gain to get by compressing the file again.

Also, we would need a binary library that can read from a buffer in memory, which is not the case for all of them.

I think this could be interesting, but it needs to be decided on a format by format basis, as there is no common way to do it. We could still use `gzstreambuf/xzstreambuf` to implement reading and decompressing the whole file in memory (basically calling underflow again and again, copying the resulting values in the buffer).

frodofine: That comment was more idea "vomit" than anything. We would definitely need to read the entire file into memory for most decoders unless they provide their own method of decompressing the files. I agree that it is unlikely that we cannot find a "one size fits all" solution for this issue. Is there a way for the non-text file classes to know that their are getting a compressed file? For example, XTC files will probably be handled by the XDR file class (when ported from molfiles, XDR is common to XTC and TRR). How would XDR now to decompress and feed the binary data to libxdr?

Luthaf: > Is there a way for the non-text file classes to know that their are getting a compressed file?

No, not really. They could check the file name for a different extension though.

Concerning XRD, I don't know how the format works. Is the whole file compressed? This does not sound efficient from a MD point of view: one need to be able to append more data every so often. Or are you talking about compressions realized by the user after the file was created by another software ? In this case, I think the best thing to do is to dispatch on file extension, like `TextFile::create` does today.

frodofine:
Looks great!

PR #169, opened 2018-06-10
Mol2 support

frodofine:
Support for reading and writing MOL2 files without using Molfiles. Addresses part of #154.


src/formats/MOL2.cpp
```diff
@@ -0,0 +1,298 @@
...
+    residues_.clear();
+    frame.resize(0);
+    frame.reserve(static_cast<size_t>(natoms));
+
+    auto junk = file_->readline();
+    bool charges = (trim(file_->readline()) != "NO_CHARGES");
+
+    while (!file_->eof()) {
+        auto curr_pos = file_->tellg();
+        auto line2 = trim(file_->readline());
```
Luthaf: Could this use the same `line` variable ?

src/formats/MOL2.cpp
```diff
@@ -0,0 +1,298 @@
...
+            scan(line, " %lu %31s %lf %lf %lf %31s %lu %31s %lf",
+                &id, &atom_name[0], &x, &y, &z, &atom_type[0], &resid, &res_name[0], &charge
+            );
+        } else {
+            scan(line, " %lu %31s %lf %lf %lf %31s %lu %31s",
+                &id, &atom_name[0], &x, &y, &z, &atom_type[0], &resid, &res_name[0]
+            );
+        }
+
+        Atom atom(atom_name);
```
Luthaf: You can create the atom with the `atom_type` directly instead of setting it later:

```cpp
Atom atom(atom_name, atom_type);
```

src/formats/MOL2.cpp
```diff
@@ -0,0 +1,298 @@
...
+    }
+
+}
+
+void MOL2Format::read_atoms(Frame& frame, size_t natoms, bool charges) {
+    auto lines = file_->readlines(natoms);
+
+    for (const auto& line : lines) {
+        unsigned long id, resid;
+        char atom_name[32], atom_type[32], res_name[32];
```
Luthaf: Could you initialize the three buffers to 0? `char atom_name[32] = {0}`

src/formats/MOL2.cpp
```diff
@@ -0,0 +1,298 @@
...
+                &id, &atom_name[0], &x, &y, &z, &atom_type[0], &resid, &res_name[0], &charge
+            );
+        } else {
+            scan(line, " %lu %31s %lf %lf %lf %31s %lu %31s",
+                &id, &atom_name[0], &x, &y, &z, &atom_type[0], &resid, &res_name[0]
+            );
+        }
+
+        Atom atom(atom_name);
+        frame.add_atom(std::move(atom), Vector3D(x, y, z));
```
Luthaf: Why not creating the atom directly here, instead of creating it and then moving it?

You can also use initializer lists to create the vector with less noise
```cpp
frame.add_atom(Atom(atom_name, atom_type), {x, y, z});
```

src/formats/MOL2.cpp
```diff
@@ -0,0 +1,298 @@
...
+    for (const auto& line : lines) {
+        unsigned long id, id_1, id_2;
+
+        scan(line, " %lu %lu %lu", &id, &id_1, &id_2);
+
+        // MOL2 is 1 index-based, not 0
+        --id_1;
+        --id_2;
+
+        if (id_1 >= frame.size() || id_2 >= frame.size()) {
```
Luthaf: Does the format specify that the bonds section appears after the atoms section? Else this test would fail

src/formats/MOL2.cpp
```diff
@@ -0,0 +1,298 @@
...
+        }
+
+        frame.add_bond(id_1, id_2);
+    }
+}
+
+std::streampos read_until(TextFile& file, const std::string& tag) {
+    while (!file.eof()) {
+        std::streampos pos = file.tellg();
+        if (file.readline().substr(0, tag.length()) == tag) {
```
Luthaf: I guess this is for `@<TRIPOS>xxx` tags. Do they always start on the first column of the line ?

src/formats/MOL2.cpp
```diff
@@ -0,0 +1,298 @@
...
+            }
+
+            if (counts.size() >= 2) {
+                nbonds = string2longlong(counts[1]);
+            }
+
+            read_until(file, "@<TRIPOS>ATOM");
+            file.readlines(static_cast<size_t>(natoms));
+
+            if (nbonds > 0) {
```
Luthaf: what happens if nbonds is 0 or negative?

src/formats/MOL2.cpp
```diff
@@ -0,0 +1,298 @@
...
+        }
+    }
+
+    return std::streampos(-1);
+}
+
+void MOL2Format::write(const Frame& frame) {
+    fmt::print(*file_, "@<TRIPOS>MOLECULE\n");
+
+    if (frame.get("name") && frame.get("name")->get_kind() == Property::STRING) {
```
Luthaf: This will do three lookup for the `name` property. How about storing is in a variable?

```cpp
auto name = frame.get("name");
if (name && name->get_kind() == Property::STRING) {
   // ...
}
```

src/formats/MOL2.cpp
```diff
@@ -0,0 +1,298 @@
...
+        fmt::print(*file_, frame.get("name")->as_string());
+    }
+    fmt::print(*file_, "\n");
+
+    // Only use numbers bigger than the biggest residue id as "resSeq" for
```
Luthaf: does mol2 have a notion of resSeq? I though this was PDB specific

src/formats/MOL2.cpp
```diff
@@ -0,0 +1,298 @@
...
+    // atoms without associated residue.
```
Luthaf: Could you change this to `start after the maximal residue id for atoms without associated residue`?

src/formats/MOL2.cpp
```diff
@@ -0,0 +1,298 @@
...
+    uint64_t max_resid = 0;
+    for (const auto& residue: frame.topology().residues()) {
+        auto resid = residue.id();
+        if (resid && resid.value() > max_resid) {
+            max_resid = resid.value();
+        }
+    }
+
+    auto bonds = std::vector<std::pair<size_t, size_t>>();
```
Luthaf: You don't need to create a new vector here: you should be able to use `frame.topology().bonds()` directly.

src/formats/MOL2.cpp
```diff
@@ -0,0 +1,298 @@
...
+  : file_(TextFile::create(path, mode)) {
+    while (!file_->eof()) {
+        auto position = forward(*file_);
+        if (position == std::streampos(-1)) {
+            break;
+        }
+        if (!file_) {
+            throw format_error("IO error while reading '{}' as MOL2", path);
+        }
+        steps_positions_.push_back(position);
```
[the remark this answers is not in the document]
frodofine: Possibly, but that is beyond the scope of this PR I think.

src/formats/MOL2.cpp
frodofine: Yes

src/formats/MOL2.cpp
frodofine: No, I meant it in the same manner. The order of residues in MOL2 is typically arbitrary.

src/formats/MOL2.cpp
frodofine: Sure

src/formats/MOL2.cpp
frodofine: Ah, gotcha!

include/chemfiles/formats/MOL2.hpp
```diff
@@ -0,0 +1,55 @@
+// Chemfiles, a modern library for chemistry file reading and writing
+// Copyright (C) Guillaume Fraux and contributors -- BSD license
+
+#ifndef CHEMFILES_FORMAT_MOL2_HPP
+#define CHEMFILES_FORMAT_MOL2_HPP
+
+#include <map>
```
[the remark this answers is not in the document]
frodofine: Sure

src/formats/MOL2.cpp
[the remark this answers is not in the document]
Luthaf: yeah, definitively =) Just throwing an idea in here

src/formats/MOL2.cpp
Luthaf: At least a warning would be good, as it might help to understand why there is an error.

frodofine:
Unless there are any complaints, I'll merge this branch in the next few hours or so.

src/formats/MOL2.cpp
```diff
@@ -70,18 +72,21 @@ void MOL2Format::read(Frame& frame) {
...
+    // Skip a line
+    file_->readline();
+
+    // If charges are specified, we need to expect an addition term for each atom
     bool charges = (trim(file_->readline()) != "NO_CHARGES");
 
     while (!file_->eof()) {
-        auto curr_pos = file_->tellg();
-        auto line2 = trim(file_->readline());
+        const auto& curr_pos = file_->tellg();
```
Luthaf: You don't need a reference here, std::streampos is not a big object, and it it returned by value by `tellg`

Luthaf:
> Unless there are any complaints, I'll merge this branch in the next few hours or so.

I'm good !

PR #171, opened 2018-06-13
Rename Frame::guess_topology to Frame::guess_bonds

Luthaf:
It makes more sense, now that the topology contains more than just bonds.


frodofine:
I agree.

PR #175, opened 2018-06-20
Use a string interner for PDB connectivity

Luthaf:
The PDB connectivity table contains a lot of identical strings. Instead of having multiple copies of `"C1"` laying around, this puts the string in a vector, and then use an index to link to it.

This divides the object file by 4 on my system when compiling with `clang -O2`. This might allow to have more residues as default in the code.


frodofine:
Everything looks good to me!

PR #176, opened 2018-06-20
Bond Order Support

### LINKED ISSUE #170: Support Storing Bond Orders
Many formats store the bond order of various bonds in molecules. Examples include MOL2, SMI, MMTF, and SDF.  Currently Chemfiles not store bond orders, which makes writing these formats properly not impossible.

My solution is to store the bond order as an integer in the bond class.  Some complications include the fact some formats include orders such as 'ar' for an aromatic bond and 'am' for amide bonds. We could make an enum for storing this, but I'd like to hear other suggestions as well.

frodofine:
There are some mergers in here, so I'll probably squash the commits. Still, I think this closes the issue for now!

Closes #170.


include/chemfiles/Connectivity.hpp
```diff
@@ -294,7 +312,7 @@ class Connectivity final {
     const sorted_set<Improper>& impropers() const;
 
     /// Add a bond between the atoms `i` and `j`
-    void add_bond(size_t i, size_t j);
+    void add_bond(size_t i, size_t j, BondOrder bo = UNKNOWN);
```
Luthaf: bo => order? Or `bond_order`?

Luthaf: Also, could you add the bond order parameter to the documentation?

include/chemfiles/Connectivity.hpp
```diff
@@ -322,6 +342,8 @@ class Connectivity final {
     mutable sorted_set<Improper> impropers_;
     /// Is the cached content up to date ?
     mutable bool uptodate_ = false;
+    /// Store the bond orders
+    std::vector<BondOrder> bondorders_;
```
Luthaf: It feels like `bond_orders_` would be more readable, but this is highly subjective, so do as you want =)

include/chemfiles/Frame.hpp
```diff
@@ -247,10 +247,11 @@ class CHFL_EXPORT Frame final {
     ///
     /// @param atom_i the index of the first atom in the bond
     /// @param atom_j the index of the second atom in the bond
+    /// @param bo the bond order of the new bond
     /// @throws OutOfBounds if `atom_i` or `atom_j` are greater than `size()`
     /// @throws Error if `atom_i == atom_j`, as this is an invalid bond
-    void add_bond(size_t atom_i, size_t atom_j) {
-        topology_.add_bond(atom_i, atom_j);
+    void add_bond(size_t atom_i, size_t atom_j, Connectivity::BondOrder bo = Connectivity::UNKNOWN) {
```
Luthaf: Same as `Connectivity::add_bond`

include/chemfiles/Topology.hpp
```diff
@@ -110,9 +110,10 @@ class CHFL_EXPORT Topology final {
     ///
     /// @param atom_i the index of the first atom in the bond
     /// @param atom_j the index of the second atom in the bond
+    /// @param bo the bond order for the bond added
     /// @throws OutOfBounds if `atom_i` or `atom_j` are greater than `size()`
     /// @throws Error if `atom_i == atom_j`, as this is an invalid bond
-    void add_bond(size_t atom_i, size_t atom_j);
+    void add_bond(size_t atom_i, size_t atom_j, Connectivity::BondOrder bo = Connectivity::UNKNOWN);
```
Luthaf: Same as `Connectivity::add_bond`

include/chemfiles/Connectivity.hpp
```diff
@@ -279,11 +279,29 @@ inline bool operator>=(const Improper& lhs, const Improper& rhs) {
 /// other data are cached from it.
 class Connectivity final {
 public:
+
+    /// Stores the type of bond
+    enum BondOrder {
```
Luthaf: Why not make this an enum class, outside of Connectivity ? The Connectivity class is more or less an implementation detail for users of chemfiles, and defining this as an enum inside Connectivity make it visible outside of the code.

include/chemfiles/Topology.hpp
```diff
@@ -126,6 +127,18 @@ class CHFL_EXPORT Topology final {
     /// @throws OutOfBounds if `atom_i` or `atom_j` are greater than `size()`
     void remove_bond(size_t atom_i, size_t atom_j);
 
+    /// Get the bond order for the given bond
+    ///
+    /// If the bond does not exist, this will return a out_of_bonds error
```
Luthaf: out_of_bonds => out_of_bounds

Also, do we want to throw an error or to return BondOrder::UNKNOWN?

include/chemfiles/capi/topology.h
```diff
@@ -230,6 +230,62 @@ CHFL_EXPORT chfl_status chfl_topology_residues_linked(
...
+    /// Aromatic bond (required by some file formats)
+    CHFL_BOND_AROMATIC = 255,
+} chfl_bond_order;
+
+/// Add a bond between the atoms at indexes `i` and `j` with bond order `bo` in the `topology`.
+///
+/// @example{tests/capi/doc/chfl_topology/bond_order.c}
+/// @return The operation status code. You can use `chfl_last_error` to learn
+///         about the error if the status code is not `CHFL_SUCCESS`.
+CHFL_EXPORT chfl_status chfl_topology_add_bond_order(
```
Luthaf: this function name sounds like it is only adding the bond order, not the bond itself. How about `chfl_topology_add_bond_with_order`? It is a bit long though ...

Luthaf: This should be mirrored on chfl_frame too, like chfl_frame_add_bond.

src/Connectivity.cpp
```diff
@@ -206,14 +225,30 @@ void Connectivity::atom_removed(size_t index) {
...
 }
+
+Connectivity::BondOrder Connectivity::bond_order(size_t i, size_t j) const {
+    auto pos = bonds_.find(Bond(i, j));
+    if (pos != bonds_.end()) {
+        auto diff = std::distance(bonds_.cbegin(), pos);
+        return bondorders_[static_cast<size_t>(diff)];
+    }
+
+    throw error(
```
Luthaf: Could you put this in an else block to make it clearer?

src/Connectivity.cpp
```diff
@@ -206,14 +225,30 @@ void Connectivity::atom_removed(size_t index) {
...
+        "out of bounds atomic index in `Connectivity::bond_order`: "
+        "No bond between {} and {} exists",
+        i, j
+    );
+}
```
Luthaf: Can you add a new line at the end of the file?

src/capi/topology.cpp
```diff
@@ -230,6 +230,39 @@ extern "C" chfl_status chfl_topology_residues_linked(
...
 }
 
+extern "C" chfl_status chfl_topology_add_bond_order(CHFL_TOPOLOGY* const topology, uint64_t i, uint64_t j, chfl_bond_order bo) {
+    CHECK_POINTER(topology);
+    CHFL_ERROR_CATCH(
+        topology->add_bond(checked_cast(i), checked_cast(j), static_cast<Connectivity::BondOrder>(bo));
+    )
+}
+
+#pragma intel optimization_level 2  /* Using -O3 with icc lead to partial copy of the bonds */
```
Luthaf: Is this needed here too ?

Luthaf:
> There are some mergers in here, so I'll probably squash the commits.

You can also rebase on top of master, this should make the merge commits go away!

include/chemfiles/Connectivity.hpp
frodofine: readable like `uptodate_` :smile: ? I was copying your style, so that's why I clubbed it into one word!

include/chemfiles/Topology.hpp
frodofine: Yeah, that's up for discussion.  I felt an error would be more appropriate and `UNKNOWN` really means unspecified. We could add something to the enum like `UNSPECIFIED` and use `UNKNOWN` for error cases. Thoughts?

include/chemfiles/capi/topology.h
frodofine: Sure thing.

include/chemfiles/Connectivity.hpp
frodofine: Yes, in may codes `bond_order` is abbreviated as `bo`, but I'm willing to change it!

src/capi/topology.cpp
frodofine: I'm not sure actually. I figured the code was similar to the other functions so the `pragma` would be appropriate.

include/chemfiles/Connectivity.hpp
frodofine: Hmm, where would be a good place for that? I thought about that but couldn't find a logical place for it.

include/chemfiles/Connectivity.hpp
Luthaf: One point for you =). We can change `uptodate_` to `up_to_date_` too (or to a better name if you can think of one!)

include/chemfiles/Connectivity.hpp
Luthaf: It could still be defined in the same file, together with `Bond`.

include/chemfiles/Topology.hpp
Luthaf: This sounds good! Is there a difference in some formats between unknown and unspecified ?

src/capi/topology.cpp
Luthaf: Ok. I'll do some tests to see if this is still needed nowadays.

include/chemfiles/Connectivity.hpp
frodofine: Like a member of `Bond` or defined outside the class?

include/chemfiles/Topology.hpp
frodofine: I don't think so.  Some formats (PDB for example) just define everything as single bonds thus all bonds would be `UNSPECIFIED` because the format does not support bond orders.

Mol2 defines an `unknown` bond and a `dummy` bond for cases where the bond order cannot be determined using SYBYL typing rules.

include/chemfiles/Connectivity.hpp
Luthaf: Either way could work, and I don't have strong preference on this. Putting it inside Bond would be better with a plain enum, so that we can use `Bond::DOUBLE`, putting it outside would need an class enum, used as `BondOrder::DOUBLE`.

include/chemfiles/Topology.hpp
Luthaf: Thinking a bit more about this, having both `UNKNOWN` and `UNSPECIFIED` can be confusing for the users, about what the difference is between the two ...

We might want to use an exception for now, and experiment a bit with the API

include/chemfiles/Connectivity.hpp
frodofine: I like it in `Bond` as a bond order is a property of a bond, not an independent concept.

include/chemfiles/Topology.hpp
frodofine: Ok, I've added some logic to MOL2 and SDF to use those format's way of interpreting an unknown bond as unspecified.

frodofine:
If you no one has complaints, I'll merge in an hour or so.

include/chemfiles/Topology.hpp
Luthaf: This still contains the typo: out_of_bonds => out_of_bounds

include/chemfiles/capi/topology.h
```diff
@@ -230,6 +230,43 @@ CHFL_EXPORT chfl_status chfl_topology_residues_linked(
     bool* result
 );
 
+/// Add a bond between the atoms at indexes `i` and `j`
+/// with bond order `bond_order` in the `topology`.
+///
+/// @example{tests/capi/doc/chfl_topology/bond_order.c}
+/// @return The operation status code. You can use `chfl_last_error` to learn
+///         about the error if the status code is not `CHFL_SUCCESS`.
+CHFL_EXPORT chfl_status chfl_topology_add_bond_w_order(
```
Luthaf: I am a bit bothered by the single `w` here, but `chfl_topology_add_bond_with_order` would be over the 31 characters limit.

What do you think of `chfl_topology_bond_with_order`?

include/chemfiles/capi/topology.h
```diff
@@ -230,6 +230,43 @@ CHFL_EXPORT chfl_status chfl_topology_residues_linked(
...
+    CHFL_TOPOLOGY* topology, uint64_t i, uint64_t j, chfl_bond_order bond_order
+);
+
+/// Get the list of bond orders in the `topology` in the
+/// pre-allocated array `orders` of size `nbos`.
```
Luthaf: `nbos` => `nbonds`

include/chemfiles/capi/topology.h
```diff
@@ -230,6 +230,43 @@ CHFL_EXPORT chfl_status chfl_topology_residues_linked(
...
+///
+/// `orders` size must be passed in the `nbos` parameter, and be equal to
```
Luthaf: `nbos` => `nbonds`

include/chemfiles/capi/topology.h
Luthaf: And same for `chfl_frame_bond_with_order` :smiley:

include/chemfiles/Topology.hpp
```diff
@@ -129,7 +129,7 @@ class CHFL_EXPORT Topology final {
 
     /// Get the bond order for the given bond
     ///
-    /// If the bond does not exist, this will return a out_of_bonds error
+    /// If the bond does not exist, this will thrown an Error.
```
Luthaf: Should this be `throw`?

frodofine: I capitalized it because it throws the `Error` class

frodofine:
I'll probably merge this tomorrow, there's another branch that depends on this!

include/chemfiles/Topology.hpp
Luthaf: No, I mean throw vs thrown =)

PR #180, opened 2018-07-09
Use configuration to create new atom types

### LINKED ISSUE #75: Allow to override element data in a configuration file

Luthaf:
The main part of this PR is e1edd74, where I added the ability to define new atom types in the configuration file:

```toml
[atoms.Zn]
charge = 1.8

[atoms]
CH4 = {mass = 16.043, charge = 0, full_name = "methane"}
Ph = {mass = 77.106, full_name = "phenyl"}
```

Additionally, I renamed the configuration file from `.chemfilesrc` to `.chemfiles.toml`, which should be clearer and give syntactic coloration.

Fix #75.


CHANGELOG.md
```diff
@@ -32,6 +32,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 * There is only one constructor for the `Frame` class: `Frame(UnitCell cell =
   UnitCell())`. The constructor taking a topology can be replaced with calls to
   `Frame::add_atom` and `Frame::add_bond`.
+* Chemfiles will now read configuration from `.chemfiles.toml` or
```
frodofine: Good idea to include the `toml` file extension.  This makes it clear how the file should be formatted.

doc/src/configuration.rst
```diff
@@ -4,22 +4,22 @@ Configuration files
...
+the filesystem. For example, with the following filesystem:
 
 .. code-block:: bash
 
     $HOME
-    ├── .chemfilesrc (1)
+    ├── .chemfiles.toml (1)
     └── work
-        ├── .chemfilesrc (2)
+        ├── chemfiles.toml (2)
```
frodofine: Should this be `.chemfiles.toml`?

doc/src/configuration.rst
```diff
@@ -53,3 +53,37 @@ the right atomic data can be accessed.
...
+    CH4 = {mass = 16.043, charge = 0, full_name = "methane"}
+    Ph = {mass = 77.106, full_name = "phenyl"}
+
+The ``[atoms]`` section contains multiple tables, one for each new atom type.
+These tables can contains the following key/values (all keys are optional):
+
+- ``full_name`` (string) : the full name of the atomic type;
+- ``mass`` (number) : the mass of the atomic type in atomic units;
+- ``charge`` (number) : the charge of the atomic type in multiples of *e*;
+- ``mass`` (number) : the Van der Waals radius of the atomic type in Angstrom;
```
frodofine: Is this meant to be `radius`, not `mass`?

doc/src/configuration.rst
```diff
@@ -53,3 +53,37 @@ the right atomic data can be accessed.
...
+- ``mass`` (number) : the covalent radius of the atomic type in Angstrom;
```
frodofine: Same comment as above.

include/chemfiles/Atom.hpp
```diff
@@ -100,12 +108,12 @@ class CHFL_EXPORT Atom final {
...
     /// with case-insensitive atom type: `Na`, `NA`, `nA` and `na` all get the
-    /// `Na` full name.
+    /// `Na` full name. To change the value returned for a given type, you can
+    /// use configuration files.
     ///
-    /// @verbatim embed:rst:leading-slashes
-    /// This function returna an :cpp:class:`chemfiles::optional` value that is
-    /// close to C++17 ``std::optional``.
-    /// @endverbatim
+    /// @verbatim embed:rst:leading-slashes This function returna an
```
frodofine: returns over returna

include/chemfiles/Configuration.hpp
```diff
@@ -16,17 +22,23 @@ namespace chemfiles {
 /// The configuration is stored in TOML files, that are accessed and read on the
 /// first access to the global Configuration instance.
 ///
-/// The configuration contains type renaming information in the `[types]` toml
-/// table. For example,
+/// The configuration contains multiples sections:
```
frodofine: multiple over multiples

include/chemfiles/Configuration.hpp
```diff
@@ -8,6 +8,12 @@
 #include <unordered_map>
 
 #include "chemfiles/mutex.hpp"
+#include "chemfiles/periodic_table.hpp"
+
+namespace toml {
+    class value;
```
frodofine: I'm wondering if it would be better to use the property class here over `value`?

doc/src/configuration.rst
Luthaf: No, both `.chemfiles.toml` and `chemfiles.toml` are supported (in this order if both exist. The idea is that files starting with a dot are not hidden on Windows, and feels weird there. I got the idea from `appveyor.yml` which does the same.

include/chemfiles/Configuration.hpp
Luthaf: I don't understand what you mean here. This is a forward declaration of `toml::value`, so that it can be used in the class declaration. As far as I know, there is no `toml::property` class.

Luthaf:
Updated, thanks!

include/chemfiles/Configuration.hpp
frodofine: Whoops, never mind.  I meant the `chemfiles::Property` class. I release this doesn't make sense, ignore my comment.

PR #181, opened 2018-07-09
Various cleanup tasks

Luthaf:
(this is based on top of #180, I will rebase when the other one is merged)

### DIFF
--- a/cmake/CompilerFlags.cmake
+++ b/cmake/CompilerFlags.cmake
@@ -45,6 +45,12 @@ if(MSVC)
     set(CMAKE_SHARED_LINKER_FLAGS "/SUBSYSTEM:CONSOLE")
 endif()
 
+if(${CMAKE_CXX_COMPILER_ID} STREQUAL "Intel")
+    # Some version of intel compiler (icc 14 at least) have only partial support
+    # for C++11 
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DMSGPACK_USE_CPP03")
+endif()
+
 if(EMSCRIPTEN)
     if(BUILD_SHARED_LIBS)
         # Shared libs where not tested and a lot of changes to the build system
@@ -97,6 +103,10 @@ set(CHEMFILES_C_WARNINGS "")
 macro(remove_msvc_warning _warn_)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /wd${_warn_}")
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /wd${_warn_}")
+endmacro()
+
+macro(remove_intel_warning _warn_)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -diag-disable ${_warn_}")
 endmacro()
 
 if(MSVC)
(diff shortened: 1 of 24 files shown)


cmake/CompilerFlags.cmake
```diff
@@ -45,6 +45,10 @@ if(MSVC)
     set(CMAKE_SHARED_LINKER_FLAGS "/SUBSYSTEM:CONSOLE")
 endif()
 
+if(${CMAKE_CXX_COMPILER_ID} STREQUAL "Intel")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DMSGPACK_USE_CPP03")
```
frodofine: I don't recall this being an issue when I tested with ICC.  I was using version 17 through which may have better support for more modern standards.

src/capi/topology.cpp
```diff
@@ -107,7 +107,6 @@ extern "C" chfl_status chfl_topology_impropers_count(const CHFL_TOPOLOGY* const
     )
 }
 
-#pragma intel optimization_level 2  /* Using -O3 with icc lead to partial copy of the bonds */
```
frodofine: So they are not needed after all!

src/capi/topology.cpp
```diff
@@ -260,7 +255,10 @@ extern "C" chfl_status chfl_topology_bond_order( const CHFL_TOPOLOGY* const topo
     CHECK_POINTER(order);
 
     CHFL_ERROR_CATCH(
-        *order = static_cast<chfl_bond_order>(topology->bond_order(i, j));
+        *order = static_cast<chfl_bond_order>(topology->bond_order(
+            static_cast<size_t>(i),
```
frodofine: Is this for 32-bit support? Should the function take `size_t` as parameter options over `uint64_t`.

src/files/GzFile.cpp
```diff
@@ -77,7 +77,7 @@ bool gzstreambuf::is_open() const {
 
 std::streampos gzstreambuf::seekoff(std::streamoff offset, std::ios_base::seekdir way, std::ios_base::openmode mode) {
     if (sync() == EOF) {
-        return std::streampos(EOF);
+        return EOF;
```
frodofine: Should it be `{EOF}`?

cmake/CompilerFlags.cmake
Luthaf: Yeah, I have ICC 14, which does not support enough of C++11. Do you think I should add a check on intel version ?

src/capi/topology.cpp
Luthaf: Nope! This was pretty old though, I don't remember why I had to add them

src/capi/topology.cpp
Luthaf: Yes, this is for 32-bit support. And I think passing uint64_t to the function is better, as this remove a source of variation on the ABI, and make for easier foreign languages bindings.

cmake/CompilerFlags.cmake
frodofine: Might be overkill because as then we'd need to check exactly which version we'd need.  Best to make a note of it for now.

src/files/GzFile.cpp
Luthaf: Yes, makes sense

Luthaf:
> I'm not going to argue much with clang-tidy, it probably knows best.

Well, the windows failure I had on this was due to a clang-tidy suggestion (changing `auto ident = next().ident()` to `const auto& ident = next().ident()`), where MSVC failed to create a temporary and extend it's lifetime ^^

frodofine:
Guess computers won't take over the world soon then.

src/capi/topology.cpp
frodofine: Fair enough just looked odd to me.

PR #182, opened 2018-07-10
Request compression in format argument

Luthaf:
This PR allow to request using GzFile/XzFile from the format parameter of trajectory, as `<format>/ GZ` or `<format>/XZ`, where `<format>` can be empty.

Ref #71

### DIFF
--- a/include/chemfiles/File.hpp
+++ b/include/chemfiles/File.hpp
@@ -28,20 +28,21 @@ public:
 
     virtual ~File() noexcept = default;
     File(File&&) = default;
-    File& operator=(File&&) = delete;
+    File& operator=(File&&) = default;
     File(File const&) = delete;
     File& operator=(File const&) = delete;
 
-    /// File name, i.e. complete path to this file on disk.
-    const std::string& filename() const { return filename_; }
-    /// File opening mode.
+    /// Get the file path used to open this file.
+    const std::string& path() const { return path_; }
+
+    /// Get the mode used to open this file.
     Mode mode() const { return mode_; }
 
 protected:
-    File(std::string path, Mode mode): filename_(std::move(path)), mode_(mode) {}
+    File(std::string path, Mode mode): path_(std::move(path)), mode_(mode) {}
 
 private:
-    const std::string filename_;
+    std::string path_;
     Mode mode_;
 };
 
@@ -54,7 +55,7 @@ private:
 class CHFL_EXPORT TextFile: public File, public std::iostream {
 public:
     /// Open the most adaptated text file class for the given `path` and `mode`
-    static std::unique_ptr<TextFile> create(const std::string& path, File::Mode mode);
+    static std::unique_ptr<TextFile> open(std::string path, File::Mode mode);
 
     /// Read a line from the file
     std::string readline();
@@ -68,7 +69,7 @@ public:
 protected:
     /// Initialize the TextFile at the given `path` and `mode`. All read and
     /// write operations will go through the provided `buffer`.
-    TextFile(const std::string& path, File::Mode mode, std::streambuf* buffer);
+    TextFile(std::string path, File::Mode mode, std::streambuf* buffer);
 
 private:
     void get_line(std::string& string);
--- a/include/chemfiles/File.hpp
+++ b/include/chemfiles/File.hpp
@@ -26,6 +26,16 @@ public:
         APPEND = 'a',
     };
 
+    /// Possible compression methods for opening a file
+    enum Compression {
+        /// Default method: plain text or binary formats
+        DEFAULT,
+        /// gzip compression
+        GZIP,
+        /// lzma compression (.xz)
+        LZMA,
+    };
+
     virtual ~File() noexcept = default;
     File(File&&) = default;
     File& operator=(File&&) = default;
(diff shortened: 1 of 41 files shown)


include/chemfiles/File.hpp
```diff
@@ -53,8 +69,9 @@ class CHFL_EXPORT File {
 /// the user of the class to the current state.
 class CHFL_EXPORT TextFile: public File, public std::iostream {
 public:
-    /// Open the most adaptated text file class for the given `path` and `mode`
-    static std::unique_ptr<TextFile> create(const std::string& path, File::Mode mode);
+    /// Open the file at the given `path` with the requested `mode` and
+    /// `compression` method using the most adapted child class of `TextFile`.
+    static std::unique_ptr<TextFile> open(std::string path, File::Mode mode, File::Compression compression);
```
frodofine: Would a default argument for compression (`DEFAULT`) be useful here?

include/chemfiles/FormatFactory.hpp
```diff
@@ -16,7 +16,7 @@
 
 namespace chemfiles {
 
-using format_creator_t = std::function<std::unique_ptr<Format>(const std::string& path, File::Mode mode)>;
+using format_creator_t = std::function<std::unique_ptr<Format>(std::string path, File::Mode mode, File::Compression compression)>;
```
frodofine: Is there a reason to copy the `path` string?

src/File.cpp
```diff
@@ -7,30 +7,23 @@
...
-static std::string extension(const std::string& filename) {
-    auto idx = filename.rfind('.');
-    if (idx != std::string::npos) {
-        return filename.substr(idx);
-    } else {
-        return "";
+std::unique_ptr<TextFile> TextFile::open(std::string path, File::Mode mode, File::Compression compression) {
+    switch (compression) {
+    case File::DEFAULT:
+        return std::unique_ptr<TextFile>(new PlainFile(std::move(path), mode));
```
frodofine: Is it to use move semantics? Does this offer performance improvements?

src/formats/MMTF.cpp
```diff
@@ -19,26 +19,15 @@ template<> FormatInfo chemfiles::format_information<MMTFFormat>() {
     );
 }
 
-static std::string extension(const std::string& filename) {
```
frodofine: Great! This was mostly a hack anyway.

tests/trajectory.cpp
```diff
@@ -167,13 +167,37 @@ TEST_CASE("Specify a format parameter") {
...
+    file.write(frame);
+    file.close();
+
+    // Full format specification
+    file = Trajectory(tmpfile, 'r', "XYZ / GZ");
+    frame = file.read();
+    CHECK(frame.size() == 1);
+    CHECK(frame[0].name() == "Fe");
+
+    // only the compression method, the format will be guessed from extension
```
frodofine: Should there be a test to see if both format and extension can be guessed?

src/Trajectory.cpp
```diff
@@ -5,30 +5,60 @@
...
+    std::string extension;
+    File::Compression compression;
+};
+
+file_open_info file_open_info::parse(const std::string& path, const std::string& format) {
+    auto info = file_open_info {"", "", File::DEFAULT};
+
+    auto slash = format.find('/');
+    if (slash != std::string::npos) {
+        const auto& compression = trim(format.substr(slash + 1));
```
frodofine: Have `XYZ/GZ` or `XYZ/ GZ` been tested?

src/File.cpp
Luthaf: More like semantic improvement: this is passed all the way down to `File`, which takes ownership of the string.

tests/trajectory.cpp
Luthaf: What do you mean ? `Trajectory("foo.xzy.gz", 'r', "")`?

src/Trajectory.cpp
Luthaf: I don't think so, I'll add them.

include/chemfiles/File.hpp
Luthaf: I don't think this is needed, as only Trajectory will call this function (through format_creator_t lambda, through FormatFactory), and it will always do so with all parameters.

src/File.cpp
frodofine: Gotcha

include/chemfiles/File.hpp
frodofine: I see, it was the first thing I saw without reading the rest of the PR.

tests/trajectory.cpp
frodofine: Yes

Luthaf: This is tested in https://github.com/chemfiles/chemfiles/blob/377b72eb9aca4893c2f9e72a67ec7be03e84da63/tests/files/xz-file.cpp#L10-L15

Luthaf: This is tested in https://github.com/chemfiles/chemfiles/blob/377b72eb9aca4893c2f9e72a67ec7be03e84da63/tests/formats/compressed-xyz.cpp#L43-L53

frodofine: Must have missed that

Luthaf: This was added in a separated PR (I think #166)

Luthaf:
Are we good ? :smiley:

frodofine:
Yes!

Luthaf:
Thank you for the review !

PR #185, opened 2018-07-24
Improved PDB Support

frodofine:
The first commit is self-explanatory.

The second is a bit more controversial. I added the ability to change the topology based on a given property.  I am wondering if this should be done in a more general sense (IE not just the `altloc` property).


Luthaf:
Sorry, I did not get a notification for your changes here.

PR #186, opened 2018-07-25
More code cleanup

Luthaf:
This PR does a bit of cleanup, and small fixes.

It also adds a new function: `Trajectory::path`, the naming of which is open to debate !


CHANGELOG.md
```diff
@@ -34,6 +34,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   `Frame::add_atom` and `Frame::add_bond`.
 * Chemfiles will now read configuration from `.chemfiles.toml` or
   `chemfiles.toml` instead of `.chemfilesrc`
+* Added `Trajectory::path` to get the file path used to create a trajectory
```
frodofine: I like Path!

CMakeLists.txt
```diff
@@ -62,6 +62,10 @@ if(MSVC)
     add_definitions("-D_CRT_SECURE_NO_WARNINGS -D_SCL_SECURE_NO_WARNINGS")
 endif()
 
+if(NOT CHFL_SYSTEM_LZMA)
+    add_definitions("-DLZMA_STATIC_API")
```
frodofine: Thanks, this fixes the build on Windows. I actually made a branch for this but never submitted the PR :roll_eyes: . I'll go delete that.

appveyor.yml
```diff
@@ -10,14 +10,21 @@ environment:
   matrix:
     - generator: MinGW Makefiles
       CXX_PATH: 'C:\mingw-w64\x86_64-6.3.0-posix-seh-rt_v5-rev1\mingw64\bin'
+      BUILD_SHARED_LIBS: ON
       ARCH: x64
     - generator: MinGW Makefiles
       CXX_PATH: 'C:\mingw-w64\i686-6.3.0-posix-dwarf-rt_v5-rev1\mingw32\bin'
+      BUILD_SHARED_LIBS: ON
```
frodofine: Should there be a test for static builds on MinGW?

external/CMakeLists.txt
```diff
@@ -60,7 +60,6 @@ set(EXTERNAL_INCLUDES "")
 # ==========
 # fmtlib: https://github.com/chemfiles/fmt
 # ==========
-set(FMT_USE_CPP14 OFF CACHE bool "Enable the addition of C++14 compiler flags.")
```
frodofine: Has FMT changed?

src/selections/expr.cpp
```diff
@@ -430,8 +430,8 @@ optional<double> Number::optimize() {
 }
 
 std::string Number::print() const {
-    if (lround(value_) == value_) {
-        return std::to_string(lround(value_));
+    if (static_cast<double>(std::lround(value_)) == value_) {
```
frodofine: Would `std::round` be better than `std::lround` with a cast?

frodofine:
Looks like MinGW is having problems, I think one needs to put the GzFile deconstructor in a `.cpp file`, but that is only based on the error message.

external/CMakeLists.txt
Luthaf: Yes, they do automatic C++ standard detection now

appveyor.yml
Luthaf: I don't know. The MSVC one seems to be catching the issues with LZMA, so it might be enough. And Appveyor is already very slow ...

src/selections/expr.cpp
Luthaf: Yes, good catch

CMakeLists.txt
Luthaf: I don't think is fixes the build, I still get errors on the static builder ... I'm trying to build it locally in a VM, but it is super slow =(

frodofine: I have a windows box I was working on.

Luthaf:
MinGW should be fixed now, I removed debuginfo and the build time went from 38min to 19min \o/

Now to fix static build !

Luthaf:
I'll merge this tomorrow if you are OK!

frodofine:
Looks good!

Issue #178, opened 2018-06-28
Feature request: Residue groups


frodofine:
I believe there is a need for residue groups in the `Frame` class to store data such as chain IDs and other related information in Biologic systems (I'm sure other applications could benefit here, but I do not know them).  Such an implementation could look like:
```
class ResidueGroup {
    sorted_set<size_t> residue_ids_
    property_map prop_map_
public:
    set / get - property
    check if a residue is in the group
    add / residues from the group
}
```
Thoughts?

Luthaf:
I am a bit worried about proliferation of ad-hoc structures here, with both Residue and ResidueGroup being loosely integrated with the Frame (both only store index of atoms/residues in a separated frame).

What is made available by `ResidueGroup` that could not be done with a `chain_id` string property on atoms/residues? If one need the ability to get all atoms with a given `chain_id`, maybe extending selections to work with properties would be another solution.

If there is a real need for ResidueGroup as a separated class, maybe we can re-organize the code to make the whole ResiudeGroup -> Residue -> Atom clearer. 

---

On this matter, see https://khinsen.wordpress.com/2013/04/10/lessons-from-sixteen-years-of-molecular-simulation-in-python/, at the end he says he would prefer having only Atom and Group instead of the whole Atoms, Groups, Molecules, Complexes, AminoAcidResidue, PeptideChain, and Protein. He does not go on to say why, but I would guess less types of groups make the code more general and easier to understand.

frodofine:
I really like the selection based idea, although I would have no idea how to implement it...

frodofine:
Yes, a selection mechanism would satisfy my needs (in my own project I build the residue group myself anyway, I was really just suggesting to move that upstream).

I could add the selector part my self as it looks easy enough, but I need to finish up some other projects before I can look at this more in detail.

frodofine:
This issue is closed, with the above PRs.

PR #189, opened 2018-08-29
Use atomic properties in selections

Luthaf:
This PR contains a bit of cleaning for the selection code (4 first commits) and then implement the use of properties in selections, as discussed in #178.

The syntax used is `[<property>]`, e.g. `[is_hetatm]`, `[chainid]`, etc. Properties with spaces are supported with quotes: `["why not?"]`

For multiple atoms, the variable part comes after the closing bracket `[chainid](#3)`.

Because there is no way to know whether we are parsing a property to be evaluated as a boolean, a string or a number, the parser try to find the one that matches by the context: if the property is followed by 'and' or 'or', then use it as a boolean. If it is followed by a string; `==` or `!=` and then a string, use it as a string. Else use it as a number. Then, during the selection evaluation, and error is thrown if a property with this name exist but has the wrong type.

Finally, missing properties are evaluated to `false` for boolean properties (which prevent making a difference between missing properties and properties set to false); to an empty string for string properties; and to NaN for numeric properties. `Vector3D` properties are not supported.

## Unresolved questions:

- [ ] What should the syntax be for variables? `[short](#3)` and `["long prop"](#3)` or `[short(#3)]` and `["long prop"(#3)]`
- [ ] Should missing numeric properties evaluate to 0 instead of NaN?

## TODO

- [x] Documentation
- [x] Changelog

### DIFF
--- a/include/chemfiles/selections/lexer.hpp
+++ b/include/chemfiles/selections/lexer.hpp
@@ -7,6 +7,7 @@
 #include <string>
 #include <vector>
 #include <cstdint>
+#include <cassert>
 
 #include <chemfiles/exports.hpp>
 #include <chemfiles/Error.hpp>
@@ -15,6 +16,9 @@ namespace chemfiles {
 namespace selections {
 
 using Variable = uint8_t;
+
+/// Check that a given string is a valid identifier
+bool is_ident(const std::string& string);
 
 /// A token in the selection stream
 class Token {
@@ -74,7 +78,10 @@ public:
     Token& operator=(Token&&) = default;
 
     /// Create an identifier token with `data` name
-    static Token ident(std::string data);
+    static Token ident(std::string data) {
+        assert(is_ident(data));
+        return Token(IDENT, std::move(data), 0.0, 0);
+    }
 
     /// Create a string with some `data` inside
     static Token string(std::string data) {
--- a/include/chemfiles/selections/lexer.hpp
+++ b/include/chemfiles/selections/lexer.hpp
@@ -17,13 +17,6 @@ namespace selections {
 using Variable = uint8_t;
 
 /// A token in the selection stream
-///
-/// Various tokens are alowed here:
-///
-/// - binary comparison operators (== != < <= > >=);
-/// - boolean operators (and not or);
-/// - numbers, the scientific notation is alowed;
-/// - identifiers, obeing to the ([a-Z][a-Z_1-9]+) regular expression
 class Token {
 public:
     /// Available token types
@@ -62,10 +55,10 @@ public:
         OR,
         /// "not" token
         NOT,
-        /// Generic identifier
+        /// Identifier, described by the '[a-Z][a-Z_1-9]+' regex
         IDENT,
-        /// Generic identifier, inside double quotes
-        RAW_IDENT,
+        /// Arbitrary string inside double quotes, might contain spaces
+        STRING,
         /// Number
         NUMBER,
         /// "#(\d+)" token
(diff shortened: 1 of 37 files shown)


include/chemfiles/selections/lexer.hpp
```diff
@@ -123,12 +125,21 @@ class Token {
     /// Get the identifier name associated with this token.
     /// The token type must be `IDENT`.
     const std::string& ident() const {
-        if (type_ != IDENT && type_ != RAW_IDENT) {
+        if (type_ != IDENT) {
             throw Error("can not get an ident value out of this token. This is a bug.");
```
frodofine: ident -> identifier

Less cryptic error message?

tests/doc/property/double.cpp
```diff
@@ -11,15 +11,15 @@ TEST_CASE() {
...
     assert(property.as_double() == 45.2);
 
     // Various overload convert integers to double
-    assert(Property(45).get_kind() == Property::DOUBLE);     // int
-    assert(Property(45l).get_kind() == Property::DOUBLE);    // long
-    assert(Property(45ll).get_kind() == Property::DOUBLE);   // long long
-    assert(Property(45u).get_kind() == Property::DOUBLE);    // unsigned
-    assert(Property(45ul).get_kind() == Property::DOUBLE);   // unsigned long
-    assert(Property(45ull).get_kind() == Property::DOUBLE);  // unsigned long long
+    assert(Property(45).kind() == Property::DOUBLE);     // int
```
frodofine: Should integer properties be added?

Luthaf: Most integers can be represented as a double value, so I don't think it is worth it.

frodofine: Fair enough.

include/chemfiles/selections/lexer.hpp
```diff
@@ -82,12 +83,13 @@ class Token {
 
     /// Create an identifier token with `data` name
     static Token ident(std::string data) {
+        assert(is_ident(data));
         return Token(IDENT, std::move(data), 0.0, 0);
     }
 
-    /// Create a raw identifier token with `data` name
-    static Token raw_ident(std::string data) {
```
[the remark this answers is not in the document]
Luthaf: what do you mean ?

Luthaf:
This is ready for a final round of review!

include/chemfiles/selections/lexer.hpp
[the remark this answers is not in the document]
frodofine: I was wondering what 'ident' meant.

include/chemfiles/selections/lexer.hpp
```diff
@@ -62,10 +63,10 @@ class Token {
         OR,
         /// "not" token
         NOT,
-        /// Generic identifier
+        /// Identifier, described by the '[a-Z][a-Z_1-9]+' regex
```
frodofine: Great, this explains this perfectly!

include/chemfiles/selections/lexer.hpp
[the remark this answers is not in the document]
Luthaf: Oh, sorry =) That's an identifier. Do you think this should be defined somewhere ?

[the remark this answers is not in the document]
frodofine: Nope, I figured it out when you removed 'raw_indent'.

frodofine:
I checked this out, certainly solves #178.

PR #191, opened 2018-08-30
Iterate over properties

Luthaf:
This replaces #163

### DIFF
--- a/include/chemfiles/Property.hpp
+++ b/include/chemfiles/Property.hpp
@@ -215,6 +215,11 @@ public:
     /// Get the property with the given `name` if it exists.
     optional<const Property&> get(const std::string& name) const;
 
+    /// Get the number of properties in this property map
+    size_t size() const {
+        return data_.size();
+    }
+
     /// Get an iterator to the first property in the property map
     iterator begin() const {
         return data_.begin();
--- a/include/chemfiles/Property.hpp
+++ b/include/chemfiles/Property.hpp
@@ -201,8 +201,10 @@ inline bool operator!=(const Property& lhs, const Property& rhs) {
 }
 
 /// A property map for inclusion in a Frame or an Atom.
-class property_map final {
+class CHFL_EXPORT property_map final {
 public:
+    using iterator = std::unordered_map<std::string, Property>::const_iterator;
+
     property_map() = default;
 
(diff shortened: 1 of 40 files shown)


include/chemfiles/Property.hpp
```diff
@@ -213,6 +215,20 @@ class property_map final {
     /// Get the property with the given `name` if it exists.
     optional<const Property&> get(const std::string& name) const;
 
+    size_t size() const {
```
frodofine: Add documentation here?

src/capi/topology.cpp
```diff
@@ -75,87 +75,87 @@ extern "C" chfl_status chfl_topology_remove(CHFL_TOPOLOGY* const topology, uint6
     )
 }
 
-extern "C" chfl_status chfl_topology_bonds_count(const CHFL_TOPOLOGY* const topology, uint64_t* nbonds) {
+extern "C" chfl_status chfl_topology_bonds_count(const CHFL_TOPOLOGY* const topology, uint64_t* const count) {
```
frodofine: I like the new consistency in variable names!

tests/capi/doc/chfl_atom/list_properties.c
```diff
@@ -0,0 +1,31 @@
...
+    chfl_property_free(property);
+
+    uint64_t count = 0;
+    chfl_atom_properties_count(atom, &count);
+    assert(count == 2);
+
+    const char* names[2] = {NULL};
+    chfl_atom_list_properties(atom, names, count);
+
+    // Properties are not ordered
```
frodofine: I'm guessing this is implementation dependent due to an unordered map.

Luthaf: Yes. I think it it better to document it as not ordered for now, this can be changed later without breaking code

include/chemfiles/Property.hpp
Luthaf: Done!

tests/capi/doc/chfl_atom/list_properties.c
frodofine: This is fine, I just wanted to make sure.

frodofine:
I checked this out. It definitely solves #178 !

PR #193, opened 2018-08-31
MMTF write support

frodofine:
The title is self-explanatory.

Addresses #149


include/chemfiles/formats/MMTF.hpp
```diff
@@ -21,20 +21,24 @@ class MMTFFormat final: public Format {
...
     void read(Frame& frame) override;
+    void write(const Frame& frame) override;
     size_t nsteps() override;
 
 private:
 
     mmtf::StructureData structure_;
 
+    std::string filename_;
+    size_t writeAtomLoc_ = 0;
```
Luthaf: What does is this used for ? It seems to be used as a flag to know whether to write or not in the destructor, but the variable name seems to indicate something else.

frodofine:
> So, correct me if I am wrong, but this code is only writing to the file in the destructor, and is adding new models to the structure in write, right?

Yes, `mmtf::compressGroupList(structure_);` can only be called once after all the models have been added.

frodofine:
Unless there are any major concerns, I'm going to merge this. I have some more features, but I think I'll include them in #185.

include/chemfiles/formats/MMTF.hpp
Luthaf: Could you add a bit of doc for this and the other members ?

frodofine: Sure thing.

Luthaf:
Sounds good to me ! Feel free to merge when CI passes =)

Issue #10, opened 2015-06-15
CIF files


Luthaf:
File Type                        | text
-------------------------------|---------
Topological information | Yes
Positions                       | Yes
Velocities                      | No

Reference of the foirmat: http://scripts.iucr.org/cgi-bin/paper?S010876739101067X

A library for guessing symmetry is available here: https://github.com/atztogo/spglib

Luthaf:
It appears that CIF files are a written using the STAR (Self-defining Text Archive and Retrieval) file format, using a subset of the STAR rules. The mmCIF is a subset of the CIF files. So implementing CIF reader will give mmCIF for free, and I think that writting CIF files using mmCIF format can not harm.

Relevant links:

* [STAR format paper](http://dx.doi.org/10.1107/97809553602060000727)
* [STAR file parser in Java](http://www.globalphasing.com/startools/)
* [C++ mmCIF reader/writer](http://sw-tools.pdb.org/apps/CIFPARSE-OBJ/index.html)
* [CIF syntax reference](http://www.iucr.org/resources/cif/spec/version1.1/cifsyntax)

frodofine:
I've looked int the C++ mmCIF parser in the link above, but I feel the license is too restrictive and would conflict with Chemfile's current license.  Therefore, I've rolled my own version which is compatible with the files provided by RCSB, the COD, the CCSD, and PyMOL. I'll push this in a day or so.

Luthaf:
Whow, that's great ! I started writting a CIF parser ([here](https://github.com/chemfiles/)), but never finished it ... 

Do you implement the whole CIF grammar ? I would like to use it for pure crystallographic CIF files too !

frodofine:
Nope, I took the dirty PyMOL approach and have it recognize a very small set of the CIF grammar (just atom site at the moment). My application is the COD and CCSD, so I'm only testing against those databases (no idea if that's 'pure' or not!).

fxcoudert:
I am interested in support for the CIF format as well.
Would a dependency on [GEMMI](https://gemmi.readthedocs.io/en/latest/) be acceptable here? 

- It boasts having the “ the fastest open-source CIF parser”
- Its CIF parser is [header-only](https://gemmi.readthedocs.io/en/latest/cif.html) so it's simple to integrate as a new dependency
- I have worked with it and can confirm it is good code, easy to use

Luthaf:
[...]

PR #194, opened 2018-09-05
Initial mmCIF support

frodofine:
Read support for the CIF format. This parser does not understand CIF grammar, but looks for keywords in the file and uses these keywords to parse the file.  I've tested it on files obtained from [RCSB](https://www.rcsb.org/), [COD](http://www.crystallography.net/cod/
), and [CCDC ](https://www.ccdc.cam.ac.uk/)(actual test not included due to licensing).

Write support coming *soon*.


src/formats/CIF.cpp
```diff
@@ -0,0 +1,364 @@
...
+    }
+
+    file_->seekg(position);
+
+    for (const auto& residue: residues_) {
+        frame.add_residue(residue.second);
+    }
+
+    // Cross format talk! Forgive me!
+    PDBFormat::link_standard_residue_bonds(frame);
```
Luthaf: I like the comment ^-^

On a side note, this is not valid for usual CIF files, only biomolecules ones from the PDB archive and consorts. I though there where using the .mmcif extension for there own extensions to the format

src/formats/PDB.cpp
```diff
@@ -208,6 +208,8 @@ void PDBFormat::read_ATOM(Frame& frame, const std::string& line,
 
             // This will be save as a string... on purpose to match MMTF
             residue.set("chainid", line.substr(21,1));
+            // This fromat makes not distinction between chainid and chainname
```
Luthaf: fromat => format

include/chemfiles/formats/CIF.hpp
```diff
@@ -0,0 +1,53 @@
...
+    size_t nsteps() override;
+private:
+    std::unique_ptr<TextFile> file_;
+    /// Map of STAR records to their index
+    std::map<std::string, size_t> atom_site_map_;
+    /// Map of residues, indexed by residue id and chainid.
+    std::map< std::pair<std::string, size_t>, Residue> residues_;
+    /// Set to true if the file is based on fractional coordinates.
+    /// Set to false if the file is based on cartn coordinates.
+    bool uses_fract_;
```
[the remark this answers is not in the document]
frodofine: Not the ones from RCSB....

include/chemfiles/formats/CIF.hpp
```diff
@@ -0,0 +1,53 @@
...
+    /// Storing the positions of all the steps in the file, so that we can
+    /// just `seekg` them instead of reading the whole step.
+    std::vector<std::streampos> steps_positions_;
+    /// The cell for all frames
+    UnitCell cell_;
+    /// Number of models written/read to the file.
+    size_t models_;
+    /// Did we wrote a frame to the file? This is used to check wheter we need
+    /// to write a final `END` record in the destructor
```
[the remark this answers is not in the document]
frodofine: Yes and no. I plan to use this eventually!

src/formats/CIF.cpp
frodofine: They use .cif, quite confusing.

src/formats/CIF.cpp
```diff
@@ -0,0 +1,364 @@
...
+#include "chemfiles/pdb_connectivity.hpp"
+
+// WARNING UGLY HACK!
+#include "chemfiles/formats/PDB.hpp"
+
+using namespace chemfiles;
+
+template<> FormatInfo chemfiles::format_information<CIFFormat>() {
+    return FormatInfo("CIF").with_extension(".cif").description(
+        "PDB (RCSB Protein Data Bank) text format"
```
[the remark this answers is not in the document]
frodofine: Sorta, I didn't change the description because I was basing it off of RCSB's standard.

include/chemfiles/formats/CIF.hpp
```diff
@@ -0,0 +1,53 @@
+// Chemfiles, a modern library for chemistry file reading and writing
+// Copyright (C) Guillaume Fraux and contributors -- BSD license
+
+#ifndef CHEMFILES_FORMAT_CIF_HPP
+#define CHEMFILES_FORMAT_CIF_HPP
+
+#include <map>
```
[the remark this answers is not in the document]
frodofine: It was for the residue map, which I want to iterate over in a specified order.

src/formats/CIF.cpp
Luthaf: Is there some kind of RCSB specific key we could check for before applying this ?

frodofine: I already use the `_atom_site.pdbx_pdb_model_num` key to identify a CIF from the PDB. This key is present in 94% of all CIF files from the PDB, so I could use this again.

On another note, I am wondering if this should be relabeled as mmCIF support.  The CIF write code I am working writes cartesian coordinates (I mirror PyMOL and RCSB here).

Luthaf: Yeah, both sounds good !

frodofine: I checked OpenBabel as well and it is done in a similar manner. They implement mmCIF and CIF as different formats (the read code is similar, but registered differently). The write code is different for the two formats, one writes cartesian and the other fractional.  This is probably the best way to proceed.  It does not matter to me as the 'mmCIF' parser here works for the 'CIF' files I am working with anyway.

frodofine:
I think this PR is ready for final review. It addresses part of #10, but not all of it.

README.md
```diff
@@ -106,6 +106,7 @@ Chemfiles can read and write files in the following formats:
 - SERC Daresbury Laboratory CSSR files;
 - GROMACS formats: .gro;
 - TRIPOS mol2 files;
+-
```
Luthaf: Is this missing a `mmCIF` format ?

frodofine: I could've sworn I did that. Good catch...

.vscode/settings.json
```diff
@@ -0,0 +1,87 @@
...
+        "cwctype": "cpp",
+        "resumable": "cpp",
+        "future": "cpp",
+        "shared_mutex": "cpp",
+        "stack": "cpp",
+        "xthread": "cpp",
+        "typeindex": "cpp",
+        "*.ipp": "cpp"
+    }
+}
```
Luthaf: You should add this file to your .gitignore =)

PR #195, opened 2018-09-12
Revert "Initial mmCIF support"

frodofine:
Reverts chemfiles/chemfiles#194.

It appears that this causes the documentation to fail to build with the following error:

```Exception occurred:
  File "/home/travis/.local/lib/python2.7/site-packages/sphinx/highlighting.py", line 26, in <module>
    from sphinx.ext import doctest
SyntaxError: unqualified exec is not allowed in function 'run' it contains a nested function with free variables (doctest.py, line 97)
The full traceback has been saved in /tmp/sphinx-err-ZmhAW2.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
```

I'm not sure what the error is, but reverting this PR for now should resolve it?


frodofine:
Nevermind, it is not this PR's fault!

Apparently Sphinx 1.8.0 just dropped and this breaks Chemfiles! I'll make another PR to show this.

PR #196, opened 2018-09-12
Set Sphinx version to avoid bug in 1.8.0

frodofine:
Sphinx v1.8.0 launched today around noon. Apparently, there is a bug in this version that causes the documentation build for Chemfiles to fail. I am forcing the Sphinx version to be set at 1.7.9.

Note: I have not submitted a bug upstream, but we should consider doing so.

### DIFF
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -1,3 +1,3 @@
-sphinx >1.6.1
+sphinx==1.7.9
 breathe >4.7
 sphinx_bootstrap_theme


Luthaf:
Is the issue related to [breathe](https://github.com/michaeljones/breathe) ? Anyway, reporting the bug upstream would be good !

frodofine:
I'll go ahead and merge this. No one likes the build failed message from Travis!

I do not know about breathe, but the error comes seems to come from Python itself. Could it be a Python 2/3 issue? On my machine 1.8.0 works fine, it's just Travis that's the issue...

PR #198, opened 2018-09-26
 Only set CMake build type if chemfiles is the main project 

Luthaf:


frodofine:
I'm at a conference today, so I'll take a deeper look when I get back tomorrow.

CMakeLists.txt
```diff
@@ -30,13 +28,15 @@ option(CHFL_BUILD_DOCTESTS "Build documentation tests as well as unit tests." ON
...
 # Set a default build type if none was specified
-if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
-    message(STATUS "Setting build type to 'release' as none was specified.")
-    set(CMAKE_BUILD_TYPE "release"
-        CACHE STRING
-        "Choose the type of build, options are: none(CMAKE_CXX_FLAGS or CMAKE_C_FLAGS used) debug release relwithdebinfo minsizerel."
-    FORCE)
-    set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS release debug relwithdebuginfo minsizerel none)
+if (${CMAKE_CURRENT_SOURCE_DIR} STREQUAL ${CMAKE_SOURCE_DIR})
+    if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
```
frodofine: I believe newer versions of CMake prefer ${CMAKE_BUILD_TYPE} over CMAKE_BUILD_TYPE. Maybe it's a good idea to consider migrating over for new commits?

scripts/ci/setup-travis.sh
```diff
@@ -9,7 +9,7 @@ else
 fi
 
 if [[ "${DO_COVERAGE}" == "ON" ]]; then
-    export CMAKE_ARGS="$CMAKE_ARGS -DCHFL_CODE_COVERAGE=ON"
+    export CMAKE_ARGS="$CMAKE_ARGS -DCMAKE_C_FLAGS=\"--coverage\" -DCMAKE_CXX_FLAGS=\"--coverage\""
```
frodofine: Do we still need to check for GCC?

CMakeLists.txt
```diff
@@ -66,17 +66,6 @@ if(NOT CHFL_SYSTEM_LZMA)
     add_definitions("-DLZMA_API_STATIC")
 endif()
 
-if(CHFL_CODE_COVERAGE)
-    message(STATUS "Code coverage enabled")
-    if(NOT CMAKE_COMPILER_IS_GNUCXX)
```
frodofine: Is this check still required?

scripts/ci/setup-travis.sh
Luthaf: No, because `$DO_COVERAGE` is only set when using GCC

CMakeLists.txt
Luthaf: The idea is that the users can set `C_FLAGS/CXX_FLAGS` if they want code coverage. If so, the flag will be checked by the compiler directly

CMakeLists.txt
Luthaf: Yeah, seems like a good idea. As long as we can build with cmake 2.8.12 (the platform one on CentOS 7 / RedHat) I'm fine with using more modern CMake features.

scripts/ci/setup-travis.sh
frodofine: That's CI only I assume.

I guess if the user tries to run coverage on a non-GCC platform their fate is their own now!

Luthaf: Yep, that's how I see it =)

PR #197, opened 2018-09-16
Start documenting all properties used by chemfiles

Luthaf:
The idea is to have a big table resuming all properties used by chemfiles when reading/writing formats.

Because writing tables by hand with ReST is a pain, I used the `csv-table` directive. Because writing CSV by hand is a pain, I wrote TOML files, and convert them to CSV in the configuration script.

@frodofine, I'll need your help to document the properties used by the MMTF and mmCIF formats!

PR #201, opened 2018-10-14
Added python-toml

frodofine:
Looks like something broke Travis!


Luthaf:
This should not be needed on master. It is used in #197, but the package should be added in `doc/requirements.txt` there. Why is it needed ?

Luthaf:
the master build looks stuck to me ("no output in 10min"), so I restarted it

frodofine:
I added it to master because the build on master broke. I've moved it to doc/requirements.txt

Luthaf:
Oh, OK, https://travis-ci.org/chemfiles/chemfiles/jobs/441303292 (which is for #197) is broken due to missing a toml dependency =/

frodofine:
Yeah, don't know why it said master branch to be honest.

I'm ok with closing this and just adding it to the PR.

PR #203, opened 2018-10-29
[PR for v0.8] Manually check for std::ios_base::failure ABI

Luthaf:
Fix #168 
See https://github.com/conda-forge/chemfiles-lib-feedstock/pull/18

The issue was that the ABI of `std::ios_base::failure` in libstdc++ changed around gcc 5, and we can not know at compile time wich ABI will be used in the end-user machine, so we are manually checking the two possible ABI.

PR #192, opened 2018-08-31
Initial InChI support

frodofine:
Still a WIP at the moment. Needs Stereo0D.

PR #204, opened 2018-10-29
Do not use execptions when reading line by line

Luthaf:
This is a better fix for #203, completely removing the use of exceptions here.

### DIFF
--- a/src/File.cpp
+++ b/src/File.cpp
@@ -67,31 +67,6 @@ void TextFile::get_line(std::string& string) {
     }
 }
 
-std::string TextFile::readline() {
-    std::string line;
-    try {
-        get_line(line);
-    } catch (const std::ios_base::failure& e) {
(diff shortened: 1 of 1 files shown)


Luthaf:
> I'm a bit confused, is this a replacement for the other PR?

No, this PR is against master, and the other one against the release-0.8 branch. I prefer not changing the code too much in the 0.8 release, and using the less hacky way here, hence the two PR =)

frodofine:
Understood!

Somewhat related question:

Do you have a release timeline for v0.9.0? I need a 'frozen' version for a DARPA-related project combining chemistry, computer science, and physics!

Luthaf:
> Do you have a release timeline for v0.9.0? 

As soon as possible! I think #197/#150 is the only remaining thing to do before starting the release process =)

frodofine:
Sounds great. I think #187 and #192 can weight for after v0.9.

PR #206, opened 2018-11-08
 Add an API to get property only if it has a given type 

Luthaf:
This allows for cleaner code in formats, when getting properties of a specific type.

### DIFF
--- a/include/chemfiles/Atom.hpp
+++ b/include/chemfiles/Atom.hpp
@@ -182,7 +182,9 @@ public:
     /// the new value.
     ///
     /// @example{tests/doc/atom/property.cpp}
-    void set(std::string name, Property value);
+    void set(std::string name, Property value) {
+        properties_.set(std::move(name), std::move(value));
+    }
 
     /// Get the `Property` with the given `name` for this atom if it exists.
     ///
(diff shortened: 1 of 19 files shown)


include/chemfiles/Atom.hpp
```diff
@@ -182,20 +182,44 @@ class CHFL_EXPORT Atom final {
     /// the new value.
     ///
     /// @example{tests/doc/atom/property.cpp}
-    void set(std::string name, Property value);
```
frodofine: I see this is a trend!

include/chemfiles/Property.hpp
```diff
@@ -200,6 +200,47 @@ inline bool operator!=(const Property& lhs, const Property& rhs) {
     return !(lhs == rhs);
 }
 
+/// Various metadata associated with a given kind, accessible at compile time
+template<Property::Kind kind>
+struct property_metadata {
+    /// Type of the property
+    using type = void;
```
frodofine: Mind explaining this line to me? Is it meant to cause an error if a bad template argument is given?

src/Property.cpp
```diff
@@ -78,3 +79,27 @@ optional<const Property&> property_map::get(const std::string& name) const {
         return nullopt;
     }
 }
+
+template<Property::Kind kind>
+optional<typename property_metadata<kind>::type> property_map::get(const std::string& name) const {
+    auto property = this->get(name);
+    if (property) {
+        if (property->kind() == kind) {
```
frodofine: See previous documentation comment.

include/chemfiles/Frame.hpp
```diff
@@ -355,20 +355,44 @@ class CHFL_EXPORT Frame final {
     /// replaced with the new value.
     ///
     /// @example{tests/doc/frame/set.cpp}
-    void set(std::string name, Property value);
+    void set(std::string name, Property value) {
+        properties_.set(std::move(name), std::move(value));
```
frodofine: Instead of doing this silently, could we emit a warning or throw an error? That or update the documentation to mention that this could lead to an error being thrown when the property is retrieved. -- Fairly minor though.

frodofine: Or we could use a multimap and store multiple properties with the same name and retrieve only if the value and property match, but I don't like this idea as much as explicitly warning the user that replacing a value may lead to an error being thrown.

include/chemfiles/Property.hpp
Luthaf: Yes, exactly =)

include/chemfiles/Frame.hpp
Luthaf: I am not sure what you mean by:

> that this could lead to an error being thrown when the property is retrieved.

Where would the error comes from ?

frodofine: Sorry, I read a `warning` as throw for some reason.

Issue #210, opened 2018-11-12
Improve handling of ATOM/HETATM in PDB and mmCIF


Luthaf:
Launched here: https://github.com/chemfiles/chemfiles/pull/197#discussion_r217929224

Right now, PDB uses HETATM as the default as I could not find a way to differentiate between ATOM/HETATM records. mmCIF uses ATOM as the default, so we might want to unify them and use the same default (I would argue for HETATM, but I could live with ATOM too).

In the longer run, we might want to follow the actual rules for this and try to find atoms in the PDB connectivity table. The only downsides I could see for this are (1) a bit of cost looking up each atom and (2) we don't include the full connectivity tables in chemfiles yet.

frodofine:
Actually, the distinction between `HETATM` and `ATOM` in `PDB` files is a bit more straightforward.  Everything is `HETATM` unless it is a specific residue linkage type, which are predefined.  Therefore, we only need to check the residue type and not worry about connectivity tables.

frodofine:
I wondering if it would be better to remain the property something like `is_standard` and reverse its meaning.

frodofine:
`is_standard_pdb` is probably best as it is assigned by the wwPDB, not RCSB.  Also, it should be a `Residue` property as it assigned based on connectivity, not on atoms. This means we only need to make the check per residue, not per atom.  We already have all the standard residues in the connectivity table, so one can just check there if need be.

Luthaf:
Sounds good to me !

Issue #207, opened 2018-11-11
Improper handling of TER records in PDB files


frodofine:
`TER` records in a `PDB` **may** increase the atom id count.  This can cause issues when connect records are read as the atom ids will not align. One example is [4HHB](https://www.rcsb.org/structure/4HHB) as it contains `TER` records and `CONECT` records. 

One potential fix is to use `std::queue` and push the atom ids every time the atom ids jump. When `CONECT` records are read, the ids are checked against the queue and the offset is adjusted when the current id is greater than the current offset.

PR #211, opened 2018-11-12
Additional PDB fixes

frodofine:
The purpose of this PR is a few fold:
 - [X] Fix issue #207 
 - [x] Fix issue #210
 - [x] Add secondary structure information stored in PDB/MMTF files.
 - [x] Read frame properties from PDB files, like `HEADER`, `TITLE`, etc.
 - [x] Add corresponding properties from mmCIF and MMTF where applicable.

### DIFF
--- a/include/chemfiles/formats/PDB.hpp
+++ b/include/chemfiles/formats/PDB.hpp
@@ -5,6 +5,7 @@
 #define CHEMFILES_FORMAT_PDB_HPP
 
 #include <map>
+#include <vector>
 
 #include "chemfiles/Format.hpp"
 #include "chemfiles/File.hpp"
@@ -53,6 +54,9 @@ private:
     std::vector<std::streampos> steps_positions_;
     /// Number of models written/read to the file.
     size_t models_;
+    /// List of all atom offsets. This maybe pushed in read_ATOM or if a TER
+    /// record is found. It is reset every time a frame is read.
+    std::vector<size_t> atom_offsets_;
     /// Did we wrote a frame to the file? This is used to check wheter we need
     /// to write a final `END` record in the destructor
     bool written_ = false;
--- a/include/chemfiles/formats/PDB.hpp
+++ b/include/chemfiles/formats/PDB.hpp
@@ -6,6 +6,7 @@
 
 #include <map>
 #include <vector>
+#include <tuple>
 
 #include "chemfiles/Format.hpp"
 #include "chemfiles/File.hpp"
@@ -42,13 +43,14 @@ private:
     void read_CRYST1(Frame& frame, const std::string& line);
     // Read ATOM and HETATM records
     void read_ATOM(Frame& frame, const std::string& line, bool is_hetatm);
-
+    // Read HELIX records
+    void read_HELIX(const std::string& line);
     // Read CONECT record
     void read_CONECT(Frame& frame, const std::string& line);
 
     std::unique_ptr<TextFile> file_;
-    /// Map of residues, indexed by residue id.
-    std::map<size_t, Residue> residues_;
+    /// Map of residues, indexed by residue id and chainid.
+    std::map<std::pair<char, size_t>, Residue> residues_;
     /// Storing the positions of all the steps in the file, so that we can
     /// just `seekg` them instead of reading the whole step.
     std::vector<std::streampos> steps_positions_;
@@ -60,6 +62,8 @@ private:
     /// Did we wrote a frame to the file? This is used to check wheter we need
     /// to write a final `END` record in the destructor
     bool written_ = false;
+    /// Store secondary structure information
+    std::vector<std::tuple<char, size_t, size_t, std::string>> secinfo_;
 };
 
 template<> FormatInfo format_information<PDBFormat>();
--- a/include/chemfiles/formats/PDB.hpp
+++ b/include/chemfiles/formats/PDB.hpp
@@ -43,8 +43,11 @@ private:
     void read_CRYST1(Frame& frame, const std::string& line);
     // Read ATOM and HETATM records
     void read_ATOM(Frame& frame, const std::string& line, bool is_hetatm);
-    // Read HELIX records
+    // Read secondary structure records. All push secinfo_ vector if line is valid
     void read_HELIX(const std::string& line);
+    // reads SHEET and TURN records. i1 and i2 are the indicies of the chain ids
+    void read_secondary(const std::string& line, size_t i1, size_t i2,
+                        const std::string& sec);
     // Read CONECT record
     void read_CONECT(Frame& frame, const std::string& line);
 
@@ -62,7 +65,8 @@ private:
     /// Did we wrote a frame to the file? This is used to check wheter we need
     /// to write a final `END` record in the destructor
     bool written_ = false;
-    /// Store secondary structure information
+    /// Store secondary structure information. First field is the chainid,
+    /// followed by the first and last residue id in the secondary structure.
     std::vector<std::tuple<char, size_t, size_t, std::string>> secinfo_;
 };
 
--- a/src/formats/MMTF.cpp
+++ b/src/formats/MMTF.cpp
@@ -190,6 +190,8 @@ void MMTFFormat::read(Frame& frame) {
             }
 
             if (groupIndex_ < structure_.secStructList.size()) {
+                // We use the raw values here to directly match the MMTF spec
+                // See https://github.com/rcsb/mmtf/blob/master/spec.md#secstructlist
                 switch(structure_.secStructList[groupIndex_]) {
                     case 0:
                         residue.set("secondary_structure", "pi helix");
--- a/src/formats/MMTF.cpp
+++ b/src/formats/MMTF.cpp
@@ -189,6 +189,37 @@ void MMTFFormat::read(Frame& frame) {
                 );
             }
 
+            if (groupIndex_ < structure_.secStructList.size()) {
+                switch(structure_.secStructList[groupIndex_]) {
+                    case 0:
+                        residue.set("secondary_structure", "pi helix");
+                        break;
+                    case 1:
+                        residue.set("secondary_structure", "bend");
+                        break;
+                    case 2:
+                        residue.set("secondary_structure", "alpha helix");
+                        break;
+                    case 3:
+                        residue.set("secondary_structure", "extended");
+                        break;
+                    case 4:
+                        residue.set("secondary_structure", "3-10 helix");
+                        break;
+                    case 5:
+                        residue.set("secondary_structure", "bridge");
+                        break;
+                    case 6:
+                        residue.set("secondary_structure", "turn");
+                        break;
+                    case 7:
+                        residue.set("secondary_structure", "coil");
+                        break;
+                    default:
+                        break;
+                }
+            }
+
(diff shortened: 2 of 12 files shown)


src/formats/PDB.cpp
```diff
@@ -181,6 +191,18 @@ void PDBFormat::read_ATOM(Frame& frame, const std::string& line,
...
 
+    if (atom_offsets_.size() == 0) {
+        try {
+            auto initial_offset = std::stoul(line.substr(6,5));
+            if (initial_offset == 0) {
+                throw format_error("0 is not a valid atom id.");
+            }
+            atom_offsets_.push_back(initial_offset - 1);
+        } catch(std::invalid_argument&) {
+            throw format_error("{} is not a valid atom id.", line.substr(6,5));
```
Luthaf: I would rather send a warning here, and try to continue parsing assuming 1 as atom id.

Luthaf: This would allow us to continue parsing files with more than 9999 atoms, where there is no clear way of defining atom ids.

src/formats/mmCIF.cpp
```diff
@@ -91,6 +93,16 @@ mmCIFFormat::mmCIFFormat(std::string path, File::Mode mode, File::Compression co
             gamma = cif_to_double(line_split[1]);
         }
 
+        if (line_split[0] == "_entry.id") {
+            pdb_idcode_ = line_split[1];
+        }
+
+        if (line_split[0] == "_struct.title") {
+            auto temp_name = trim(line.substr(13));
+            std::copy(temp_name.begin() + 1, temp_name.end() - 1,
```
Luthaf: Why the std::copy here, instead of `name_ = temp_name.substr(1, temp_name.size() - 2)` ?

src/formats/mmCIF.cpp
```diff
@@ -190,6 +202,14 @@ void mmCIFFormat::read(Frame& frame) {
     frame.resize(0);
     residues_.clear();
 
+    if (name_.size()) {
+        frame.set("name", name_);
+    }
+
+    if (pdb_idcode_.size()) {
```
Luthaf: Could you use `!pdb_icode.empty()` instead? It is a bit clearer. Or even `pdb_icode != ""`

src/formats/mmCIF.cpp
```diff
@@ -190,6 +202,14 @@ void mmCIFFormat::read(Frame& frame) {
     frame.resize(0);
     residues_.clear();
 
+    if (name_.size()) {
```
Luthaf: Same as below for pdb_icode

src/formats/PDB.cpp
frodofine: I think using an integer for the `initial_offset` would be best, that way if someone is crazy enough to have a negative index, nothing goes crazy.

src/formats/mmCIF.cpp
frodofine: That's fine, but I don't like to do subtraction on 'size()' because there's the chance that it is zero and the result explodes. I'll add a check as well.

src/formats/mmCIF.cpp
frodofine: That makes too much sense 😸

frodofine:
I think this is ready for final review!

include/chemfiles/formats/PDB.hpp
```diff
@@ -42,13 +43,14 @@ class PDBFormat final: public Format {
...
+    // Read HELIX records
+    void read_HELIX(const std::string& line);
     // Read CONECT record
     void read_CONECT(Frame& frame, const std::string& line);
 
     std::unique_ptr<TextFile> file_;
-    /// Map of residues, indexed by residue id.
-    std::map<size_t, Residue> residues_;
+    /// Map of residues, indexed by residue id and chainid.
+    std::map<std::pair<char, size_t>, Residue> residues_;
```
Luthaf: Why isn't the residue id enough to identify the residue?

include/chemfiles/formats/PDB.hpp
```diff
@@ -60,6 +62,8 @@ class PDBFormat final: public Format {
     /// Did we wrote a frame to the file? This is used to check wheter we need
     /// to write a final `END` record in the destructor
     bool written_ = false;
+    /// Store secondary structure information
+    std::vector<std::tuple<char, size_t, size_t, std::string>> secinfo_;
```
Luthaf: Could you add a comment explaining what the char, and the two size_t refers to?

src/formats/MMTF.cpp
```diff
@@ -189,6 +189,37 @@ void MMTFFormat::read(Frame& frame) {
                 );
             }
 
+            if (groupIndex_ < structure_.secStructList.size()) {
+                switch(structure_.secStructList[groupIndex_]) {
+                    case 0:
```
Luthaf: Is there some enum that could be used instead of raw numbers?

src/formats/PDB.cpp
```diff
@@ -138,6 +142,17 @@ void PDBFormat::read(Frame& frame) {
             // Else we have read a frame
             got_end = true;
             continue;
+        case Record::HELIX:
+            read_HELIX(line);
+            continue;
+        case Record::SHEET:
+            secinfo_.emplace_back(std::make_tuple(line[21], std::stoul(line.substr(22, 4)),
```
Luthaf: I would rather have a function to do the reading and handle (with an error) the case of lines that are not long enough, or that contain strings where a number is expected.

src/formats/PDB.cpp
```diff
@@ -199,6 +224,35 @@ void PDBFormat::read_CRYST1(Frame& frame, const std::string& line) {
     }
 }
 
+void PDBFormat::read_HELIX(const std::string& line) {
+    auto chain = line[19];
```
Luthaf: Could you add a check on the line length?

src/formats/PDB.cpp
```diff
@@ -199,6 +224,35 @@ void PDBFormat::read_CRYST1(Frame& frame, const std::string& line) {
...
+    auto start = std::stoul(line.substr(21,4));
```
Luthaf: Could you wrap this is a try block to provide a better error message in case of failure?

src/formats/PDB.cpp
```diff
@@ -199,6 +224,35 @@ void PDBFormat::read_CRYST1(Frame& frame, const std::string& line) {
...
+    auto end = std::stoul(line.substr(33,4));
+
+    switch (line[39] - '0') {
```
Luthaf: This could get a small comment on what is going on =)

include/chemfiles/formats/PDB.hpp
Luthaf: And they would have the same residue id?

I'll double check what happens with this code when using non-standard PDB (that could have a residue id but no chain id)

src/formats/MMTF.cpp
Luthaf: Ok.

src/formats/PDB.cpp
```diff
@@ -225,12 +223,32 @@ void PDBFormat::read_CRYST1(Frame& frame, const std::string& line) {
...
+
+    auto chain1 = line[19];
+    auto chain2 = line[31];
+    size_t start, end;
+
+    try {
+        start = std::stoul(line.substr(21,4));
+        end = std::stoul(line.substr(33,4));
+    } catch (std::invalid_argument&) {
+        warning("Secondary structure record too short: '{}'", line);
```
Luthaf: Copy and paste error in the warning message?

src/formats/PDB.cpp
```diff
@@ -81,13 +94,25 @@ void PDBFormat::read_step(const size_t step, Frame& frame) {
...
+        case Record::HEADER:
+            if (line.size() < 66) {continue;}
+            frame.set("classification", trim(line.substr(10, 40)));
+            frame.set("deposition_date", trim(line.substr(50, 9)));
+            frame.set("pdb_idcode", trim(line.substr(62, 4)));
+            continue;
+        case Record::TITLE:
+            frame.set("name", trim(
+                      frame.get<Property::STRING>("name").value_or("") +
+                      line.substr(10, 70)));
```
Luthaf: Should we add a check on the line length here too?

src/formats/PDB.cpp
```diff
@@ -117,6 +142,22 @@ void PDBFormat::read(Frame& frame) {
...
+            read_secondary(line, 21, 32, "extended");
+            continue;
+        case Record::TURN:
+            read_secondary(line, 19, 30, "turn");
+            continue;
+        case Record::TER:
+            if (line.size() >= 12) {
+                try {
+                    atom_offsets_.push_back(std::stoul(line.substr(6,5)));
+                } catch(std::invalid_argument&) {} // Do nothing
```
Luthaf: Should this send a warning?

src/formats/mmCIF.cpp
```diff
@@ -1,6 +1,8 @@
 // Chemfiles, a modern library for chemistry file reading and writing
 // Copyright (C) Guillaume Fraux and contributors -- BSD license
 
+#include <algorithm>
```
Luthaf: The header is no longer needed right?

include/chemfiles/formats/PDB.hpp
```diff
@@ -41,21 +43,31 @@ class PDBFormat final: public Format {
...
     size_t models_;
+    /// List of all atom offsets. This maybe pushed in read_ATOM or if a TER
+    /// record is found. It is reset every time a frame is read.
+    std::vector<size_t> atom_offsets_;
     /// Did we wrote a frame to the file? This is used to check wheter we need
     /// to write a final `END` record in the destructor
     bool written_ = false;
+    /// Store secondary structure information. First field is the chainid,
+    /// followed by the first and last residue id in the secondary structure.
+    std::vector<std::tuple<char, size_t, size_t, std::string>> secinfo_;
```
Luthaf: This could use a `struct secinfo_meta {char chainid; size_t first; size_t last}` instead of the tuple for more readability. But I'm fine with the std::tuple too.

Luthaf:
Is this the last changes you need for 0.9? If so I'll create a new branch and a first release candidate when the PR is merged.

include/chemfiles/formats/PDB.hpp
frodofine: Yes, the residue IDs all start at 1 for chains A, B, C, D, E, and F. However, the residue sequence for each of them are different!

frodofine:
Yeah, my other changes are not needed.

src/formats/PDB.cpp
frodofine: Originally I thought not, but actually, it makes sense as we should inform the user the line was ignored.

src/formats/PDB.cpp
frodofine: Sure

include/chemfiles/formats/PDB.hpp
frodofine: I don't like making structures that are effectively single use. Since this is single use (only used in `PDB` and only to read secondary structure) I think a Tuple is better.

Luthaf: I find a struct with names more readable that `std::get<0>`, but let's agree to disagree here =)

PR #213, opened 2018-11-18
Update documentation before the release

Luthaf:
A few simple fixes to the documentation


doc/src/installation.rst
```diff
@@ -50,15 +50,8 @@ In order to build the core library, you will need a C++11 capable compiler.
 Chemfiles is automatically tested agaist GCC (>= 4.8) on Linux, OSX and Windows
 (mingw-w64); clang (>= 3.3) on Linux and OS X and MSCV 14 on Windows. It was
 also compiled sucessfully with Intel C++ compilers. Please report any sucessfull
```
frodofine: successful not sucessfull :smile: 

successfully not sucessfully

doc/src/installation.rst
```diff
@@ -69,29 +62,26 @@ All these dependencies can be installed in one command:
...
 
 On Windows
 """"""""""
 
-You can use either MSVC 2015 compiler, or `mingw-w64`_ provided gcc. `MSYS2`_
-offer a package manager to install all the needed libraries. I recomend using it
-if you have no preference over your compiler. After the initial installation
-steps, you can run the following to install a recent C++ compiler:
+You can use either MSVC 2015 compiler, or gcc as provided by `mingw-w64`_.
+`MSYS2`_ offer a package manager to install all the needed libraries. I recomend
```
frodofine: offers

This is a matter of opinion, but given MSVC 2017 is free for community use I think it would be better to recommend it over the 3rd party `mingw-w64`.

Luthaf: Yeah, this is old advice from old times =) I'll update it!

frodofine: I didn't know if it was because gcc is opensource or free!

Luthaf: A bit of both, plus chemfiles was testes more with GCC than MSVC

PR #215, opened 2018-11-21
Add DLL export for extern templates

Luthaf:


frodofine:
It's an American holiday, so I'll review when I get a change.

Luthaf:
This is now part of #217, let's discuss there!

PR #217, opened 2018-11-27
Various cleanup and fixes

Luthaf:
So I compiled the master/0.9.0-rc1 with all compilers I got, and fixed a bunch of warnings on the way. Here are the results!

Warnings/Formatting commits:
- 7da2015
- fa65540 
- bb44644 

Bug fixes:
- a964578 
- 9c7d56e
- 4671006 

Compilation fixes:
- 7fe323b & 04ee926 
- 664c1ea 
- f76fa22

### DIFF
--- a/cmake/CompilerFlags.cmake
+++ b/cmake/CompilerFlags.cmake
@@ -163,19 +163,26 @@ else()
     add_warning_flag("-Wmissing-prototypes")
     add_warning_flag("-Wmissing-variable-declarations")
 
-    # Disable some strong warning with clang that are OK here
+    # Disable some strong warning that are OK here
     add_warning_flag("-Wno-unknown-pragmas")
+    add_warning_flag("-Wno-overlength-strings")
+
+    # Disable some warning implied by -Weverything
+    # -Weverything is not activated by default, but adding these flags help
+    # when adding it manually to check for new warnings 
+    add_warning_flag("-Wno-c++98-compat")
+    add_warning_flag("-Wno-c++98-compat-pedantic")
     add_warning_flag("-Wno-weak-vtables")
     add_warning_flag("-Wno-weak-template-vtables")
-    add_warning_flag("-Wno-switch-enum")
-    # We are not doing C here
     add_warning_flag("-Wno-padded")
-    # Sometime this OK
     add_warning_flag("-Wno-float-equal")
-    add_warning_flag("-Wno-double-promotion")
-    # Yes, chemfiles uses globals
+    add_warning_flag("-Wno-switch-enum")
+    add_warning_flag("-Wno-documentation-unknown-command")
     add_warning_flag("-Wno-exit-time-destructors")
     add_warning_flag("-Wno-global-constructors")
+    # This warning trigger when compiling for 64-bit,
+    # but the code is relevant for 32-bit
+    add_warning_flag("-Wno-tautological-type-limit-compare")
     # Not everyone is as smart as clang for code reachability
     add_warning_flag("-Wno-covered-switch-default")
     add_warning_flag("-Wno-unreachable-code-break")
--- a/cmake/CompilerFlags.cmake
+++ b/cmake/CompilerFlags.cmake
@@ -54,18 +54,12 @@ if(${EMSCRIPTEN})
 
     set(EMCC_FLAGS "")
     set(EMCC_FLAGS "${EMCC_FLAGS} -s DISABLE_EXCEPTION_CATCHING=0")
-    set(EMCC_FLAGS "${EMCC_FLAGS} -s LINKABLE=1")
     set(EMCC_FLAGS "${EMCC_FLAGS} -s ERROR_ON_UNDEFINED_SYMBOLS=1")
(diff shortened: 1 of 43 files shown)


src/UnitCell.cpp
```diff
@@ -44,7 +44,11 @@ UnitCell::UnitCell(double a, double b, double c, double alpha, double beta, doub
     beta_(beta),
     gamma_(gamma)
 {
-    if (alpha_ == 90 && beta_ == 90 && gamma_ == 90) {
+    auto is_90 = [](double angle) {
+        // We think that 89.999° is close enough to 90°
```
frodofine: Very true!

src/formats/AmberNetCDF.cpp
```diff
@@ -109,7 +109,14 @@ UnitCell AmberNetCDFFormat::read_cell() {
     assert(length.size() == 3);
     assert(angles.size() == 3);
 
-    return {length[0], length[1], length[2], angles[0], angles[1], angles[2]};
+    return {
+        static_cast<double>(length[0]),
```
frodofine: What is ``length``'s type here?  The auto above is misleading in this case.

src/formats/AmberNetCDF.cpp
```diff
@@ -122,9 +129,9 @@ void AmberNetCDFFormat::read_array(span<Vector3D> array, const std::string& name
     auto data = array_var.get(start, count);
 
     for (size_t i = 0; i < natoms; i++) {
-        array[i][0] = data[3 * i + 0];
-        array[i][1] = data[3 * i + 1];
-        array[i][2] = data[3 * i + 2];
+        array[i][0] = static_cast<double>(data[3 * i + 0]);
```
frodofine: Same goes for ``data`` here.

src/formats/MMTF.cpp
```diff
@@ -86,10 +86,14 @@ void MMTFFormat::read(Frame& frame) {
     frame.resize(0);
 
     if (structure_.unitCell.size() == 6) {
-        frame.set_cell(UnitCell(
-            structure_.unitCell[0], structure_.unitCell[1], structure_.unitCell[2],
-            structure_.unitCell[3], structure_.unitCell[4], structure_.unitCell[5]
-        ));
+        frame.set_cell({
+            static_cast<double>(structure_.unitCell[0]),
```
frodofine: ``structure_.unitCell`` is already a vector of doubles. Why are the ``static_cast``s needed?

cmake/CompilerFlags.cmake
```diff
@@ -163,19 +153,26 @@ else()
     add_warning_flag("-Wmissing-prototypes")
     add_warning_flag("-Wmissing-variable-declarations")
 
-    # Disable some strong warning with clang that are OK here
+    # Disable some strong warning that are OK here
     add_warning_flag("-Wno-unknown-pragmas")
+    add_warning_flag("-Wno-overlength-strings")
```
frodofine: I'm not familiar with this warning. I guess this is a strong warning here?

Luthaf:
>  I must ask though, what is chemfiles.tar.gz?

I generated a tarball to send it and compile it on supercomputers, but forgot to delete it after. I'll clean it up!

cmake/CompilerFlags.cmake
Luthaf: No, this is a warning about strings longer than 4096 characters, which is the minimum a C99 conforming implementation must support. It comes from `-Wpedantic`.

src/formats/MMTF.cpp
Luthaf: I am pretty sure I got a warning without them. I'll check again.

src/formats/AmberNetCDF.cpp
Luthaf: The type is `NcFloat`, a netcdf array of floats.

src/formats/MMTF.cpp
Luthaf: UnitCell is defined as a vector of float, see https://github.com/chemfiles/mmtf-cpp/blob/63e58a9057ee684b5a0a409932e7fc2126894fc0/include/mmtf/structure_data.hpp#L154

src/UnitCell.cpp
Luthaf: I added a change below, to actually set them to 90 in case they were not. I think this is less surprising than having an ORTHOROMBIC cell with non-diagonal cell matrix.

Luthaf:
> I know some projects (Avogadro for example) have a CI script that checks if any new code doesn't match the coding guidelines ala Clang format. Is this of interest?

I don't really know. It looks like a barrier to entry (contributors must have clang-format installed and run it) to me, and doing a few manual changes from time to time does not really bother me. Do you think it would be worthwhile?

frodofine:
>  Do you think it would be worthwhile?

Probably not.

Luthaf:
Are we good here?

---

We can think again about automated formatting checking when/if the project grows big enough that it starts to become a pain =)

frodofine:
Looks good!

src/formats/MMTF.cpp
frodofine: Odd, Ok.

PR #220, opened 2018-12-22
Improve residue id handling

Luthaf:
This PR came from an issue where converting LAMMPS Data to PDB would create a warning about residue id 18446744073709551615 (size_t(-1)).

So the commits fix the LAMMPS Data reader wrt molecule id, and force usage of the same function for all parsing of string to numeric types, checking that the parsed value fit in the given type.

### DIFF
--- a/include/chemfiles/utils.hpp
+++ b/include/chemfiles/utils.hpp
@@ -11,7 +11,7 @@
 #include <cstdarg>
 #include <algorithm>
 
-#include "chemfiles/Error.hpp"
+#include "chemfiles/ErrorFmt.hpp"
 
 namespace chemfiles {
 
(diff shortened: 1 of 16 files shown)


src/selections/lexer.cpp
```diff
@@ -195,14 +195,14 @@ Token Tokenizer::variable() {
...
+    size_t data = 0;
     auto number = input_.substr(start, count);
     try {
-        data = std::stoi(number);
+        data = parse<size_t>(number);
     } catch (const std::exception&) {
         throw selection_error("could not parse number in '{}'", number);
     }
-    if (data > UINT8_MAX) {
+    if (data > static_cast<size_t>(UINT8_MAX)) {
```
frodofine: Would `std::numeric_limits` be better here?

src/formats/PDB.cpp
```diff
@@ -337,47 +339,47 @@ void PDBFormat::read_ATOM(Frame& frame, const std::string& line,
...
+    auto altloc = line.substr(16, 1);
     if (altloc != " ") {
         atom.set("altloc", altloc);
     }
 
     try {
-        auto x = std::stod(line.substr(31, 8));
-        auto y = std::stod(line.substr(38, 8));
-        auto z = std::stod(line.substr(46, 8));
+        auto x = parse<double>(trim(line.substr(31, 8)));
```
frodofine: Trimming is now required due to the length check, correct?

include/chemfiles/utils.hpp
```diff
@@ -52,37 +52,48 @@ std::string user_name();
 /// Get the process current directory
 std::string current_directory();
 
-/// Convert a string to double, throwing a `chemfiles::Error` if the string is
-/// not a valid double.
-inline double string2double(const std::string& string) {
+/// Convert a string to `T`, throwing a `chemfiles::Error` if the string is not
+/// a valid `T`.
+template<typename T> inline T parse(const std::string& string);
```
frodofine: Using templates is a much better idea then individual functions!

src/selections/lexer.cpp
Luthaf: According to https://en.cppreference.com/w/cpp/types/numeric_limits, there is no specialization of `std::numeric_limits ` for `uint8_t` in the standard. So UINT8_MAX might be better.

src/formats/PDB.cpp
Luthaf: Yep!

PR #222, opened 2019-01-07
Use the shared allocator for all C API pointer

Luthaf:
This allows to only provide chfl_free for memory release, instead of having separate functions for atoms, cell, topology, ...

I would like to incorporate this in the 0.9 release, what do you think? It feels like it belongs to the same changeset as the shared allocator.

### DIFF
--- a/scripts/ci/lint-code.py
+++ b/scripts/ci/lint-code.py
@@ -6,24 +6,30 @@ better wrappers in chemfiles
 """
 import os
 import sys
+import re
 from glob import glob
 
 ROOT = os.path.join(os.path.dirname(__file__), "..", "..")
 ERRORS = 0
 
 PARSING_FUNCTIONS = {
-    'stof': 'parse<double>',
-    'stod': 'parse<double>',
-    'stold': 'parse<double>',
-    'stoi': 'parse<long long>',
-    'stol': 'parse<long long>',
-    'stoll': 'parse<long long>',
-    'stoul': 'parse<size_t>',
-    'stoull': 'parse<size_t>',
+    # Match 'std::<...>',  '::<...>' and '<...>'
+    r"(?P<name>\b(std)?(::)?stof\b)": "parse<double>",
+    r"(?P<name>\b(std)?(::)?stod\b)": "parse<double>",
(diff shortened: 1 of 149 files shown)


scripts/ci/lint-code.py
```diff
@@ -6,24 +6,30 @@
...
-    'stof': 'parse<double>',
-    'stod': 'parse<double>',
-    'stold': 'parse<double>',
-    'stoi': 'parse<long long>',
-    'stol': 'parse<long long>',
-    'stoll': 'parse<long long>',
-    'stoul': 'parse<size_t>',
-    'stoull': 'parse<size_t>',
+    # Match 'std::<...>',  '::<...>' and '<...>'
+    r"(?P<name>\b(std)?(::)?stof\b)": "parse<double>",
```
frodofine: This is unrelated, but that's fine!

src/capi/property.cpp
```diff
@@ -97,8 +98,3 @@ extern "C" chfl_status chfl_property_get_vector3d(const CHFL_PROPERTY* const pro
         value[2] = vector[2];
     )
 }
-
-extern "C" chfl_status chfl_property_free(const CHFL_PROPERTY* property) {
```
frodofine: Why doesn't `chfl_free` return `CHFL_SUCCESS`? Is it to remove clutter?

src/capi/misc.cpp
```diff
@@ -15,6 +16,10 @@ void chemfiles::set_last_error(const std::string& message) {
     CAPI_LAST_ERROR = message;
 }
 
+extern "C" void chfl_free(const void* const object) {
```
frodofine: Might make more sense to put this in the shared allocated cpp file, but that's minor.

src/capi/property.cpp
Luthaf: Yes, and to match the signature of `free`

Luthaf:
> It looks good overall, but if this is to be included with 0.9 all of the bindings will need to be updated!

Yes, I've got the idea of doing this while working on the fortran bindings. For now, only the Rust binding have been updated to 0.9, so this should not be too much work.

PR #223, opened 2019-01-07
CML Support

frodofine:
Addresses Issue #4, not needed for 0.9 (just wanted to get this started).

Still todo:
- [x] Property support
- [x] Write support


src/formats/CML.cpp
```diff
@@ -0,0 +1,219 @@
...
+        return;
+    }
+
+    root_ = document_.child("molecule");
+    if (root_) {
+        ismulti_ = false;
+        return;
+    }
+
+    throw format_error("unsupported starting node");
```
Luthaf: We might want to mention CML in the error

src/formats/CML.cpp
```diff
@@ -0,0 +1,219 @@
...
+            frame.set("name", attribute.as_string());
+        } else {
+            warning("Unknown molecule attribute: " + name);
+        }
+    }
+
+    auto crystal_node = current.child("crystal");
+    if (crystal_node) {
+        double a, b, c, alpha, beta, gamma;
+        a = b = c = alpha = beta = gamma = 0.0;
```
Luthaf: We should initialize alpha, beta and gamma to 90

src/formats/CML.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+        ref_to_id[id] = frame.size() - 1;
+    }
+
+    for (auto bond : current.child("bondArray").children("bond")) {
+        auto atomref = bond.attribute("atomRefs2");
+        auto order = bond.attribute("order");
+
+        if (!atomref) {
+            warning("Bad bond");
```
Luthaf: This should mention CML and maybe more information on the bond

src/formats/CML.cpp
```diff
@@ -0,0 +1,219 @@
...
+            continue;
+        }
+
+        auto ids = split(atomref.as_string(), ' ');
+        if (ids.size() != 2) {
+            warning("Bad bond size");
```
Luthaf: Same, this should mention CML and maybe more information on the bond

src/formats/CML.cpp
```diff
@@ -0,0 +1,219 @@
...
+                case 'T':
+                case 't':
+                    bo = Bond::TRIPLE;
+                    break;
+                case 'A':
+                case 'a':
+                    bo = Bond::AROMATIC;
+                    break;
+                default:
+                    warning("Unknown bond order");
```
Luthaf: Could you print the bond order in the message ?

tests/formats/cml.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+    SECTION("Read next step") {
+        Trajectory file("data/cml/drugs.cml");
+        auto frame = file.read();
+        CHECK(frame.size() == 17);
+        CHECK(frame.get("name")->as_string() == "naproxen");
+
+        // Check positions
+        auto positions = frame.positions();
+        CHECK(approx_eq(positions[0]  == Vector3D(0.0000, -4.6200, 0.0000), 1e-3));
```
Luthaf: This is not doing what you think it does =) It should be `approx_eq(positions[0], Vector3D(0.0000, -4.6200, 0.0000), 1e-3)` (no ==). Else it is converting a bool to double and comparing it to 1e-3

tests/formats/cml.cpp
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    SECTION("Read a specific step") {
+        Trajectory file("data/cml/drugs.cml");
+        // Read frame at a specific positions
+        auto frame = file.read_step(2);
+        CHECK(frame.step() == 2);
+        CHECK(frame.get("name")->as_string() == "tylenol");
+        auto positions = frame.positions();
+        CHECK(approx_eq(positions[0] == Vector3D(0.0000, 6.1600, 0.0000), 1e-3));
```
Luthaf: Same as above

tests/formats/cml.cpp
```diff
@@ -0,0 +1,78 @@
...
+        auto topology = frame.topology();
+        CHECK(topology.size() == 11);
+        CHECK(topology[0].type() == "C");
+
+        frame = file.read_step(0);
+        CHECK(frame.step() == 0);
+        positions = frame.positions();
+        CHECK(approx_eq(positions[0] == Vector3D(0.0000, -4.6200, 0.0000), 1e-3));
```
Luthaf: Same as above

tests/formats/cml.cpp
```diff
@@ -0,0 +1,78 @@
...
+        CHECK(cell.a() == 3.0);
+        CHECK(cell.b() == 3.0);
+        CHECK(cell.c() == 3.0);
+        CHECK(cell.alpha() == 90.0);
+        CHECK(cell.beta() == 90.0);
+        CHECK(cell.gamma() == 120.0);
+
+        auto positions = frame.positions();
+        auto fract0 = frame.cell().matrix().invert() * positions[0];
+        CHECK(approx_eq(fract0 == Vector3D(-1.77493, 0.980333, 0.0000), 1e-3));
```
Luthaf: Same as above

tests/formats/cml.cpp
Luthaf: I fixed a few of these in #217 (https://github.com/chemfiles/chemfiles/pull/217/files#diff-ff5a736e4c9aa54737d30385699c44d2)

external/CMakeLists.txt
```diff
@@ -218,4 +218,10 @@ else()
 endif()
 list(APPEND EXTERNAL_INCLUDES ${LIBLZMA_INCLUDE_DIRS})
 
+# ==========
+# pugixml: http://github.com/chemfiles/pugixml
```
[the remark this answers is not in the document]
frodofine: I'll do that soonish.

tests/formats/cml.cpp
frodofine: Whoops! I wrote this during a plane flight, maybe I should be more careful with 'plane' code.

frodofine:
I think this is ready for a second review!

src/formats/CML.cpp
```diff
@@ -0,0 +1,581 @@
+// Chemfiles, a modern library for chemistry file reading and writing
+// Copyright (C) Guillaume Fraux and contributors -- BSD license
+
+#include <cassert>
+#include <sstream>
+
+#include <fmt/format.h>
+#include <fmt/ostream.h>
+
+#include "pugixml.cpp"
```
Luthaf: Could we create a CMake OBJECT library instead ? This way if we implement other formats wich need XML, the library is already there

src/formats/CML.cpp
```diff
@@ -0,0 +1,581 @@
...
+        root_.append_attribute("xmlns") = "http://www.xml-cml.org/schema";
+        root_.append_attribute("xmlns:cml") = "http://www.xml-cml.org/dict/cml";
+        root_.append_attribute("xmlns:units") = "http://www.xml-cml.org/units/units";
+        root_.append_attribute("xmlns:convention") = "http://www.xml-cml.org/convention";
+        root_.append_attribute("convention") = "convention:molecular";
+        root_.append_attribute("xmlns:iupac") = "http://www.iupac.org";
+        return;
+    }
+
+    if (mode_ == File::APPEND) { // may need to check the file a CML node in the future, now just append
```
Luthaf: I don't understand this comment, could you expand a bit ?

src/formats/CML.cpp
```diff
@@ -0,0 +1,581 @@
...
+        current_ = molecules.begin();
+        root_ = document_;
+        return;
+    }
+
+    throw format_error("[CML] no supported starting node found");
+}
+
+CMLFormat::~CMLFormat() {
+    if (num_added_ && mode_ == File::WRITE) {
```
Luthaf: `if (num_added_ != 0 && mode_ == File::WRITE) {` is recommanded by clang-tidy. You can run it by building the code with clang and setting `CHFL_USE_CLANG_TIDY=ON` with cmake

src/formats/CML.cpp
Luthaf: Looks like this still applies =)

src/formats/CML.cpp
```diff
@@ -0,0 +1,581 @@
...
+            }
+
+            read_property_(frame, scalar_node, title_str);
+        }
+    }
+
+    // Read the atoms
+    std::unordered_map<std::string, size_t> ref_to_id;
+
+    // TODO: Read string vector versions instead of just the nodes
```
Luthaf: Is this still TODO ?

src/formats/CML.cpp
```diff
@@ -0,0 +1,581 @@
...
+    for (auto atom : current.child("atomArray").children("atom")) {
+        double x2, y2, x3, y3, z3, xf, yf, zf;
+        x2 = y2 = x3 = y3 = z3 = xf = yf = zf = 0.0;
```
Luthaf: I would rather do this as `double x = 0, y = 0, z =0; bool is_frac = false`, and then set only x/y for x2,y2, set all three for x3/y3/z3, and set is_frac for xf/yf/zf

src/formats/CML.cpp
```diff
@@ -0,0 +1,581 @@
...
+                hydrogen_count = attribute.as_double();
+            } else {
+                warning("[CML] Unknown atom attribute: " + name);
+            }
+        }
+
+        auto new_atom = Atom(title, element);
+        Vector3D cart, velo;
+
+        if (crystal_node && (xf != 0.0 || yf != 0.0 || zf != 0.0)) {
```
Luthaf: what happens if crystal_node does not exist but xf/yf/zf are set ?

src/formats/CML.cpp
```diff
@@ -0,0 +1,581 @@
...
+        frame.add_atom(std::move(new_atom), cart, velo);
+        ref_to_id[id] = frame.size() - 1;
+    }
+
+    for (auto bond : current.child("bondArray").children("bond")) {
+        auto atomref = bond.attribute("atomRefs2");
+        auto order = bond.attribute("order");
+
+        if (!atomref) {
+            warning("[CML] Bad bond");
```
Luthaf: Could this error message be expanded a bit ?

src/formats/CML.cpp
```diff
@@ -0,0 +1,581 @@
...
+            continue;
+        }
+
+        auto ids = split(atomref.as_string(), ' ');
+        if (ids.size() != 2) {
+            warning("[CML] Bad bond size");
```
Luthaf: Same, this could read as `[CML] bondArray contains a bond of size {} instead of 2`

src/formats/CML.cpp
```diff
@@ -0,0 +1,581 @@
...
+            continue;
+        }
+
+        auto id1 = ref_to_id.find(ids[0]);
+        auto id2 = ref_to_id.find(ids[1]);
+        if (id1 == ref_to_id.end() || id2 == ref_to_id.end()) {
+            warning("[CML] Bad bond ref");
```
Luthaf: Again, could this be modified to read like `[CML] invalid atomic indexes in bond: {} -- {}`

src/formats/CML.cpp
```diff
@@ -0,0 +1,581 @@
...
+// Used for isotopes and formal charges
+static bool is_double_integer(double val) {
+    return std::abs(val - std::trunc(val)) < 1e-3;
+}
+
+void CMLFormat::write(const Frame& frame) {
+    auto mol = root_.append_child("molecule");
+
+    if (mode_ == File::WRITE) {
+        mol.append_attribute("id") = ("m" + std::to_string(++num_added_)).c_str();
```
Luthaf: Could you separate the ++num_added_ to its own line ?

tests/formats/cml.cpp
```diff
@@ -0,0 +1,198 @@
...
+    CHECK(frame2[2].get("force")->as_vector3d() == Vector3D{1., 2., 3.});
+
+    CHECK(frame2[3].charge() == 1);
+    CHECK(frame2[3].get("hydrogen_count")->as_double() == 3);
+}
+
+TEST_CASE("Append CML file") {
+    auto tmpfile = NamedTempPath(".cml");
+    const auto EXPECTED_CONTENT =
+    "<molecule title=\"appended\">\n"
```
Luthaf: You can use C++ raw strings here to remove most of the `\n` and `\"`:

```cpp
auto string = R"(foo bar "yeah"

no explicit new line \o/)";
```

src/formats/CML.cpp
frodofine: Yes, I could only find one program that created CML files in this manner and I had to modify its settings quite a bit. Other CML parsers (Avogadro, chemkit) ignore this method in any case.

src/formats/CML.cpp
frodofine: xf/yf/zf get ignored.

Luthaf: So then x3/y3/z3 get used ? Can both appear in the same file ?

tests/formats/cml.cpp
frodofine: I think a lot of other formats need this done as well! Maybe I'll make another PR for that.

src/formats/CML.cpp
frodofine: Yes, the CML standard is quite loose. You can have x2, y2, x3, y3, z3, xf, yf, and zf all defined for the same file.

src/formats/CML.cpp
frodofine: I think I'm going to leave this as is for now given that we don't know what order we could get x2/x3/xf.

Luthaf: Sounds good. I did not knew the all of them could be present in a file

PR #228, opened 2019-03-07
Fix bug in shared_allocator

Luthaf:
I wrote this while trying to fix the Windows tests for Chemfiles.jl. 

I found a bug in shared_alloc, which means we will want to do a 0.9.1 release with this patch.

### DIFF
--- a/include/chemfiles/shared_allocator.hpp
+++ b/include/chemfiles/shared_allocator.hpp
@@ -12,9 +12,6 @@
 namespace chemfiles {
 
 struct shared_metadata {
-    shared_metadata(long count_, std::function<void(void)> deleter_):
-        count(count_), deleter(std::move(deleter_)) {}
-
     /// Number of pointer sharing this reference. No need to use atomic
     /// references counting, as the allocator is protected by a mutex
     long count;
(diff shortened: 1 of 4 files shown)


Luthaf:
Hey @frodofine! 

Do you think you will have time to review this? If not I'll merge this and finish the 0.9.1 release =). No pressure though, it can wait!

include/chemfiles/shared_allocator.hpp
```diff
@@ -12,16 +12,19 @@
...
-
     /// Number of pointer sharing this reference. No need to use atomic
     /// references counting, as the allocator is protected by a mutex
     long count;
     /// How to delete the pointer when we are done with it
     std::function<void(void)> deleter;
 };
 
+namespace {
+    [[noreturn]] inline void UNINITIALIZED_DELETER() {
```
frodofine: C++ attributes!

include/chemfiles/shared_allocator.hpp
```diff
@@ -39,8 +42,9 @@ class shared_allocator {
     /// value of type T with the given arguments.
     template<class T, typename ... Args>
     static T* make_shared(Args&& ... args) {
+        auto instance = instance_.lock();
```
frodofine: Interesting, is this what fixed the bug?

Luthaf: No, this is just to make the code more precise. The main fix here is moving `pointers_.erase(it);` out of the `if (count == 0)` block in free/release. A tangential fix is not using iterators after they have been deleted ...

Luthaf:
Thank you very much!

PR #233, opened 2019-03-13
Use Travis for Windows CI

Luthaf:


Luthaf:
This cuts down the CI wait time to 45 min on Travis, instead of 2h on Appveyor + ~30 min on Travis. The main difference is that Travis allows running CI in parallel, while Appveyor runs all builders in serial.

Here is a breakdown of the run time for all configurations:

|              | Travis | Appveyor |
|--------------|:------:|:--------:|
| MSVC 64-bit  |  20min |   26min  |
| MSVC 32-bit  |  18min |   21min  |
| MSVC static  |  20min |   26min  |
| MinGW 32-bit |    /   |   15min  |
| MinGW 64-bit |  14min |   25min  |

PR #236, opened 2019-04-09
Add a cache for sub-selections

Luthaf:
Sub selections (the `name O` in `is_bonded(#1, name O)` where evaluated at every call to is_match, which is once by atom/bond/pairs in the system depending on the selection context.

This PR adds a way for selection element to cache data (more specifically a way to know when to clear the cached data), and uses it for sub selections.

On a test locally using [cfiles](https://github.com/chemfiles/cfiles), this makes the time for hydrogen bonds detection in a small trajectory go from 172s to 60s (almost three time faster!!) for the `name O and is_bonded (#1, name C) and is_bonded (#1, name H)` selection.

### DIFF
--- a/include/chemfiles/selections/expr.hpp
+++ b/include/chemfiles/selections/expr.hpp
@@ -25,6 +25,9 @@ public:
     virtual std::string print(unsigned delta = 0) const = 0;
     /// Check if the `match` is valid in the given `frame`.
     virtual bool is_match(const Frame& frame, const Match& match) const = 0;
+    /// Clear any cached data. This must be called before using the selection
+    /// with a new frame
+    virtual void clear() = 0;
     /// Optimize the AST corresponding to this Selector. Currently, this only
     /// perform constant propgations in mathematical expressions.
     virtual void optimize() {}
@@ -47,6 +50,7 @@ public:
     And(Ast lhs, Ast rhs): lhs_(std::move(lhs)), rhs_(std::move(rhs)) {}
     std::string print(unsigned delta) const override;
     bool is_match(const Frame& frame, const Match& match) const override;
+    void clear() override;
 private:
     Ast lhs_;
     Ast rhs_;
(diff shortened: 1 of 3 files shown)


include/chemfiles/selections/expr.hpp
```diff
@@ -125,6 +136,10 @@ class SubSelection {
     std::unique_ptr<Selection> selection_;
     /// Variable to use if selection_ is nullptr
     Variable variable_;
+    /// Cache matches for the selection on the first call to eval
+    mutable std::vector<size_t> matches_;
```
frodofine: Why is this marked mutable? Is it to be modified by a `const` function?

include/chemfiles/selections/expr.hpp
```diff
@@ -25,6 +25,9 @@ class Selector {
     virtual std::string print(unsigned delta = 0) const = 0;
     /// Check if the `match` is valid in the given `frame`.
     virtual bool is_match(const Frame& frame, const Match& match) const = 0;
+    /// Clear any cached data. This must be called before using the selection
```
frodofine: Is it possible to cache a reference to the original Frame and call clear when a different frame is used. Just a thought.

include/chemfiles/selections/expr.hpp
Luthaf: Yes. `Selector::is_match` is marked as const because it should behave as a const function for any external observer. Having a cache should not change this, and thus the modified fields needs to be marked as `mutable`

include/chemfiles/selections/expr.hpp
Luthaf: This would be hard, as references can not be rebound to a new object in C++. One could use std::reference_wrapper, or a raw pointer though. 

Do you have any use case in mind ?

frodofine: Nope, I was just going off that comment, It would be useful so that the user would not need to do this themself - no reason other than that.

PR #238, opened 2019-04-17
Fix PDB atom type issue

frodofine:
There's a problem in how the atom type in PDB files was being read. This leads to a confusion between CA (as in alpha-carbon) and Ca (as in calcium). Below is a quick fix.

Previously, the additional test failed as the mass of an alpha carbon was 40!


Luthaf:
Looks good!

PR #239, opened 2019-04-22
Bug fixes

frodofine:
This PR has two bug fixes. 

- The first is rather minor and was the result of a typo in the PDB format code which resulted in improper linkage between peptide residues.

- The second is more complicated and involves the removal of atoms from a topology. Currently, when an atom is removed from a topology, the residues in that topology are not updated which leads to a misalignment of indicies above the index removed as well as a 'dangling' atom index. I've fixed this in two steps: the first is adding a method to the `Residue` class when removes atoms and updates the indicies above the removed index; the second is running this in a loop during the removal of an atom from a topology.

Note: I'm willing to update the other language bindings for residue atom removal as needed.

### DIFF
--- a/include/chemfiles/Residue.hpp
+++ b/include/chemfiles/Residue.hpp
@@ -148,7 +148,16 @@ private:
     /// Additional properties of this residue
     property_map properties_;
 
+    /// Update the atomic indexes in this residue after an atom has been
+    /// removed from the containing topology.
+    ///
+    /// This function removes the atom with index `i` from this residue if it
+    /// exists, and shifts all the indexes bigger than `i` by -1.
+    void atom_removed(size_t i);
+
     friend bool operator==(const Residue& lhs, const Residue& rhs);
+
+    friend class Topology;
 };
 
 inline bool operator==(const Residue& lhs, const Residue& rhs) {
--- a/src/Residue.cpp
+++ b/src/Residue.cpp
@@ -16,3 +16,16 @@ void Residue::add_atom(size_t i) {
 bool Residue::contains(size_t i) const {
     return atoms_.find(i) != atoms_.end();
 }
+
+void Residue::atom_removed(size_t i) {
+    auto iter = atoms_.find(i);
+    if (iter != atoms_.end()) {
+        atoms_.erase(iter);
+    }
+
+    for (size_t j = 0; j < atoms_.size(); ++j) {
+        if (atoms_[j] > i) {
+            --atoms_[j];
+        }
+    }
+}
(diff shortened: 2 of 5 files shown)


Luthaf:
Thank you for sending this! the deadline for my PhD thesis being on Friday, I think I'll have time to look at it next week :smile:

frodofine:
I totally understand, I'll be in that boat next year!

My lab is actively seeking Post-docs, so if you're interesting in the cornfields of Indiana, let me know!

Luthaf:
> The second is more complicated and involves the removal of atoms from a topology. Currently, when an atom is removed from a topology, the residues in that topology are not updated which leads to a misalignment of indicies above the index removed as well as a 'dangling' atom index. I've fixed this in two steps: the first is adding a method to the Residue class when removes atoms and updates the indicies above the removed index; the second is running this in a loop during the removal of an atom from a topology.

Great bug you found there =) 

I liked the previous fact that residues were append only from the user perspective. Maybe we could keep this property by making `Residue::remove_atom` private, and the Topology class friend of the Residue class. This removes the need for new C API functions, and allow to backport the fixes to the 0.9 branch.

Also, I am slightly bothered by the fact that `remove_atom` also shifts the atomic indexes ... Maybe rename it to `atom_removed`, as in "hey residue, this atom has been removed. Please update yourself!"

frodofine:
Sounds good.

I'm a bit more concerned about the CI failures. Sphinx needs to be updated (I think Travis has changed) and this requires a newer version of python. I've already addressed this by changing the distro to Xenial instead of trusty.

Additionally, GCC5 is no longer available for the version of MacOSX used in the current script. I can't seem to get his to work again...

Luthaf:
Yeah, I have the same issues with #240 ... I could fix the Python part while staying on trusty, but the osx failure is really weird.

frodofine:
Alright, ready for review!

include/chemfiles/Residue.hpp
```diff
@@ -148,7 +148,13 @@ class CHFL_EXPORT Residue final {
     /// Additional properties of this residue
     property_map properties_;
 
+    /// Remove an atom with index `i` from this residue.
+    /// Called by the topology class when an atom is removed from the topology
```
Luthaf: How about

```cpp
/// Update the atomic indexes in this residue after an atom has been removed from the containing topology.
///
/// This function removes the atom with index `i` from this residue if it exists, and shifts all the indexes bigger than `i` by -1.
```

frodofine: Sure!

PR #243, opened 2019-05-08
Error messages uniformization

Luthaf:
This PR makes the warning and error messages more uniform, starting in lower cases and made of a single sentence. It also adds a script to ensure uniformity with future changes.

The second commit run the Python linting scripts from ctest, together with the other tests. This should remove the need to wait for CI to catch the errors.


Luthaf:
@frodofine I would like to add more context to these warnings, but I am not sure I fully understand them. What is the issue here ? 

https://github.com/chemfiles/chemfiles/blob/c2a86fd3950c02ce16ba365fcf4d99c6f2bdbffc/src/formats/PDB.cpp#L513-L527

tests/CMakeLists.txt
```diff
@@ -111,7 +111,7 @@ file(GLOB all_test_files
 
 foreach(test_file IN LISTS all_test_files)
     chfl_cpp_test(${test_file})
-endforeach(test_file)
```
frodofine: Better CMake practice!

tests/CMakeLists.txt
```diff
@@ -222,3 +222,18 @@ file(WRITE ${CMAKE_CURRENT_BINARY_DIR}/local-file.toml
 "[types]
 Oz = \"F\"
 ")
+
+# Code linting with Python scripts
+find_package(PythonInterp REQUIRED)
```
frodofine: This is deprecated [here](https://cmake.org/cmake/help/latest/module/FindPythonInterp.html).

src/formats/CML.cpp
```diff
@@ -50,29 +50,29 @@ CMLFormat::CMLFormat(std::string path, File::Mode mode, File::Compression compre
 
     auto result = document_.load(*file_);
     if (!result) {
-        throw format_error("[CML] Parsing error: '{}'", result.description());
```
frodofine: I like the idea of putting the same of the format in the error message like is done for `xzstreambuf`. This way users can know what file format an error is associated with when an error is thrown.

frodofine: I can make my own PR to add these to the messages and update the check script.

src/formats/CML.cpp
```diff
@@ -100,9 +100,9 @@ static void read_property_(T& container, pugi::xml_node& node, const std::string
     std::string data_type_str = "xsd:string";
     auto data_type = node.attribute("dataType");
     if (!data_type) {
-        warning("[CML] {} has no data type, assuming string.", title);
+        warning("{} has no data type, assuming string", title);
```
frodofine: Same comment as above. In fact, it's even more important for warnings because in this case, the program continues and this could lead to odd results downstream.

src/formats/CSSR.cpp
```diff
@@ -138,7 +138,7 @@ void CSSRFormat::write(const Frame& frame) {
     auto connectivity = std::vector<std::vector<size_t>>(frame.size());
     for (auto& bond : frame.topology().bonds()) {
         if (bond[0] > 9999 || bond[1] > 9999) {
-            warning("Atomic index is too big for connectivity record in CSSR, removing the bond");
```
frodofine: This is an alternative strategy (placing the `in CSSR`), but it could be more difficult to standardize.

tests/CMakeLists.txt
Luthaf: Since CMake 3.12, but we support cmake version all the way back to 2.8.12 (which is the one available natively on CentOS 7 and other old distro used in HPC centers). I initially used `find_package(Python REQUIRED)`, but this failed on Travis. 

I could add a cmake version check around this though.

src/formats/CML.cpp
Luthaf: I see what you mean. I kinda like the `... in <format>` or `... for <format>` approach, since it flows better. I could also live with `format: ...` as the message.

A possible way to ensure this is used everywhere would be to include a mandatory context to the warning function:

```cpp
void warning(std::string context, std::string message, Args...);

warning("cml format", "...", ..., ..., ...);
// or 
warning("cml reader", "...", ..., ..., ...);
warning("cml writer", "...", ..., ..., ...);

warning("pdb format", "...", ..., ..., ...);
```

tests/CMakeLists.txt
frodofine: That would work.

src/formats/CML.cpp
frodofine: That would work as well!

Luthaf:
OK, I think I'll merge this tomorrow, if the tests pass =)

Luthaf:
Forgot that I also wanted to update the message in `warning("{}_{} does not contain {}", residue.name(), resid, second_name);`: https://github.com/chemfiles/chemfiles/pull/243#issuecomment-490841659

@frodofine, what is the issue you are warning against here? The fact that we found a specific atom in a residue, but this atom is not in the standard PDB tables for this residue ?

frodofine:
Correct, standard residues have very specific definition. For example, `ALA` needs to have atoms `N`, `O`,  `C`, `H`, `CA`,  and `CB`.  If, for whatever reason, an `ALA` residue has an atom named `CG`, then it is non-standard and that warning is triggered.

Luthaf:
Ok, thanks !

PR #247, opened 2019-05-24
fix(README): replace outdated method in example

ezavod:
See #126. Not a big issue, but I'm just getting started and the first thing I tried didn't work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,7 +174,7 @@ int main() {
     chemfiles::Trajectory trajectory("filename.xyz");
 
     auto frame = trajectory.read();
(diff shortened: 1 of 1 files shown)


Luthaf:
Thank you very much for taking the time to do this!

Luthaf:
And let us know if you find something hard to understand in the documentation, beginners feedback is very appreciated!

ezavod:
@Luthaf Thanks for offering your help and the fast response. I'm mainly interested in the rust bindings. Just noticed that there are similar issues. I could do a PR there if your interested because I just got that example to work.

Luthaf:
> I could do a PR there if your interested because I just got that example to work.

Please do!

PR #248, opened 2019-06-12
Misc fixes

frodofine:
- Fixed a bug in reading PDB X coordinate

- Fixed a bug in reading PDB residues

- Fixed a bug when a MOL2 file does not begin with @TRIPOS<MOLECULE>

- Added support for SDF charges

### DIFF
--- a/include/chemfiles/formats/PDB.hpp
+++ b/include/chemfiles/formats/PDB.hpp
@@ -46,10 +46,12 @@ private:
                         const std::string& sec);
     // Read CONECT record
     void read_CONECT(Frame& frame, const std::string& line);
+    // Runs when a chain is terminated to update residue information
+    void chain_ended(Frame& frame);
 
     std::unique_ptr<TextFile> file_;
-    /// Map of residues, indexed by residue id and chainid.
-    std::map<std::pair<char, size_t>, Residue> residues_;
+    /// Map of residues, indexed by residue id, chain id and insertion code.
+    std::map<std::tuple<char, size_t, char>, Residue> residues_;
     /// Storing the positions of all the steps in the file, so that we can
     /// just `seekg` them instead of reading the whole step.
     std::vector<std::streampos> steps_positions_;
--- a/include/chemfiles/formats/PDB.hpp
+++ b/include/chemfiles/formats/PDB.hpp
@@ -50,8 +50,13 @@ private:
     void chain_ended(Frame& frame);
 
     std::unique_ptr<TextFile> file_;
-    /// Map of residues, indexed by residue id, chain id and insertion code.
-    std::map<std::tuple<char, size_t, char>, Residue> residues_;
+
+    /// This typedef represent the 'full' name of a residue, this means the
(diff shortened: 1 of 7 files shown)


include/chemfiles/formats/PDB.hpp
```diff
@@ -46,10 +46,12 @@ class PDBFormat final: public Format {
...
     // Read CONECT record
     void read_CONECT(Frame& frame, const std::string& line);
+    // Runs when a chain is terminated to update residue information
+    void chain_ended(Frame& frame);
 
     std::unique_ptr<TextFile> file_;
-    /// Map of residues, indexed by residue id and chainid.
-    std::map<std::pair<char, size_t>, Residue> residues_;
+    /// Map of residues, indexed by residue id, chain id and insertion code.
+    std::map<std::tuple<char, size_t, char>, Residue> residues_;
```
Luthaf: The travis failure might be related to std::tuple not having std::hash implementation. Maybe you can replace the tuple by a small struct `residue_index` ?

frodofine: It's a C++ version thing (I use 14 by default by accident). I like using a tuple because I get the comparison operator for free. I may need to write my own anyway to fix how secondary structures are parsed as the use of `' '` is a hack.

frodofine:
I can add some additional test cases for bad SDF files as well.

frodofine:
I'll merge this tomorrow if there are no complaints!

src/formats/SDF.cpp
```diff
@@ -40,6 +40,10 @@ SDFFormat::SDFFormat(std::string path, File::Mode mode, File::Compression compre
             steps_positions_.push_back(position);
         }
     }
+
+    if (mode == File::READ && steps_positions_.empty()) {
```
Luthaf: Why is this needed ? `Trajectory::read` should throw an error instead if the file contains 0 steps

tests/formats/pdb.cpp
```diff
@@ -228,10 +273,11 @@ TEST_CASE("Read files in PDB format") {
 
     SECTION("Read atomic insertion codes") {
         auto frame = Trajectory("data/pdb/insertion-code.pdb").read();
+        auto& topo = frame.topology();
 
-        CHECK(frame[0].get<Property::STRING>("insertion_code").value() == "a");
-        CHECK(frame[1].get<Property::STRING>("insertion_code").value() == "c");
-        CHECK(frame[2].get<Property::STRING>("insertion_code").value() == "x");
+        CHECK(topo.residue_for_atom(0)->get("insertion_code")->as_string() == "a");
```
Luthaf: Looks like this is now a residue property instead of an atomic property, this should be updated in the docs!

frodofine: It was already a residue property! This test was wrong!

src/formats/SDF.cpp
frodofine: Ok, I've changed this.

PR #249, opened 2019-06-17
fix(tng): Correct Conversion of Length Units

ezavod:
At the moment only the box is converted from nm to angtrom. However all length dependent units should be converted. Especially positions and velocities.
Furthermore TNG does not always use nm. In fact one could specify an arbitrary exponent for length units. This is handled by these changes.

Relevant GROMACS source: https://github.com/gromacs/gromacs/blob/ac628532b96f338ace10fb78a01d47e109b81bb8/src/gromacs/fileio/tngio.cpp#L1258-L1283

**TODO:**
 - [x] update tests

### DIFF
--- a/include/chemfiles/formats/TNG.hpp
+++ b/include/chemfiles/formats/TNG.hpp
@@ -27,6 +27,8 @@ private:
 
     /// Reference to the associated file
     TNGFile tng_;
+    /// Scale factor for all lenght dependent data e.g. positions, velocities, forces, box shape
+    float distance_scale_factor_ = -1;
     /// The next step to read
     int64_t step_ = 0;
     /// The number of atoms in the current frame
--- a/src/formats/TNG.cpp
+++ b/src/formats/TNG.cpp
@@ -45,6 +45,11 @@ TNGFormat::TNGFormat(std::string path, File::Mode mode, File::Compression compre
     if (compression != File::DEFAULT) {
         throw format_error("TNG format do not support compression");
     }
+
+    int64_t exp = -1;
+    CHECK(tng_distance_unit_exponential_get(tng_, &exp));
+    // calculate the scale factor from a given length scale to angstrom
+    distance_scale_factor_ = pow(10.0, exp + 10.0);
 }
 
 size_t TNGFormat::nsteps() {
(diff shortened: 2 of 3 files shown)


include/chemfiles/formats/TNG.hpp
```diff
@@ -27,6 +27,8 @@ class TNGFormat final: public Format {
 
     /// Reference to the associated file
     TNGFile tng_;
+    /// Scale factor for all lenght dependent data e.g. positions, velocities, forces, box shape
+    float distanceScaleFactor_ = -1;
```
Luthaf: Could you use the `snake_case` instead? This is the main convention in chemfiles =)

src/formats/TNG.cpp
```diff
@@ -45,6 +45,11 @@ TNGFormat::TNGFormat(std::string path, File::Mode mode, File::Compression compre
     if (compression != File::DEFAULT) {
         throw format_error("TNG format do not support compression");
     }
+
+    int64_t exp = -1;
+    CHECK(tng_distance_unit_exponential_get(tng_, &exp));
```
Luthaf: Is this guaranteed to be set in tng files?

Luthaf:
Looks good!

> haven't figured out how to build the tests yet

You can activate the tests with `CHFL_BUILD_TESTS`, either on the command line with `cmake -DCHFL_BUILD_TESTS=ON`, or with the curse TUI `ccmake`. If you get an error doing this let me know!

include/chemfiles/formats/TNG.hpp
ezavod: Sure, done.

src/formats/TNG.cpp
ezavod: `tng_io` [says](https://github.com/chemfiles/tng/blob/c44bdcd5554d498aed93527e9f7c5c1f77b160cd/src/lib/tng_io.c#L296-L299) it is using nm as default. The corresponding value is set [here](https://github.com/chemfiles/tng/blob/c44bdcd5554d498aed93527e9f7c5c1f77b160cd/src/lib/tng_io.c#L8788) when initializing a new file. So this value should be set always.

ezavod:
Thanks for pointing me to the test flag, I somehow overlooked that. I'll have a look at the tests soon.

Luthaf:
Thank you very much!

Luthaf:
Under which name should I add you to the [AUTHORS](https://github.com/chemfiles/chemfiles/blob/master/AUTHORS) file? `ezavod <ezavod@gmx.de>` or something else?

ezavod:
Thanks for deeming my contribution noteworthy and asking. For this profession related project `Len Kimms <len.kimms@wwu.de>` would be suitable.

PR #253, opened 2019-06-18
Initial SMI support

### LINKED ISSUE #241: feature request: Supporting SMILES format
`SMILES` is an easy-to-use format for chemical structure input. I think it should be supported.
[`SMILES` on Wikipedia](https://en.wikipedia.org/wiki/Simplified_molecular-input_line-entry_system)
[OpenSMILES specification](http://opensmiles.org/opensmiles.html)
[SMILES tokens](http://www.dalkescientific.com/writings/diary/archive/2004/01/05/tokens.html)
[Daylight SMILES](http://www.daylight.com/smiles/)

frodofine:
Warning: this code is a bit rough and fairly difficult to read. It will need to be refactored and all I can really guarantee is that is won't crash - probably.

Write support is terrible as it cannot current write rings, this will need to be fixed before it is merged.

Fixes #241

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,7 @@ Chemfiles can read and write files in the following formats:
 - TRIPOS mol2 files;
 - MacroMolecular Crystallographic Information Framework (CIF) files;
 - MacroMolecular Transmission Format (MMTF) files;
+- Daylight SMILES: .smi;
 
 For text-based formats (XYZ, PDB, LAMMPS Data, Tinker, ...) Chemfiles will
 transparently read compressed files if the file extension matches the expected
--- a/include/chemfiles/Connectivity.hpp
+++ b/include/chemfiles/Connectivity.hpp
@@ -30,6 +30,10 @@ public:
         QINTUPLET = 5,
 
         // space for more bond types if needed
+        DOWN = 250,
+        UP = 251,
+        DATIVER = 252,
+        DATIVEL = 253,
         AMIDE = 254,
         AROMATIC = 255,
     };
--- a/include/chemfiles/Connectivity.hpp
+++ b/include/chemfiles/Connectivity.hpp
@@ -32,8 +32,8 @@ public:
         // space for more bond types if needed
         DOWN = 250,
         UP = 251,
-        DATIVER = 252,
-        DATIVEL = 253,
+        DATIVE_R = 252,
+        DATIVE_L = 253,
         AMIDE = 254,
         AROMATIC = 255,
     };
--- a/include/chemfiles/Connectivity.hpp
+++ b/include/chemfiles/Connectivity.hpp
@@ -22,20 +22,20 @@ public:
 
     /// Stores the type of bond
     enum BondOrder {
-        UNKNOWN = 0,
-        SINGLE = 1,
-        DOUBLE = 2,
-        TRIPLE = 3,
-        QUADRUPLE = 4,
-        QINTUPLET = 5,
+        UNKNOWN = 0,    ///< Bond order is unknown or unspecified
+        SINGLE = 1,     ///< Single bond
+        DOUBLE = 2,     ///< Double bond
+        TRIPLE = 3,     ///< Triple bond
+        QUADRUPLE = 4,  ///< Qudadruple bond
+        QINTUPLET = 5,  ///< Quintuplet bond 
 
         // space for more bond types if needed
-        DOWN = 250,
-        UP = 251,
-        DATIVE_R = 252,
-        DATIVE_L = 253,
-        AMIDE = 254,
-        AROMATIC = 255,
+        DOWN = 250,     ///< Single bond direction from frist atom to second is 'down'. Used for cis-trans isomers
+        UP = 251,       ///< Single bond direction from first atom to second is 'up'. Used for cis-trans isomers
+        DATIVE_R = 252, ///< Dative bond where the electrons are localized to the first atom
+        DATIVE_L = 253, ///< Dative bond where the electrons are localized to the second atom
+        AMIDE = 254,    ///< Amide bond (C(=O)-NH)
+        AROMATIC = 255, ///< Aromatic bond (for example the ring bonds in benzene)
     };
 
     /// Create a new `Bond` containing the atoms `i` and `j`.
--- a/include/chemfiles/formats/SMI.hpp
+++ b/include/chemfiles/formats/SMI.hpp
@@ -4,8 +4,10 @@
 #ifndef CHEMFILES_FORMAT_SMI_HPP
 #define CHEMFILES_FORMAT_SMI_HPP
 
+#include <vector>
 #include <stack>
 #include <map>
+#include <unordered_map>
 
 #include "chemfiles/Format.hpp"
 #include "chemfiles/File.hpp"
@@ -33,11 +35,11 @@ private:
     std::vector<std::streampos> steps_positions_;
 
     /// [for reading] Stores location of a branching path
-    std::stack<size_t> branch_point_;
+    std::stack<size_t, std::vector<size_t>> branch_point_;
 
     /// [for reading] Stores a mapping between a ring ID and the atom which
     /// starts the ring and a stored bond order
-    std::map <size_t, std::pair<size_t, Bond::BondOrder>> rings_ids_;
+    std::unordered_map <size_t, std::pair<size_t, Bond::BondOrder>> rings_ids_;
 
     /// [for reading] The current atom being added (active atom)
     size_t current_atom_;
@@ -56,7 +58,7 @@ private:
     bool first_atom_;
 
     /// [for reading] adds an atom defined by `atom_name` to the topology
-    Atom& add_atom(Topology& topo, const std::string& atom_name);
+    Atom& add_atom(Topology& topo, std::string atom_name);
 
     /// [for reading] adds an atom defined by the string `smiles` starting at position i
     void process_property_list_(Topology& topo, const std::string& smiles, size_t& i);
@@ -67,14 +69,15 @@ private:
     /// [for writing] stores how many branches need to be closed
     size_t branch_stack_;
 
-    /// [for writing] stores locations of ring closures
+    /// [for writing] stores locations of ring closures. This is ordered
+    /// to ensure rings get printed in numeric order (where possible)
     std::multimap<size_t, size_t> ring_stack_;
 
     /// [for writing] stores how many rings we need to close
     size_t ring_count_;
 
     /// [for writing] stores how many rings each atom is in
-    std::map<size_t, size_t> ring_atoms_;
+    std::unordered_map<size_t, size_t> ring_atoms_;
 
     /// [for writing] writes a single atom to the internal file
     /// this function is recursive
--- a/include/chemfiles/formats/SMI.hpp
+++ b/include/chemfiles/formats/SMI.hpp
@@ -32,28 +32,52 @@ private:
     /// just `seekg` them instead of reading the whole step.
     std::vector<std::streampos> steps_positions_;
 
-    /// Stores location of a branching path
+    /// [for reading] Stores location of a branching path
     std::stack<size_t> branch_point_;
 
-    /// Stores a mapping between a ring ID and the atom which starts the ring
-    /// and a stored bond order
+    /// [for reading] Stores a mapping between a ring ID and the atom which
+    /// starts the ring and a stored bond order
     std::map <size_t, std::pair<size_t, Bond::BondOrder>> rings_ids_;
 
-    /// The current atom being added (active atom)
+    /// [for reading] The current atom being added (active atom)
     size_t current_atom_;
 
-    /// The previous active atom
+    /// [for reading] The previous active atom
     size_t previous_atom_;
 
-    /// The current bond ordeer
+    /// [for reading] The current bond ordeer
     Bond::BondOrder current_bond_order_;
 
-    /// Should we connect the previous atom to the first atom
+    /// [for reading] Should we connect the previous atom to the first atom
+    /// [for writing] Should we add a '.' after the current molecule
     bool first_atom_;
 
+    /// [for reading] adds an atom defined by `atom_name` to the topology
     Atom& add_atom(Topology& topo, const std::string& atom_name);
 
+    /// [for reading] adds an atom defined by the string `smiles` starting at position i
     void process_property_list_(Topology& topo, const std::string& smiles, size_t& i);
+
+    /// [for writing] stores the graph of the topology
+    std::vector<std::vector<size_t>> adj_list_;
+
+    /// [for writing] stores how many branches need to be closed
+    size_t branch_stack_;
+
+    /// [for writing] stores locations of ring closures
+    std::multimap<size_t, size_t> ring_stack_;
+
+    /// [for writing] stores how many rings we need to close
+    size_t ring_count_;
+
+    /// [for writing] stores how many rings each atom is in
+    std::map<size_t, size_t> ring_atoms_;
+
+    /// [for writing] writes a single atom to the internal file
+    /// this function is recursive
+    void write_atom(const Frame& frame,
+        std::vector<bool>& hit_atoms,
+        size_t current_atom, size_t previous_atom);
 };
 
 template<> FormatInfo format_information<SMIFormat>();
--- a/include/chemfiles/formats/SMI.hpp
+++ b/include/chemfiles/formats/SMI.hpp
@@ -1,0 +1,63 @@
+// Chemfiles, a modern library for chemistry file reading and writing
+// Copyright (C) Guillaume Fraux and contributors -- BSD license
+
+#ifndef CHEMFILES_FORMAT_SMI_HPP
+#define CHEMFILES_FORMAT_SMI_HPP
+
+#include <stack>
+#include <map>
+
+#include "chemfiles/Format.hpp"
+#include "chemfiles/File.hpp"
+#include "chemfiles/Connectivity.hpp"
+#include "chemfiles/Topology.hpp"
+
+namespace chemfiles {
+
+/// [SMI] file format reader and writer.
+///
+/// [SMI]: http://opensmiles.org/opensmiles.html
+class SMIFormat final: public Format {
+public:
+    SMIFormat(const std::string& path, File::Mode mode, File::Compression compression);
+
+    void read_step(size_t step, Frame& frame) override;
+    void read(Frame& frame) override;
+    void write(const Frame& frame) override;
+    size_t nsteps() override;
+private:
+    /// Text file where we read from
+    std::unique_ptr<TextFile> file_;
+    /// Storing the positions of all the steps in the file, so that we can
+    /// just `seekg` them instead of reading the whole step.
+    std::vector<std::streampos> steps_positions_;
+
+    /// Stores location of a branching path
+    std::stack<size_t> branch_point_;
+
+    /// Stores a mapping between a ring ID and the atom which starts the ring
+    /// and a stored bond order
+    std::map <size_t, std::pair<size_t, Bond::BondOrder>> rings_ids_;
+
+    /// The current atom being added (active atom)
+    size_t current_atom_;
+
+    /// The previous active atom
+    size_t previous_atom_;
+
+    /// The current bond ordeer
+    Bond::BondOrder current_bond_order_;
+
+    /// Should we connect the previous atom to the first atom
+    bool first_atom_;
+
+    Atom& add_atom(Topology& topo, const std::string& atom_name);
+
+    void process_property_list_(Topology& topo, const std::string& smiles, size_t& i);
+};
+
+template<> FormatInfo format_information<SMIFormat>();
+
+} // namespace chemfiles
+
+#endif
--- a/include/chemfiles/formats/SMI.hpp
+++ b/include/chemfiles/formats/SMI.hpp
@@ -48,6 +48,9 @@ private:
     /// [for reading] The current bond ordeer
     Bond::BondOrder current_bond_order_;
 
+    /// [for reading] The
+    std::vector<Residue> mol_vector_;
+
     /// [for reading] Should we connect the previous atom to the first atom
     /// [for writing] Should we add a '.' after the current molecule
     bool first_atom_;
--- a/src/formats/SMI.cpp
+++ b/src/formats/SMI.cpp
@@ -4,6 +4,7 @@
 #include <set>
 #include <cctype>
 #include <map>
+#include <unordered_set>
 #include <algorithm>
 
 #include <fmt/format.h>
@@ -62,23 +63,23 @@ void SMIFormat::read_step(const size_t step, Frame& frame) {
 static size_t read_number(const std::string& smiles, size_t& start) {
     size_t number = 0;
     if (start >= smiles.size() || std::isdigit(smiles[start]) == 0) {
-        --start;
+        start--;
         return number;
     }
     do {
         number *= 10;
         number += static_cast<size_t>(smiles[start] - '0');
-        ++start;
+        start++;
     } while (start < smiles.size() && (std::isdigit(smiles[start]) != 0));
-    --start;
+    start--;
     return number;
 }
 
-static std::set<char> aromatic_organic {'b', 'c', 'n', 'o', 's', 'p'};
+static std::unordered_set<char> AROMATIC_ORGANIC {'b', 'c', 'n', 'o', 's', 'p'};
 
 // Note: Including H here is non-standard, but allows for HC(H)(H)H
-static std::set<std::string> aliphatic_organic {"B", "C", "N", "O", "S", "P",
-                                                "F", "Cl", "Br", "I", "H"};
+static std::unordered_set<std::string> ALIPHATIC_ORGANIC {"B", "C", "N", "O", "S", "P",
+                                                          "F", "Cl", "Br", "I", "H"};
 
 // Valid chirality tags
 static std::set<std::string> chirality_tags {"TH", "SP", "TB", "OH", "AL"};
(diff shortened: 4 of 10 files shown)


Luthaf:
The code does not seems to compile with MSVC.

As this is a bit dense, I don't think I'll have time to review it until next week, after my defence =)

frodofine:
No problem, I know the code is dense and I will try and simplify it over time. It was 'ported' over from a different project which has a different coding style (I also removed all the comments, which I will re-add).

The MSVC failure is due to a missing include (cctype).

frodofine:
Alright, this is ready for an *initial* review.  Portions may be rough, but I tried to comment those sections.

README.md
```diff
@@ -108,6 +108,7 @@ Chemfiles can read and write files in the following formats:
 - TRIPOS mol2 files;
 - MacroMolecular Crystallographic Information Framework (CIF) files;
 - MacroMolecular Transmission Format (MMTF) files;
+- Daylight SMILES: .smi;
```
Luthaf: Since we are implementing OpenSMILES, should this be the name we use?

include/chemfiles/formats/SMI.hpp
```diff
@@ -0,0 +1,90 @@
...
+    std::multimap<size_t, size_t> ring_stack_;
+
+    /// [for writing] stores how many rings we need to close
+    size_t ring_count_;
+
+    /// [for writing] stores how many rings each atom is in
+    std::map<size_t, size_t> ring_atoms_;
+
+    /// [for writing] writes a single atom to the internal file
+    /// this function is recursive
```
Luthaf: If the function is recursive, we may want to check that the number of atoms in the frame will not cause a stack overflow.

How hard would it be to rewrite the function as a while loop?

tests/formats/smi.cpp
```diff
@@ -0,0 +1,296 @@
...
+            // We can throw either when creating the trajectory, or when reading
+            // the frame, depending on the type of error
+            auto file = Trajectory(entry.path().string());
+            file.read();
+        };
+        CHECK_THROWS(test());
+    }
+}
+
+TEST_CASE("Write SMI File") {
```
Luthaf: What is the behavior of the writer if there are multiple molecules in the frame? For example, how is `H-O-H       H-O-H` written? We may want a test for this case.

src/formats/SMI.cpp
```diff
@@ -0,0 +1,705 @@
...
+    case Bond::TRIPLE: fmt::print(*file_, "#"); break;
+    case Bond::QUADRUPLE: fmt::print(*file_, "$"); break;
+    case Bond::AROMATIC: fmt::print(*file_, ":"); break;
+    case Bond::DATIVEL: fmt::print(*file_, "<-"); break;
+    case Bond::DATIVER: fmt::print(*file_, "->"); break;
+    case Bond::UP: fmt::print(*file_, "/"); break;
+    case Bond::DOWN: fmt::print(*file_, "\\"); break;
+    case Bond::UNKNOWN:
+    default:
+        fmt::print(*file_, "~");
```
Luthaf: How about using `~` for UNKNOWN bonds, and adding a warning in the default case, to future proof this a bit?

src/formats/SMI.cpp
```diff
@@ -0,0 +1,705 @@
...
+    return ring_atoms;
+}
+
+static void write_atom_smiles(std::unique_ptr<TextFile>& file_, const Atom& atom) {
+    bool needs_brackets = false;
+    auto type = atom.type();
+
+    // The mass must be an integer is the only check we need as all atoms in the
+    // periodic table do have non-integer masses. Therefore, if the the mass is
+    // not an integer, then we know the user has set an isotope
```
Luthaf: I am not sure I understand this comment. Pure atoms have integer mass, as well as isotopes. But the masses in `periodic_table.cpp` are not integers, since they are the average value with the natural isotopes occurrences.

src/formats/SMI.cpp
```diff
@@ -0,0 +1,705 @@
...
+    double mass;
+    size_t mass_int = 0;
+    if (std::modf(atom.mass(), &mass) == 0.0 && mass != 0.0) {
+        needs_brackets = true;
+        mass_int = static_cast<size_t>(mass);
```
Luthaf: Could this use std::floor or std::ceil to be more explicit? I believe the cast behave as std::floor, but I am not 100% sure.

src/formats/SMI.cpp
```diff
@@ -0,0 +1,705 @@
...
+    }
+
+    // If any of these values are set / not a default value
+    auto smi_class = atom.get("smiles_class");
+    auto chirality = atom.get<Property::STRING>("chirality").value_or("");
+    auto explicit_h =atom.get<Property::DOUBLE>("hydrogen_count").value_or(0.0);
```
Luthaf: missing a space after the `=`

src/formats/SMI.cpp
```diff
@@ -0,0 +1,705 @@
...
+
+    // Charge is similar to mass. It must be an integer, otherwise it is ignored
+    double charge;
+    int charge_int = 0;
+    if (std::modf(atom.charge(), &charge) == 0.0 && charge != 0.0) {
+        needs_brackets = true;
+        charge_int = static_cast<int>(charge);
```
Luthaf: Same as above with std::ceil/std::floor. Maybe also add a warning to inform the user that the charge changed below them.

src/formats/SMI.cpp
```diff
@@ -0,0 +1,705 @@
...
+    }
+
+    // Before the atom is printed, we must know if we are printing a property set.
+    if (explicit_h != 0.0 || smi_class || !chirality.empty() ||
+        aliphatic_organic.count(type) == 0) {
```
Luthaf: On the readability side, could the condition go on the above line? Or add some separation with the next line?

src/formats/SMI.cpp
```diff
@@ -0,0 +1,705 @@
...
+    // Mass must be first, before the element is printed
+    if (mass_int != 0) {
+        fmt::print(*file_, "{}", mass_int);
+    }
+
+    fmt::print(*file_, "{}", type);
+
+    if (smi_class && smi_class->kind() == Property::DOUBLE) {
+        fmt::print(*file_, ":{}", static_cast<int>(smi_class->as_double()));
+    }
```
Luthaf: Add a warning if "smile_class" is set but not a double? Why don't you use `atom.get<Property::DOUBLE>` instead?

src/formats/SMI.cpp
```diff
@@ -0,0 +1,705 @@
...
+
+    if (chirality != "") {
+        // If it's clockwise, then we just print the symbol @@ and be done with it
+        if (chirality.find("CW") == 0) {
```
Luthaf: Is `--- CW foobar` a valid chirality flag? I think this code would accept it.

src/formats/SMI.cpp
```diff
@@ -0,0 +1,705 @@
...
+            fmt::print(*file_, "@@");
+        } else {
+            fmt::print(*file_, "@");
+            // TODO check if the chirality flag is valid instead of trusting the user
+            if (chirality.size() > 4) {
```
Luthaf: Maybe inform the user that we are dropping the flag if it is smaller than 4 characters?

src/formats/SMI.cpp
Luthaf: Maybe make explicit_h an integer, and warn the user if the value is not an integer?

src/formats/SMI.cpp
```diff
@@ -0,0 +1,705 @@
...
+    frame.set_topology(topol);
+
+    if (i < smiles.size()) {
+        auto name = smiles.substr(i);
+        frame.set("name", trim(name));
+    }
+}
+
+/// Depth first search to find all rings.
+/// Not gaurenteed to fast, nor efficient, but just to work
```
Luthaf: guaranteed

src/formats/SMI.cpp
```diff
@@ -0,0 +1,705 @@
...
+                                                "F", "Cl", "Br", "I", "H"};
+
+// Valid chirality tags
+static std::set<std::string> chirality_tags {"TH", "SP", "TB", "OH", "AL"};
+
+static std::string read_property_atom(const std::string& smiles, size_t& i) {
+    auto old = i;
+    if (smiles[i] == '\'') {
+        do {
+             ++i;
```
Luthaf: same as above concerning `++i` vs `i++` :smiley:

src/formats/SMI.cpp
```diff
@@ -0,0 +1,705 @@
...
+        return smiles.substr(old + 1, i - old - 2);
+    } else {
+        do {
+            ++i;
+        } while( i < smiles.size() && std::islower(smiles[i]) != 0);
+        return smiles.substr(old, i - old);
+    }
+}
+
+Atom& SMIFormat::add_atom(Topology& topol, const std::string& atom_name) {
```
Luthaf: you can pass atom_name by value and move it to construct the new atom (and move it too at the call site)

src/formats/SMI.cpp
```diff
@@ -0,0 +1,705 @@
...
+    }
+
+    first_atom_ = false;
+    previous_atom_ = current_atom_;
+    current_bond_order_ = Bond::SINGLE;
+    mol_vector_.back().add_atom(topol.size() - 1);
+    return topol[topol.size() - 1];
+}
+
+void SMIFormat::process_property_list_(Topology& topol, const std::string& smiles, size_t& i) {
```
Luthaf: `i` is the start of the string here? Should we rename it to `start` then? (here and in other similar functions)

src/formats/SMI.cpp
```diff
@@ -0,0 +1,705 @@
...
+    double mass = 0.0;
+    if (std::isdigit(smiles[i]) != 0) {
+        mass = static_cast<double>(read_number(smiles, i));
+        ++i;
+    }
+
+    bool is_aromatic = std::islower(smiles[i]) != 0;
```
Luthaf: std::islower uses the current user locale, which may break the code here for non-ascii atom names. We may want to write our own. There are already uses of such functions from cctype in the code, so this can be another PR.

src/formats/SMI.cpp
```diff
@@ -0,0 +1,705 @@
...
+            warning("SMI Reader", "unknown property code: '{}'", smiles[i]);
+            break;
+        }
+        ++i;
+    }
+}
+
+void SMIFormat::read(Frame& frame) {
+    // Initialize all the reading variables
+    branch_point_ = std::stack<size_t>();
```
Luthaf: why not `branch_point_.clear()`?

src/formats/SMI.cpp
```diff
@@ -0,0 +1,705 @@
...
+    size_t i;
+    for (i = 0; i < smiles.size(); ++i) {
+
+        if (std::isblank(smiles[i]) != 0) {
+            break;
+        }
+
+        if (smiles[i] == '[') {
+            // TODO: Don't pass the full string, instead pass an iterator
+            // or maybe a substring till the closing ']'
```
Luthaf: we really need a C++11 port of std::string_view =/

src/formats/SMI.cpp
```diff
@@ -0,0 +1,705 @@
...
+            if (aromatic_organic.count(smiles[i]) == 0) {
+                throw format_error("SMI Reader", "aromatic element not found: '{}'", smiles[i]);
+            }
+            add_atom(topol, smiles.substr(i, 1)).set("is_aromatic", true);
+            continue;
+        }
+
+        // We are not in a property list!
+        // Therefore, aliphatic atoms can be written without brackets.
+        // Br and Cl are tricky as, but the rest are single character
```
Luthaf: Should we account for `At` too as a two letter halogen element?

Luthaf:
>  I can't comment on the standard compliance.

There seems to be a formal grammar for OpenSMILES. This means the reader could use a full recursive-descent style parser. I don't know whether it is worth the trouble, but this would give us full compliance with the spec.

include/chemfiles/Connectivity.hpp
```diff
@@ -30,6 +30,10 @@ class CHFL_EXPORT Bond final {
         QINTUPLET = 5,
 
         // space for more bond types if needed
+        DOWN = 250,
```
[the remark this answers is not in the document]
frodofine: `DATIVER` and `DATIVEL` are RDKit extensions to allow for dative bonds. These bonds describe situations where the bonding electrons are localized to one atom, for example, a ligand coordinating to a metal center. This allows for `(N->)[Co+6](<-N)(<-N)(<-N)(<-N)(<-N)`. I'll rename it to `DATIVE_R` and `DATIVE_L` for clarity.

`UP` and `DOWN` are for storing cis/trans isomers and rely on both position and bond type.

tests/formats/smi.cpp
```diff
@@ -0,0 +1,207 @@
...
+    SECTION("Read a specific step") {
+        Trajectory file("data/smi/test.smi");
+        auto frame = file.read_step(1);
+        CHECK(frame.size() == 6);
+        auto topol = frame.topology();
+        CHECK(topol.bonds().size() == 6);
+    }
+
+    SECTION("Read entire file") {
+        Trajectory file("data/smi/rdkit_problems.smi");
```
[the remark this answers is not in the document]
frodofine: RDKit is BSD licensed.

src/formats/SMI.cpp
```diff
@@ -0,0 +1,412 @@
...
+#include "chemfiles/File.hpp"
+#include "chemfiles/Frame.hpp"
+#include "chemfiles/periodic_table.hpp"
+#include "chemfiles/warnings.hpp"
+
+using namespace chemfiles;
+
+template<> FormatInfo chemfiles::format_information<SMIFormat>() {
+    return FormatInfo("SMI").with_extension(".smi").description(
+        "SMI text format"
```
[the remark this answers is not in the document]
frodofine: No, SMI - I believe. The SMI format is `<SMILES_STRING>\t<NAME>\n` where each line is a different compound. What defines `<SMILES_STRING>` is the standard I linked.

src/formats/SMI.cpp
```diff
@@ -0,0 +1,412 @@
...
+#include "chemfiles/ErrorFmt.hpp"
+#include "chemfiles/File.hpp"
+#include "chemfiles/Frame.hpp"
+#include "chemfiles/periodic_table.hpp"
+#include "chemfiles/warnings.hpp"
+
+using namespace chemfiles;
+
+template<> FormatInfo chemfiles::format_information<SMIFormat>() {
+    return FormatInfo("SMI").with_extension(".smi").description(
```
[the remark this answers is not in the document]
frodofine: See above.

src/formats/SMI.cpp
```diff
@@ -0,0 +1,412 @@
...
+    }
+
+    fmt::print(*file_, "\n");
+}
+
+bool forward(TextFile& file) {
+    if (!file) {return false;}
+
+    try {
+        auto line = file.readline();
```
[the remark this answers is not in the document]
frodofine: Yes, although this could be changed.

src/formats/SMI.cpp
```diff
@@ -0,0 +1,412 @@
...
+    file_->seekg(steps_positions_[step]);
+    read(frame);
+}
+
+/// Attempts to read a number from the string. The `start` argument is updated
+/// to the last numberic character in `smiles`.
+static size_t read_number(const std::string& smiles, size_t& start) {
+    size_t number = 0;
+    if (start >= smiles.size() || std::isdigit(smiles[start]) == 0) {
+        --start;
```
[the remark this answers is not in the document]
frodofine: Prefix feels more natural to me because I "read" it as 'decrease start' and 'start decrease' feels odd. I've changed it in any case.

README.md
frodofine: `Daylight SMILES` is what most people who use SMILES refer to it as. I used OpenSMILES as a link because their description of the format is more cohesive and complete. We technically can't call our parser `OpenSMILES` because we don't check to see if the aromaticity is correct and we accept whatever the user gives.

include/chemfiles/formats/SMI.hpp
frodofine: Potentially, I originally wrote this as a while loop while prototyping, but switched to a recursive function instead as this is typically how graph algorithms are written (in my experience). I'd need to rethink branching, but it is certainly doable, but probably in a different PR or at a later date.

tests/formats/smi.cpp
frodofine: That would be `O.O.O`, I've added a test case.

src/formats/SMI.cpp
frodofine: Sorry, the comment should have ended with 'Therefore, if the mass **is** an integer,  then we know the user has set an isotope.'

src/formats/SMI.cpp
frodofine: Sure, but it doesn't matter. This code is only executed when the double is actually an integer (IE zeros after the period), for example, 35.0000.

src/formats/SMI.cpp
frodofine: I'm not sure a warning is needed. The code is only executed if the charge is something like *1.000*, or *-2.0000*. If the charge is *0.434* or something, nothing happens (SMILES cannot store non-integer charges).

src/formats/SMI.cpp
frodofine: I might make that 'is the double actually an integer' idea its own function, that should clear things up.

src/formats/SMI.cpp
frodofine: It would accept `CW foobar` (the check is to see if the string begins with `CW`). I put a note in the documentation that we do not check chirality - at all. So if the user sets something funny, they'll get a funny result. I can add warnings, however.

src/formats/SMI.cpp
frodofine: See comment about making this a function.

src/formats/SMI.cpp
frodofine: Hmmm. SMILES strings shouldn't have non-ascii atom names! Definitely a separate PR.

src/formats/SMI.cpp
frodofine: That function doesn't exist!

src/formats/SMI.cpp
frodofine: Agreed

src/formats/SMI.cpp
frodofine: Not according to the standard, but I break from the standard in many places anyway!

src/formats/SMI.cpp
```diff
@@ -0,0 +1,412 @@
...
+
+void SMIFormat::read_step(const size_t step, Frame& frame) {
+    assert(step < steps_positions_.size());
+    file_->seekg(steps_positions_[step]);
+    read(frame);
+}
+
+/// Attempts to read a number from the string. The `start` argument is updated
+/// to the last numberic character in `smiles`.
+static size_t read_number(const std::string& smiles, size_t& start) {
```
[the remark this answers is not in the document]
frodofine: The main reason was to ensure that the `start` reference got updated correctly.

include/chemfiles/formats/SMI.hpp
```diff
@@ -0,0 +1,63 @@
+// Chemfiles, a modern library for chemistry file reading and writing
+// Copyright (C) Guillaume Fraux and contributors -- BSD license
+
+#ifndef CHEMFILES_FORMAT_SMI_HPP
+#define CHEMFILES_FORMAT_SMI_HPP
+
+#include <stack>
+#include <map>
```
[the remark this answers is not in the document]
frodofine: In one case, yes. I've added a comment.

frodofine:
I've addressed (hopefully) all of your comments. Thank you for reviewing this as I know it is a lot of code to support something 'simple'. I don't think a recursive-descent parser is needed, especially since I break from the standard in a few places anyway. There's an extension to smiles created by ChemAxon that should be implemented (eventually) that we do not support and could break such a parser because I don't think its grammar is as well defined.

src/formats/SMI.cpp
Luthaf: > I can add warnings, however.

This would be nice, to remove possible sources of confusion :smiley:

include/chemfiles/formats/SMI.hpp
Luthaf: I am fine with a different PR, could you create an issue so that we don't forget to do it?

src/formats/SMI.cpp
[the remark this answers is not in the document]
Luthaf: sound good ! :+1:

src/formats/SMI.cpp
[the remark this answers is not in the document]
Luthaf: no I mean you can write a wrapper around std::stoi to correctly update start without writing your own integer parsing code =)

src/formats/SMI.cpp
```diff
@@ -0,0 +1,737 @@
...
+
+static void write_atom_smiles(std::unique_ptr<TextFile>& file_, const Atom& atom) {
+    bool needs_brackets = false;
+    auto type = atom.type();
+
+    // The mass must be an integer is the only check we need as all atoms in the
+    // periodic table have non-integer masses. Therefore, if the the mass is
+    // an integer, then we know the user has set an isotope.
+    int mass_int = 0;
+    if (double_to_int(atom.mass(), mass_int)) {
```
Luthaf: If the mass is say `12.5`, this will completely ignore it, right?

src/formats/SMI.cpp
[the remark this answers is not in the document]
Luthaf: How often are smi files edited by hand? If they are, I would argue that is it better to ignore empty lines altogether.

tests/formats/smi.cpp
```diff
@@ -0,0 +1,306 @@
...
+    frame.add_atom(Atom("C"), {0, 0, 0});
+    frame[4].set("chirality", "CW");
+
+    frame.add_bond(1, 2, Bond::DATIVE_R);
+    frame.add_bond(2, 3, Bond::DATIVE_L);
+    frame.add_bond(3, 4, Bond::DOWN);
+
+    file.write(frame);
+
+    // Reinitialize and test for descrete molecules
```
Luthaf: discrete? (I think)

src/formats/SMI.cpp
frodofine: Correct. This format does not support non-integer masses.

tests/formats/smi.cpp
frodofine: Oops, not the biggest deal though!

src/formats/SMI.cpp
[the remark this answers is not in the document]
frodofine: No, they are typically created by software such as MarvinSketch or ChemDraw.

src/formats/SMI.cpp
Luthaf: Should we have a warning for this? The unfortunate issue I can see here is that we will set non-integer masses when creating Atom (see: https://github.com/chemfiles/chemfiles/blob/bc375d2d5fdc613b9085f31920946598ab7ebe01/src/periodic_table.cpp#L13 we set the hydrogen mass to 1.008). 

At the same time we don't want to output a mass of 1 for all hydrogen atoms, but we don't want specific masses set by the users to be ignored silently.

frodofine: Many formats do not support floating point masses (SDF/CML/MOl2/HIN comes to mind) so I think there should be a larger PR addressing this issue.

Luthaf: Ok, maybe open an issue so that we remember to do it?

frodofine:
Okay, this is ready for a second review.

src/formats/SMI.cpp
```diff
@@ -0,0 +1,766 @@
...
+
+        // To print a start bracket, we need to be branching (> 2 non-ring bonds)
+        // and we don't want to brank the last neighbor printed
+        if (neighbors_printed - ring_start < current_atom_bonds.size() - 2 &&
+            current_atom_bonds.size() > 2) {
+            fmt::print(*file_, "(");
+            branch_stack_++;
+        }
+
+        // DFS like recursion
```
Luthaf: DFS => Depth-first search?

src/formats/SMI.cpp
```diff
@@ -0,0 +1,766 @@
...
+            if (neigh_ring_iter == ring_atoms.end()) {
+                neigh_ring_iter = ring_atoms.insert({neighbor, 0}).first;
+            }
+            ++neigh_ring_iter->second;
+
+            continue; // No need to see this atom again
+        }
+
+        // Continue the search
+        find_rings_helper(adj_list, hit_atoms, ring_bonds,
```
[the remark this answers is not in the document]
frodofine: Yeah, I did that for my readability.

src/formats/SMI.cpp
```diff
@@ -0,0 +1,766 @@
...
+            if (ring_bonds.count(Bond(neighbor, current_atom))) {
+                continue;
+            }
+            ring_bonds.insert(Bond(neighbor, current_atom));
+
+            auto neigh_ring_iter = ring_atoms.find(neighbor);
+            if (neigh_ring_iter == ring_atoms.end()) {
+                neigh_ring_iter = ring_atoms.insert({neighbor, 0}).first;
+            }
+            ++neigh_ring_iter->second;
```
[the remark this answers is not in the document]
frodofine: All three are equivalent, I made it consistent.

frodofine:
I'll merge this later today unless there are any complaints.

frodofine:
> Is this done? I don't completely understand the behavior of the code.

Yep! That's done. I had to add the whole ring finding code to do it though.

PR #259, opened 2019-07-08
Unify text format and speed-up finding steps offset

Luthaf:
The first part of this PR unify all the text formats (except mmCIF, because the logic to find offsets is a bit complex) under a new abstract class `TextFormat`. This new class is responsible for finding the "start of step" offsets in the file.

The second part add new functions to TextFile to just skip a (few) lines. This is useful when looking for offset, as we don't have to allocate memory for string we immediately throw away.

---

Using the new code, I get up to 3x speedup with a C++ benchmark.

This PR was motivated by https://github.com/MDAnalysis/mdanalysis/pull/1862#issuecomment-509199512, as it looks like chemfiles is slower than MDAnalysis when reading XYZ.

With this PR, I get nice gains with a small file containing a lot of frames:
```python
In [4]: mda.coordinates.XYZ.XYZReader('./water.xyz')
Out[4]: <XYZReader ./water.xyz with 100 frames of 297 atoms>

In [5]: %timeit mda.coordinates.XYZ.XYZReader('./water.xyz')
1.08 ms ± 48.3 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

# Before
In [8]: %timeit chemfiles.ChemfilesReader('./water.xyz')
15.4 ms ± 215 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

# After
In [2]: %timeit chemfiles.ChemfilesReader('./water.xyz')
5.09 ms ± 84.8 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

But not so much for bigger files with a single frame.
```python
In [3]: mda.coordinates.XYZ.XYZReader('./big.xyz')
Out[3]: <XYZReader ./big.xyz with 1 frames of 29997 atoms>

In [4]: %timeit mda.coordinates.XYZ.XYZReader('./big.xyz')
55.3 ms ± 1.79 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

# Before
In [3]: %timeit chemfiles.ChemfilesReader('./big.xyz', format='XYZ')
87.2 ms ± 999 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

# After
In [5]: %timeit chemfiles.ChemfilesReader('./big.xyz', format='XYZ')
74.6 ms ± 680 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

We are still slower than MDAnalysis, but there is also quite a lot of overhead when doing C++ => C => Python (chemfiles) => Python (MDAnalysis).

Overall, this PR is still a win and I think worth to merge.


Luthaf:
@richardjgowers, this might interest you =) the improvement should be in the next version (0.9.2)

Luthaf:
So I can confirm that locally, a C++ program using chemfiles is as fast as MDAnalysis when reading an XYZ file. 

Using chemfiles from MDAnalysis is still slower, but the difference must come from the conversion from C++ to C to Python.

include/chemfiles/Format.hpp
```diff
@@ -28,7 +34,7 @@ class CHFL_EXPORT Format {
     Format(Format&&) = delete;
     Format& operator=(Format&&) = delete;
 
-    /// @brief Read a specific step from the trajectory file.
+    /// Read a specific `step` from the trajectory file.
```
frodofine: This is extremely nit-picky, but I prefer *italics* for concepts that are not code. Since there is no method called `step` I would use italics instead of backticks - but that's extremely nit-picky.

include/chemfiles/Format.hpp
```diff
@@ -37,23 +43,25 @@ class CHFL_EXPORT Format {
     /// @param frame The frame to fill
     virtual void read_step(size_t step, Frame& frame);
 
-    /// @brief Read a specific step from the trajectory file.
+    /// Read the next step from the trajectory file.
```
frodofine: Should it be `step` instead of step?

include/chemfiles/File.hpp
```diff
@@ -90,6 +94,7 @@ class CHFL_EXPORT TextFile: public File, public std::iostream {
 
 private:
     void get_line(std::string& string);
+    void skip_line();
```
frodofine: `skip_line` and `skipline` seem very close in name to me. Is this intentional?

include/chemfiles/formats/CSSR.hpp
```diff
@@ -17,19 +17,13 @@ class Frame;
 /// writer. Only one frame can be read or written to this format.
 ///
 /// [CSSR]: http://www.chem.cmu.edu/courses/09-560/docs/msi/modenv/D_Files.html#944777
-class CSSRFormat final: public Format {
+class CSSRFormat final: public TextFormat {
 public:
     CSSRFormat(std::string path, File::Mode mode, File::Compression compression);
 
-    void read_step(size_t step, Frame& frame) override;
```
frodofine: This is much cleaner! I link the new interface!

src/File.cpp
```diff
@@ -74,24 +75,45 @@ void TextFile::get_line(std::string& string) {
     }
 }
 
+void TextFile::skip_line() {
+    std::istream::sentry sentry(*this, true);
+    std::streambuf* buffer = std::istream::rdbuf();
```
frodofine: This line is a bit confusing. I believe that you are calling the function `rdbuf` in a parent class, but this is not obvious.

src/File.cpp
```diff
@@ -110,7 +132,40 @@ std::vector<std::string> TextFile::readlines(size_t n) {
...
+
     return lines;
 }
+
+void TextFile::skipline() {
+    // Disable exceptions checking, and manually check bellow
+    auto state = this->exceptions();
+    this->exceptions(std::fstream::goodbit);
+
+    skip_line();
```
frodofine: If the intent is to make `skip_line` private, would `skipline_` be a better name? This matches the convention used by private class members.

src/formats/CSSR.cpp
```diff
@@ -40,24 +40,18 @@ template<> FormatInfo chemfiles::format_information<CSSRFormat>() {
...
-}
-
-void CSSRFormat::read_step(const size_t step, Frame& frame) {
-    assert(step == 0);
-    file_->rewind();
-    read(frame);
-}
+void CSSRFormat::read_next(Frame& frame) {
+    if (file_->tellg() != 0) {
+        throw format_error("CSSR format only support reading one frame");
```
frodofine: Should be 'supports'

src/formats/CSSR.cpp
```diff
@@ -189,3 +183,15 @@ void CSSRFormat::write(const Frame& frame) {
         fmt::print(*file_, " {:7.3f}\n", frame[i].charge());
     }
 }
+
+std::streampos CSSRFormat::forward() {
```
frodofine: Yep, really like the new interface!

src/formats/LAMMPSData.cpp
```diff
@@ -773,38 +770,38 @@ size_t DataTypes::improper_type_id(size_t type_i, size_t type_j, size_t type_k,
...
+        throw format_error("LAMMPS Data format only support writting one frame");
     }
-    written_ = true;
-    types_ = DataTypes(frame.topology());
 
-    write_header(frame);
-    write_types();
-    write_masses();
-    write_atoms(frame);
+    auto types = DataTypes(frame.topology());
```
frodofine: No longer a member function, explains all the changes!

src/formats/MOL2.cpp
```diff
@@ -263,22 +229,22 @@ std::streampos forward(TextFile& file) {
                 nbonds = parse<size_t>(counts[1]);
             }
 
-            read_until(file, "@<TRIPOS>ATOM");
-            file.readlines(natoms);
+            read_until(*file_, "@<TRIPOS>ATOM");
```
frodofine: Very welcome improvement!

src/formats/SMI.cpp
```diff
@@ -756,17 +729,23 @@ void SMIFormat::write(const Frame& frame) {
...
+std::streampos SMIFormat::forward() {
+    if (!*file_) {
+        return std::streampos(-1);
+    }
 
+    auto position = file_->tellg();
     try {
-        auto line = file.readline();
+        auto line = file_->readline();
+        while (trim(line).empty()) {
```
frodofine: This is to skip blank lines?

include/chemfiles/File.hpp
Luthaf: yeah, this is somehow the same as `getline` and `get_line` =) I could rename them to `get_line_impl` and `skip_line_impl` if you prefer!

src/File.cpp
Luthaf: yes, this is calling the function on the parent std::istream. I could also write it as `this->rdbuf()` instead.

include/chemfiles/Format.hpp
Luthaf: The idea is more to use backticks for parameter to the function, describing them directly in the main description.

include/chemfiles/Format.hpp
Luthaf: with backticks? I don't think so, see the answer above.

src/formats/SMI.cpp
Luthaf: Yes, I had an issue without it where too many steps would be detected, so I went with this instead.

If you don't agree with it, I can revert this!

Luthaf:
>  have you tried a pure C++ benchmark? That should be faster than Python!

Yes, the pure C++ benchmark is as fast as Python, not faster. This makes sense, as all Python is doing here is calling into the standard library (written in C) to read lines and parse floats. So I would not expect high gains by simply using c++, even more so that this code is IO bound, not really CPU bound.

> You mentioned that the C++ benchmark is 3x faster, but I just want to make sure!

That's 3x faster than current master =)

include/chemfiles/Format.hpp
frodofine: Ok, that's fine!

include/chemfiles/Format.hpp
frodofine: Makes sense now.

include/chemfiles/File.hpp
frodofine: Yeah, that's fine as long as the public API makes sense.

src/File.cpp
frodofine: That would be preferable as it is easier to know what is going on.

src/formats/SMI.cpp
frodofine: Nope, that's fine - just confirming.

frodofine:
Ok, that makes sense. XYZ files are quite simple, so there's probably little to gain from a C++ vs Python.

PR #260, opened 2019-07-10
Fix TNG iteration bug

### LINKED ISSUE #242: TNG iteration
Iterating a TNG file doesn't seem to work

```python
import chemfiles

t = chemfiles.Trajectory('./cobro.tng')

for frame in t:

    print(frame)

Frame with 19385 atoms

/home/richard/miniconda3/envs/chemfiles/lib/python3.6/site-packages/chemfiles/trajectory.py:57: ChemfilesWarning: Error while calling tng_util_pos_read_range(tng_, step_, step_, buffer.ptr(), &unused) in the TNG library
  yield self.read_step(step)

---------------------------------------------------------------------------
ChemfilesError                            Traceback (most recent call last)
<ipython-input-77-a83bd13609e1> in <module>
----> 1 for frame in t:
      2     print(frame)

~/miniconda3/envs/chemfiles/lib/python3.6/site-packages/chemfiles/trajectory.py in __iter__(self)
     55         self.__check_opened()
     56         for step in range(self.nsteps):
---> 57             yield self.read_step(step)
     58 
     59     def __repr__(self):

~/miniconda3/envs/chemfiles/lib/python3.6/site-packages/chemfiles/trajectory.py in read_step(self, step)
     77         self.__check_opened()
     78         frame = Frame()
---> 79         self.ffi.chfl_trajectory_read_step(self, c_uint64(step), frame)
     80         return frame
     81 

~/miniconda3/envs/chemfiles/lib/python3.6/site-packages/chemfiles/_utils.py in _check_return_code(status, _function, _arguments)
     82     """Check that the function call was OK, and raise an exception if needed"""
     83     if status.value != 0:
---> 84         raise ChemfilesError(_last_error())
     85 
     86 

ChemfilesError: Error while calling tng_util_pos_read_range(tng_, step_, step_, buffer.ptr(), &unused) in the TNG library
```

I think this is probably because the TNG library expects frames to be indexed against the MD frame (ie if you run 200 frames, saving every 10 frames, you'll need to load frame 0, 10, 20 etc), whereas chemfiles counts frames as 1, 2, 3?

Luthaf:
Using @ezavod  fix for the tng library in https://github.com/chemfiles/tng/commit/9ade626ce79ed38e9a42f00332f256b504bce3f8, this fixes the bug in #242 while keeping existing files working.

fix #242


src/Trajectory.cpp
```diff
@@ -23,6 +23,8 @@
 
 using namespace chemfiles;
 
+#define SENTINE_VALUE (static_cast<size_t>(-1))
```
frodofine: `SENTINE` or `SENTINEL`?

scripts/boost.sh
```diff
@@ -26,7 +26,7 @@ INCLUDES=(
 
     iterator range config detail exception functional predef system type_traits
     core io utility filesystem mpl preprocessor smart_ptr asio date_time bind
-    regex
+    regex container_hash
```
frodofine: I don't see `container_hash` being included. Is it needed?

frodofine: Upon a second look, it appears as though the version of boost was updated. I'm curious as to what bug was fixed as it could be affecting another project of mine.

src/formats/TNG.cpp
```diff
@@ -71,20 +71,49 @@ TNGFormat::TNGFormat(std::string path, File::Mode mode, File::Compression compre
     CHECK(tng_distance_unit_exponential_get(tng_, &exp));
     // calculate the scale factor from a given length scale to angstrom
     distance_scale_factor_ = pow(10.0, static_cast<double>(exp) + 10.0);
+
+    // work around a bug in tng_num_frames_get (https://redmine.gromacs.org/issues/2937)
```
frodofine: I won't argue with the GROMACS devs!

src/formats/TNG.cpp
```diff
@@ -71,20 +71,49 @@ TNGFormat::TNGFormat(std::string path, File::Mode mode, File::Compression compre
...
+
+        if (status == TNG_SUCCESS) {
+            current_frame = next_frame;
+            tng_steps_.push_back(current_frame);
+        } else if (status == TNG_FAILURE) {
+            // We found the end of the file
+            break;
+        } else {
+            // TNG_CRITICAL: throw error
+            CHECK(status);
```
frodofine: I assume this throws then!

scripts/boost.sh
frodofine: Looks like `converter/cast.hpp` is also needed.

Luthaf: no, this was an issue on my side, I deleted the corresponding commit. I had boost installed system-wide at a different version, and it got picked by the compiler instead of the packaged one.

I'll try to see how I can prevent this from happening again =)

src/formats/TNG.cpp
Luthaf: you are right, I made it more explicit =)

PR #262, opened 2019-09-03
feat(XTC): replace Molfile with our own implementation

ezavod:
Use `xdrfile` to make use of offsets for `read_step`. Eliminates the large memory usage.
Ticks a box in #154.


src/files/XTCFile.cpp
```diff
@@ -0,0 +1,63 @@
...
+#define CHECK(x) check_xtc_error((x), (STRING(x)))
+
+XTCFile::XTCFile(std::string path, File::Mode mode)
+    : File(std::move(path), mode, File::DEFAULT), handle_(nullptr) {
+    const char* openmode;
+    if (mode == File::READ) {
+        openmode = "r";
+        CHECK(read_xtc_header(this->path().c_str(), &natoms_, &nframes_, &offsets_));
+    } else if (mode == File::WRITE || mode == File::APPEND) {
+        throw chemfiles::FileError("XTC format does not support writing yet");
```
Luthaf:
```suggestion
        throw file_error("XTC format does not support writing yet");
```

src/files/XTCFile.cpp
```diff
@@ -0,0 +1,63 @@
...
+    case exdrFLOAT:
+    case exdrUINT:
+    case exdr3DX:
+    case exdrCLOSE:
+    case exdrMAGIC:
+    case exdrNOMEM:
+    case exdrENDOFFILE:
+    case exdrFILENOTFOUND:
+    case exdrNR:
+        throw chemfiles::FileError("error while calling " + function +
```
Luthaf: You can use `file_error` and formatting here

src/files/XTCFile.cpp
```diff
@@ -0,0 +1,63 @@
...
+                                   " in the XTC library: " + std::string(exdr_message[status]));
+    case exdrOK:
+        // Do nothing, this is good
+        break;
+    default:
+        throw chemfiles::FileError("unknown status code from XTC library: " +
```
Luthaf: You can use file_error and formatting here

src/files/XTCFile.cpp
```diff
@@ -0,0 +1,63 @@
...
+        throw chemfiles::FileError("XTC format does not support writing yet");
+    }
+
+    handle_ = xdrfile_open(this->path().c_str(), openmode);
+    if (!handle_) {
+        throw file_error("could not open the file at {}", this->path());
+    }
+}
+
+XTCFile::~XTCFile() { xdrfile_close(handle_); }
```
Luthaf: This is missing a `free(offsets_)`

include/chemfiles/files/XTCFile.hpp
```diff
@@ -0,0 +1,47 @@
...
+class XTCFile final : public File {
+  public:
+    XTCFile(std::string path, File::Mode mode);
+    ~XTCFile() override;
+    XTCFile(XTCFile&&) noexcept = default;
+    XTCFile& operator=(XTCFile&&) = default;
+    XTCFile(XTCFile const&) = delete;
+    XTCFile& operator=(XTCFile const&) = delete;
+
+    unsigned long get_nframes();
```
Luthaf: You can just call the function `nframes` (or even `nsteps`), and it should be const.

It is also missing documentation. Something like "Get the number of frames/steps in the file, as indicated in the file header"

include/chemfiles/files/XTCFile.hpp
```diff
@@ -0,0 +1,47 @@
...
+    int64_t get_offset(int64_t step);
```
Luthaf: You can just call the function `offset`, and it should be const. It is also missing documentation.

include/chemfiles/files/XTCFile.hpp
```diff
@@ -0,0 +1,47 @@
...
+    int get_natoms();
```
Luthaf: You can just call the function `natoms`, and it should be const. It is also missing documentation.

include/chemfiles/files/XTCFile.hpp
```diff
@@ -0,0 +1,47 @@
...
+#define CHEMFILES_XTC_FILE_HPP
+
+#include "chemfiles/File.hpp"
+#include "xdrfile.h"
+#include "xdrfile_xtc.h"
+
+namespace chemfiles {
+
+/// Simple RAII capsule for `XDRFILE*`, handling the creation and
+/// destruction of the file as needed.
```
Luthaf: This could also mention that the class reads the file header to get frames offset (either here or in the constructor documentation)

src/formats/XTC.cpp
```diff
@@ -0,0 +1,88 @@
...
+    CHECK(xdr_seek(xtc_, xtc_.get_offset(step_), SEEK_SET));
+    read(frame);
+}
+
+void XTCFormat::read(Frame& frame) {
+    int natoms = xtc_.get_natoms();
+    int md_step = 0;
+    float time = 0;
+    matrix box;
+    rvec* x = new rvec[natoms];
```
Luthaf: This can be replaced by using a std::vector and using [std::vector::data](https://en.cppreference.com/w/cpp/container/vector/data) when calling the C function. This removes the need for manual memory management (the current version of the code is using free on a `new` allocated object, which is UB)

src/formats/XTC.cpp
```diff
@@ -0,0 +1,88 @@
...
+    float precision = 0;
+
+    CHECK(read_xtc(xtc_, natoms, &md_step, &time, box, x, &precision));
+
+    frame.set("md_step", md_step); // actual step of MD Simulation
```
Luthaf: Could you add these to the properties documentations (https://github.com/chemfiles/chemfiles/blob/a291902c2208ffc5f96bb9b0b08861a4f62a0542/doc/src/properties/frame.toml)

src/formats/XTC.cpp
```diff
@@ -0,0 +1,88 @@
...
+    frame.set("time", time);       // time in pico seconds
+    frame.set("xtc_precision", precision);
+    frame.resize(static_cast<size_t>(natoms));
+
+    set_positions(x, frame);
```
Luthaf: Why make these functions instead of witting the code inline?

src/files/XTCFile.cpp
ezavod: Done.

src/files/XTCFile.cpp
ezavod: That's kind of embarrassing. My short term memory is bad =)

include/chemfiles/files/XTCFile.hpp
ezavod: Added docs and made const.

src/formats/XTC.cpp
ezavod: Thought it would be better for readability especially because there is some conversion done. That might look messy when inline. Should they be inline?

include/chemfiles/files/XTCFile.hpp
ezavod: Done

include/chemfiles/files/XTCFile.hpp
ezavod: Done

include/chemfiles/files/XTCFile.hpp
ezavod: Done.

src/formats/XTC.cpp
ezavod: Hope that I've done this correct. Anything else, that should be mentioned in de description of the attributes?

src/formats/XTC.cpp
ezavod: This doesn't seem to work. I guess that's because the 3D array doesn't have a default value and this breaks stuff. Added proper delete for now.

Luthaf: Isn't rvec just a typedef to `float[3]`? If so, you can create a std::vector<float> of size 3*natoms, and use that.

src/formats/XTC.cpp
Luthaf: Not necessarily, I was just wondering =)

Luthaf:
Travis failure is linked to xdrfile not compiling in C89 mode. We can change the standard to C99 here, or update the code to be ANSI C compliant. I would rather go for the second choice, since the code might be useful to others

src/formats/XTC.cpp
ezavod: Works now, must have messed something up yesterday.

Luthaf: > It must be vector of type float[3].

You can use `reinterpret_cast<float[3]*>(float*)`

doc/src/properties/frame.toml
```diff
@@ -32,3 +32,19 @@ MMTF = "Date (YYYY-MM-DD format) of the deposition in the PDB. Read from the ``d
 type = "string"
 
 CML = "The text of the ``title`` attribute in a molecule node. Only set if the attribute is present."
+
+[md_step]
+type = "number"
+
+XTC = "Actual step of the MD simulation."
```
Luthaf: "MD simulation" => "simulation" or "molecular dynamics simulation"

ezavod: Done.

src/formats/XTC.cpp
ezavod: Nice little trickery, didn't know about that =)

Luthaf:
I updated the CHANGELOG for this =)

PR #263, opened 2019-09-09
Small formats fix

Luthaf:


src/formats/mmCIF.cpp
```diff
@@ -82,27 +82,27 @@ mmCIFFormat::mmCIFFormat(std::string path, File::Mode mode, File::Compression co
             in_loop = false;
         }
 
-        if (line_split[0] == "_cell_length_a") {
+        if (line_split[0] == "_cell_length_a" || line_split[0] == "_cell.length_a") {
```
frodofine: Ok, `_cell.length_a` is new to me.

Luthaf: It seems to be standard, see http://mmcif.wwpdb.org/dictionaries/mmcif_pdbx_v50.dic/Categories/cell.html

PR #264, opened 2019-09-09
fix(tests): missing fabs in approx_eq

ezavod:
Unfortunately this breaks tests for mmcif format. The test file contains the correct entries like `_cell.length_a`, but chemfiles only seems to parse deprecated entries `_cell_length_a`.
Don't really know how to fix this easily, because I've never used this format :(

### DIFF
--- a/tests/external/helpers.cpp
+++ b/tests/external/helpers.cpp
@@ -61,7 +61,7 @@ bool approx_eq(const chemfiles::Matrix3D& lhs, const chemfiles::Matrix3D& rhs, d
 }
 
 bool approx_eq(double a, double b, double tolerance) {
-    return (a - b) < tolerance;
+    return fabs(a - b) < tolerance;
 }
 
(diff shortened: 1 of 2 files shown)


Luthaf:
> The test file contains the correct entries like _cell.length_a, but chemfiles only seems to parse deprecated entries _cell_length_a.

I am pretty sure I had this fixed somewhere ... That's the problem with hoarding on branches!

Luthaf:
Found it!: https://github.com/chemfiles/chemfiles/pull/263/commits/e759976c8763b28b1af11d1896d13f8840b2dcec

ezavod:
> Found it!

I'm kinda lost, but good that this is already taken care of :+1:

Luthaf:
Sorry, the commit in question was burried in an old branch of mine and I just cherry-picked it into #263. Since I just merged it, you can rebase this on top of master and it should Just Work™

PR #265, opened 2019-09-10
feat(TRR): replace Molfile with our own implementation

ezavod:
Use xdrfile to make use of offsets for read_step. Eliminates the large memory usage. This also supports reading velocities.
Ref. #154.


include/chemfiles/files/TRRFile.hpp
```diff
@@ -0,0 +1,51 @@
...
+    ~TRRFile() override;
+    TRRFile(TRRFile&&) noexcept = default;
+    TRRFile& operator=(TRRFile&&) = default;
+    TRRFile(TRRFile const&) = delete;
+    TRRFile& operator=(TRRFile const&) = delete;
+
+    /// get the number of frames/steps in the file, as indicated in the file header
+    unsigned long nframes() const;
+    /// get the offset corresponding to a specific frame/step
+    int64_t offset(int64_t step) const;
```
Luthaf: Shoul the step be a `size_t` value rather than a signed integer?

Luthaf: The same apply to XTCFile =)

src/FormatFactory.cpp
```diff
@@ -38,7 +39,6 @@ using namespace chemfiles;
 
 namespace chemfiles {
     extern template class Molfile<DCD>;
-    extern template class Molfile<TRR>;
```
Luthaf: You can also remove the corresponding value in MolfileFormat enum

src/formats/TRR.cpp
```diff
@@ -0,0 +1,117 @@
...
+    std::vector<float> v(static_cast<size_t>(natoms) * 3);
+    std::vector<float> f(static_cast<size_t>(natoms) * 3);
+    uint8_t has_prop = 0;
+    bool has_box = false;
+    bool has_positions = false;
+    bool has_velocities = false;
+
+    CHECK(read_trr(trr_, natoms, &md_step, &time, &lambda, box,
+                   reinterpret_cast<float(*)[3]>(x.data()), reinterpret_cast<float(*)[3]>(v.data()),
+                   NULL, &has_prop));
```
Luthaf: NULL => nullptr

Luthaf: Could you add an inline comment specifying what is this parameter?

src/formats/TRR.cpp
```diff
@@ -0,0 +1,117 @@
...
+    int natoms = trr_.natoms();
+    int md_step = 0;
+    float time = 0;
+    float lambda = 0;
+    matrix box;
+    std::vector<float> x(static_cast<size_t>(natoms) * 3);
+    std::vector<float> v(static_cast<size_t>(natoms) * 3);
+    std::vector<float> f(static_cast<size_t>(natoms) * 3);
+    uint8_t has_prop = 0;
+    bool has_box = false;
```
Luthaf: You don't need to declare variables here, you can just declare on first use below

doc/src/properties/frame.toml
```diff
@@ -36,15 +36,23 @@ CML = "The text of the ``title`` attribute in a molecule node. Only set if the a
...
 type = "number"
 
 XTC = """The precision used to compress the coordinates. Only used for ten or more atoms.
 Default is 1000."""
+
+[has_positions]
+type = "bool"
+
+TRR = """Set to ``true`` if the frame contains positions. All positions are zero if not.
+This is a workaround, because chemfiles does not support optional positions yet."""
```
Luthaf: I would remove the "This is a workaround, because chemfiles does not support optional positions yet.", since it could be interpreted as "We are going to support optional positions in the future", which is not certain =)

src/formats/TRR.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+    CHECK(read_trr(trr_, natoms, &md_step, &time, &lambda, box,
+                   reinterpret_cast<float(*)[3]>(x.data()), reinterpret_cast<float(*)[3]>(v.data()),
+                   NULL, &has_prop));
+
+    has_box = bool(has_prop & TRR_HAS_BOX);
+    has_positions = bool(has_prop & TRR_HAS_POSITIONS);
+    has_velocities = bool(has_prop & TRR_HAS_VELOCITIES);
+
+    frame.set("md_step", md_step); // actual step of MD Simulation
```
Luthaf: You can use `frame.set_step` for this, since this is the intended usage of it =). The same applies to XTC

include/chemfiles/files/TRRFile.hpp
ezavod: Done, also for XTC.

src/FormatFactory.cpp
ezavod: Done for both formats.

src/formats/TRR.cpp
ezavod: Done. This is for forces. If we give a `NULL/nullptr`, they won't be parsed.

doc/src/properties/frame.toml
ezavod: Sure, it wasn't my intention to signal changes. Just thought it might be helpful to explain why the interface is different between velocities and positions :)

src/formats/TRR.cpp
ezavod: Done for both, was confused, because I thought step in chemfiles was counted differently.

src/formats/TRR.cpp
ezavod: Done for the `bool` variables.

src/formats/TRR.cpp
Luthaf: >  was confused, because I thought step in chemfiles was counted differently.

src/formats/Molfile.cpp
```diff
@@ -335,13 +333,7 @@ template<> FormatInfo chemfiles::format_information<Molfile<DCD>>() {
...
 template<> FormatInfo chemfiles::format_information<Molfile<TRJ>>() {
     return FormatInfo("TRJ").with_extension(".trj").description(
-        "GROMACS .xtc binary format"
-    );
-}
-
-template<> FormatInfo chemfiles::format_information<Molfile<XTC>>() {
-    return FormatInfo("XTC").with_extension(".xtc").description(
-        "GROMACS .xtc binary compressed portable format"
+        "GROMACS .trj binary format"
```
Luthaf: great typo spotting =). BTW, do you also plan in replacing the TRJ reader?

ezavod: Thanks. I'm only working with XTC/TRR trajectories so reading them was somewhat of a high priority for me. I probably should stop procrastinating for a while, so I wouldn't have time for this. :/ But if I have a bit more time in a while and nobody else did it until then I could give it a try.

Luthaf: ok =)

PR #267, opened 2019-09-10
Use string_view instead of string where possible

Luthaf:
This PR improves the performance of text format by around 20%. Using [hyperfine](https://github.com/sharkdp/hyperfine) on `xyz` and `lammps-data` tests, i get

**before**
```
$ hyperfine --warmup 3 ./xyz ./lammps-data
Benchmark #1: ./xyz
  Time (mean ± σ):      72.1 ms ±   2.0 ms    [User: 64.2 ms, System: 4.9 ms]
  Range (min … max):    69.5 ms …  77.3 ms    40 runs

Benchmark #2: ./lammps-data
  Time (mean ± σ):      92.0 ms ±   2.7 ms    [User: 82.9 ms, System: 5.6 ms]
  Range (min … max):    87.2 ms …  98.2 ms    31 runs
```

**After**
```
$ hyperfine --warmup 3 ./xyz ./lammps-data
Benchmark #1: ./xyz
  Time (mean ± σ):      56.2 ms ±   1.5 ms    [User: 48.5 ms, System: 4.6 ms]
  Range (min … max):    53.4 ms …  60.4 ms    51 runs

Benchmark #2: ./lammps-data
  Time (mean ± σ):      74.8 ms ±   1.9 ms    [User: 65.8 ms, System: 5.2 ms]
  Range (min … max):    72.0 ms …  79.2 ms    38 runs
```

-----

The first part of the PR introduces a C++11 implementation of C++17 `std::string_view` (https://github.com/martinmoene/string-view-lite) in chemfiles, and uses it for `split`/`trim`, removing a lot of small strings allocations.

The second part introduces our own `strtod` and `strtoll`-like functions, which are faster than the standard library ones since they (a) don't care for hexadecimal or octal notations, (b) don't support user-specific locale and (c) don't ensure perfect rounding, especially for denormal floats. The last point mean that we can be off from strtod, usually by a **relative** error of 1e-15, and no more than 1e-14 in my tests.

I have a follow-up PR using string_view in `TextFile`, but this one is already pretty big, so let's start with this!


src/formats/SMI.cpp
```diff
@@ -1,6 +1,8 @@
 // Chemfiles, a modern library for chemistry file reading and writing
 // Copyright (C) Guillaume Fraux and contributors -- BSD license
 
+#include <iostream>
```
frodofine: Is this needed?

src/formats/SMI.cpp
```diff
@@ -150,16 +174,16 @@ void SMIFormat::process_property_list_(Topology& topol, const std::string& smile
...
                 // Set the type of the chirality
-                chirality_type += " " + smiles.substr(i + 1, 2);
+                chirality_type += " " + smiles.substr(i + 1, 2).to_string();
                 i += 2;
                 chirality_type += std::to_string(read_number(smiles, ++i));
             }
             new_atom.set("chirality", chirality_type);
             break;
         default:
+            std::cout << smiles << std::endl;
```
frodofine: Should be removed

src/formats/SMI.cpp
```diff
@@ -220,22 +245,24 @@ void SMIFormat::read_next(Frame& frame) {
...
+            if (!is_aromatic_organic(smiles[i])) {
+                throw format_error("SMI Reader: aromatic element not found: '{}'", smiles[i]);
             }
             auto name = smiles.substr(i, 1);
-            add_atom(topol, std::move(name)).set("is_aromatic", true);
+            this->add_atom(topology, name).set("is_aromatic", true);
             continue;
         }
 
+        std::cout << smiles << std::endl;
```
frodofine: Remove

src/formats/SMI.cpp
```diff
@@ -717,7 +743,7 @@ void SMIFormat::write_next(const Frame& frame) {
         }
         auto not_hit = std::find(written.begin(), written.end(), false);
         auto current_atom = static_cast<size_t>(std::distance(written.begin(), not_hit));
-        write_atom(frame, written, current_atom, current_atom);
+        this->write_atom(frame, written, current_atom, current_atom);
```
frodofine: Is the use of `this` recommended for style now?

src/formats/Tinker.cpp
```diff
@@ -157,7 +154,7 @@ std::streampos TinkerFormat::forward() {
             return std::streampos(-1);
         } else {
             // Get the number of atoms in the line
-            natoms = parse<size_t>(split(trim(line), ' ')[0]);
+            natoms = parse<size_t>(split(line, ' ')[0]);
```
frodofine: No longer trimmed, no sure if that's important.

src/formats/XYZ.cpp
```diff
@@ -71,6 +71,8 @@ void XYZFormat::write_next(const Frame& frame) {
     }
 }
 
+#include <iostream>
```
frodofine: Remove this?

src/formats/XYZ.cpp
```diff
@@ -90,10 +92,10 @@ std::streampos XYZFormat::forward() {
 
     try {
         file_->skiplines(natoms + 1);
-    } catch (const FileError& e) {
+    } catch (const FileError&) {
         // We could not read the lines from the file
         throw format_error(
-            "not enough lines for XYZ format: {}", file_->path(), e.what()
+            "not enough lines for XYZ format in '{}'", file_->path()
```
frodofine: Consider rewriting to *XYZ format: not enough lines* to ensure consistency with other error messages.

src/formats/mmCIF.cpp
```diff
@@ -83,41 +83,42 @@ mmCIFFormat::mmCIFFormat(std::string path, File::Mode mode, File::Compression co
         }
 
         if (line_split[0] == "_cell_length_a" || line_split[0] == "_cell.length_a") {
-            a = cif_to_double(line_split[1]);
+            a = cif_to_double(line_split[1].to_string());
```
frodofine: Maybe `cif_to_double` should take a `string_view` argument.

src/formats/PDB.cpp
```diff
@@ -314,7 +319,7 @@ void PDBFormat::read_ATOM(Frame& frame, const std::string& line,
 
     if (atom_offsets_.empty()) {
         try {
-            auto initial_offset = parse<long long>(trim(line.substr(6, 5)));
+            auto initial_offset = parse<long long>(line.substr(6, 5));
```
frodofine: Ok, parse does trim for you!

src/formats/MMTF.cpp
```diff
@@ -24,6 +24,7 @@
 #include "chemfiles/Connectivity.hpp"
 
 #include "chemfiles/types.hpp"
+#include "chemfiles/parse.hpp"
```
frodofine: Is this needed? Maybe for `trim`, but that should be handled by the external library.

src/formats/LAMMPSData.cpp
```diff
@@ -54,177 +56,146 @@ template<> FormatInfo chemfiles::format_information<LAMMPSDataFormat>() {
...
         style_ = TRI;
-    } else if (name == "wavepacket") {
+    } else if (name_ == "wavepacket") {
         style_ = WAVEPACKET;
-    } else if (name == "hybrid") {
+    } else if (name_ == "hybrid") {
         style_ = HYBRID;
     } else {
-        throw format_error("unknown atom style '{}'", name);
+        throw format_error("unknown atom style '{}'", name_);
```
frodofine: Maybe add *LAMMPSData format* to error message

Luthaf:
> but I think so debug code made it into the commit.

Whoops, thanks for spotting it!

src/formats/PDB.cpp
Luthaf: yep!

src/formats/mmCIF.cpp
Luthaf: it can't, since it needs to erase the potential `(` and `)`.

tests/formats/gro.cpp
```diff
@@ -122,17 +122,6 @@ TEST_CASE("Read files in Gromacs .gro format") {
         CHECK(residue.contains(1));
         CHECK(residue.contains(2));
     }
-/* I will finish this in a different PR that may address more errors than these
```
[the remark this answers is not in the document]
Luthaf: Yeah, this should be an issue rather than a comment burried in code. I'll open it tomorrow if you don't beat me!

src/formats/SMI.cpp
Luthaf: I don't know, I was feeling like it =). What do you think?

include/chemfiles/formats/LAMMPSData.hpp
```diff
@@ -46,7 +46,7 @@ class atom_style final {
     mutable bool warned_ = false;
 
 public:
-    atom_style(const std::string& name);
+    explicit atom_style(std::string name);
```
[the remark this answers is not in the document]
Luthaf: Yes, we need to move the style name anyway

tests/formats/gro.cpp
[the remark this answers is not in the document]
frodofine: Go ahead!

src/formats/SMI.cpp
frodofine: Doesn't really matter to me.

PR #273, opened 2019-10-02
fix(trajectory): do not allow reading frames in append mode

### LINKED ISSUE #272: Do not allow reading trajectories in append mode
It is currently [possible to read in append](https://github.com/chemfiles/chemfiles/blob/e13319b23ff86a5f949875dd5d66ccacf609b999/src/Trajectory.cpp#L137) mode.
IMHO this is not good:
- In C appending allows writing only (see [7.21.5.3  The `fopen` function](http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1570.pdf))
- When implementing writing/appending for some formats (I will hopefully have time to tackle XTC+TRR in a while) the underlying library (e.g. xdrfile) does not support reading while appending in most cases. So one would have to rewrite/do ugly workarounds to allow sthis anyways :(

Any thoughts on this?

Could do a PR that forbids this, it doesn't seem to break any tests :D

ezavod:
Fixes #272.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   based format.
 * Added a native implementation of XTC reader, replacing the VMD molfile
   version. The new code should be faster and use less memory.
+* Added a native implementation of TRR reader, replacing the VMD molfile
+  version. The new code should be faster and use less memory.
+* Remove the ability to read frames from a trajectory that was opened in
+  append mode. This mode is now write only.
 
 ## 0.9.0 (18 Nov 2018)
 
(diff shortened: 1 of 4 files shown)


Luthaf:
You will want to also update the documentation in Trajectory.hpp, and perhaps in the doc/ directory too.

ezavod:
Sorry, good point, am not used to having such elaborate comments/documentation. There seem to be no occurrences in doc/

Luthaf:
And this should be a CHANGELOG entry (sorry for not thinking about this earlier ...)

ezavod:
Sure, I also forgot to add the TRR stuff the last time.

PR #274, opened 2019-10-03
feat: xtc writer

### LINKED ISSUE #2: Gromacs XTC files
| File Type | binary |
| --- | --- |
| Topological information | ??? |
| Positions | Yes |
| Velocities | Yes |

Reference http://www.gromacs.org/Developer_Zone/Programming_Guide/XTC_Library

ezavod:
Fixes #2.

~Not sure why valgrind fails, will look into it...~
I think I found them, needs fixing in xdrfile.


src/formats/XTC.cpp
```diff
@@ -53,6 +53,32 @@ void XTCFormat::read(Frame& frame) {
     step_++;
 }
 
+void XTCFormat::write(const Frame& frame) {
+    int natoms = static_cast<int>(frame.size());
+    if (xtc_.nframes() == 0 && step_ == 0) {
+        xtc_.set_natoms(natoms);
+    } else if (natoms != xtc_.natoms()) {
+        throw format_error(
+            "XTC format does not support varying numbers of atoms: Expected {}, but got {}",
```
Luthaf: Expected => expected (majuscule) =)

src/formats/XTC.cpp
```diff
@@ -63,6 +89,16 @@ void XTCFormat::set_positions(const std::vector<float>& x, Frame& frame) {
     }
 }
 
+void XTCFormat::get_positions(std::vector<float>& x, const Frame& frame) {
+    auto positions = frame.positions();
```
Luthaf: Could you add an `assert(x.size() == 3 * positions.size())` here ? Just guarding against possible changes.

src/formats/XTC.cpp
```diff
@@ -85,3 +121,17 @@ void XTCFormat::set_cell(matrix box, Frame& frame) {
     // Factor 10 because the cell lengthes are in nm in the XTC format
     frame.set_cell({a.norm() * 10, b.norm() * 10, c.norm() * 10, alpha, beta, gamma});
 }
+
+void XTCFormat::get_cell(matrix box, const Frame& frame) {
+    // Factor 10 because the cell lengthes are in nm in the XTC format
+    const auto& matrix = frame.cell().matrix() / 10.0;
```
Luthaf: `const auto&` can be replaced by `auto` here: you are creating a new matrix in the division anyway.

tests/formats/xtc.cpp
```diff
@@ -79,3 +80,317 @@ TEST_CASE("Read files in XTC format") {
...
+    auto file = Trajectory(tmpfile, 'w');
+
+    auto frame = Frame();
+    frame.add_atom(Atom("A"), {1, 2, 3});
+    file.write(frame);
+
+    frame = Frame();
+    frame.add_atom(Atom("A"), {1, 2, 3});
+    frame.add_atom(Atom("B"), {4, 5, 6});
+    CHECK_THROWS_AS(file.write(frame), FormatError);
```
Luthaf: Can you replace this by `CHECK_THROWS_WITH` and check that the error message is as expected?

tests/formats/xtc.cpp
```diff
@@ -79,3 +80,317 @@ TEST_CASE("Read files in XTC format") {
...
+    file.close();
+
+    file = Trajectory(tmpfile, 'a');
+    frame = Frame();
+    frame.add_atom(Atom("A"), {1, 2, 3});
+    frame.add_atom(Atom("B"), {4, 5, 6});
+    CHECK_THROWS_AS(file.write(frame), FormatError);
```
Luthaf: Same as above

src/formats/XTC.cpp
ezavod: Done, probably also good in `set_positions`.

src/formats/XTC.cpp
ezavod: Done.

tests/formats/xtc.cpp
ezavod: Didn't know about that :)

tests/formats/xtc.cpp
ezavod: Done.

src/formats/XTC.cpp
Luthaf: Yes !

PR #278, opened 2019-10-29
Extended xyz

### LINKED ISSUE #63: Usefull comment line in XYZ files
The XYZ format has a comment line that currently contains [`Written by the chemfiles library`](https://github.com/chemfiles/chemfiles/blob/606b9f7668fdbff907f945e30e3f48eb45da29ca/src/formats/XYZ.cpp#L105). We should change this to put actual data inside this line: the frame step, the cell parameters, ...

Even if this is not standard, it can be useful for some people, particularly when chemfiles is used as the backend for writing trajectories.

The only downside I can see is that it can be a bit surprising that chemfiles writes some additional data to an XYZ file, but then is not able to read it back.

Luthaf:
Fix #63 

This PR is based on top of #271, only the last commit is relevant.

---

This implement the extended XYZ format, using the XYZ comment line to store additional data. Most notably, this enable storing the unit cell for XYZ files!

Future extensions to this can include storing residue information, or other PDB style data in XYZ files.

The implementation is based on the libatoms documentation of the format, most notably supported by ASE: https://libatoms.github.io/QUIP/io.html#module-ase.io.extxyz. I was unfortunately unable to find an archived version of the original link in #63 (https://camtools.cam.ac.uk/wiki/site/5b59f819-0806-4a4d-0046-bcad6b9ac70f/Extendedxyz.html) to check for difference. It looks like it should be able to handle the version implemented by atomsk too (http://atomsk.univ-lille1.fr/doc/en/format_xyz.html)


Luthaf:
Notes for myself:

-[ ] update CHANGELOG
-[ ] update documentation, showing which formats can and can not read unit cell info

include/chemfiles/formats/XYZ.hpp
```diff
@@ -24,6 +24,11 @@ class XYZFormat final: public TextFormat {
     void read_next(Frame& frame) override;
     void write_next(const Frame& frame) override;
     optional<uint64_t> forward() override;
+
+private:
+    // used to give better error message in `forward`, this refers to the
+    // current step being checked.
+    size_t forward_step_ = 0;
```
ezavod: Would `current_step` be more descriptive?

src/formats/XYZ.cpp
```diff
@@ -78,19 +103,396 @@ optional<uint64_t> XYZFormat::forward() {
...
+            matrix[0][1], matrix[1][1], matrix[2][1],
+            matrix[0][2], matrix[1][2], matrix[2][2]
+        );
+    }
+
+    auto sorted_properties = std::map<string_view, Property>{frame.properties().begin(), frame.properties().end()};
+    // support for generic frame properties
+    for (const auto& it: sorted_properties) {
+        if (contains_space(it.first)) {
+            warning("Extended XYZ", "frame property '{}' contains spaces, skipping", it.first);
```
ezavod: Are we sure, that spaces in keys are explicitly invalid? Couldn't find anything about this in the spec and QUIP seems to [allow](https://github.com/libAtoms/QUIP/blob/10245cffc56cf5d0a3de6e5d3475386d7e4f3c80/src/libAtoms/xyz.c#L701-L742) quoted keys (at least for reading; based on their comments, so I'm not fully sure about that).

src/formats/XYZ.cpp
```diff
@@ -78,19 +103,396 @@ optional<uint64_t> XYZFormat::forward() {
...
+    }
+
+    if (properties.count("Properties") == 1) {
+        return parse_property_list(properties.at("Properties"));
+    } else {
+        return {};
+    }
+}
+
+// this function throws `chemfiles::Error` if the values are not convertible to
```
ezavod: Begin with uppercase `This` because it's a sentence.

src/formats/XYZ.cpp
```diff
@@ -78,19 +103,396 @@ optional<uint64_t> XYZFormat::forward() {
...
+    return result;
+}
+
+/// mapping name=>properties values. The use of string_view is safe since the
+/// original string_view (as returned by file.readline()) will stay alive longer
+/// than this map.
+using extended_xyz_properties_map_t = std::unordered_map<string_view, string_view>;
+
+/// A simple parser for the extended XYZ comment line. This parser bails out as
+/// soons as possible if  the line does not seems to follow extended XYZ
```
ezavod: Two white spaces after "if  " :D

src/formats/XYZ.cpp
```diff
@@ -78,19 +103,396 @@ optional<uint64_t> XYZFormat::forward() {
...
+/// convention.
+///
+/// This line should contains key/values pairs separated by spaces, while the
+/// key and the value are separated by an = sign. If the value contains spaces,
+/// it is then quoted, both single quotes and double quotes are accepted here.
+/// Example: Properties=species:S:1:pos:R:3:bad:I:1 Lattice=" 3 0 0 0 4 0 0 0 5" name='simple file'
```
ezavod: Serves the leading space in `Lattice="_3...` any purpose?

src/formats/XYZ.cpp
```diff
@@ -78,19 +103,396 @@ optional<uint64_t> XYZFormat::forward() {
...
+    extended_xyz_parser& operator=(const extended_xyz_parser&) = delete;
+
+    extended_xyz_properties_map_t parse() {
+        extended_xyz_properties_map_t properties;
+        while (!done()) {
+            auto name = get_property_name();
+            if (!name.empty() && current() == '=') {
+                advance();
+            } else {
+                // this does not look like a property, bail out
```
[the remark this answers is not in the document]
Luthaf: Nice catch! This code was written with the assumption that it can also read files not following the extended XYZ convention.

Trying to continue reading could work, the only issue I can see with this is if a file contains `Properties=` in the comment line but is not intended to be an extended XYZ file

src/formats/XYZ.cpp
Luthaf: I'll try a file with quoted key with QUIP to check this. My reading of the documentation made me think quotes were only allowed in values, but you are right that they are not explicitly forbidden for keys ...

src/formats/XYZ.cpp
Luthaf: no, not really

src/formats/XYZ.cpp
Luthaf: ASE reads quoted keys with spaces fine, so I'll have to update this

src/formats/XYZ.cpp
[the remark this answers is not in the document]
ezavod: Didn't think about that. Maybe its ok to bail out, if the line doesn't start with an key/value pair (AFAIK QUIP always starts with the lattice). So one would only have a problem if a non exyz comment line starts with something that looks like a valid key/value pair.

src/formats/XYZ.cpp
Luthaf: This is more complex than I though, I've asked ASE maintainers about this: https://gitlab.com/ase/ase/issues/474

Luthaf: The implementation should be working now, but we are still missing a standard other than 'whatever ASE does'

src/formats/XYZ.cpp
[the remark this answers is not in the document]
Luthaf: Should be working now, I've added such None-valued property in the test files: https://github.com/chemfiles/tests-data/commit/44c73902de2ac14ca9a083b3da6916bc44aba6dc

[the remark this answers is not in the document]
ezavod: I haven't kept fully track of the discussion on the ase repo, but ase's [newly added documentation](https://gitlab.com/ase/ase/-/blob/master/ase/io/extxyz.py#L702-704) seems to suggest that keys without value should be `true`. I'm currently a little bit lost in the documentations and don't know which one is "right". (So there are no `None` type values? But why handles QUIP such cases as `None` types?)
Chemfiles currently skips keys without value. Should we read such values as `true` instead?

src/formats/XYZ.cpp
ezavod: I guess out best bet is to do [whatever ase does](https://gitlab.com/ase/ase/-/blob/master/ase/io/extxyz.py#L604-713) for now. The current state of this PR is already quite well and might be useful to some users (e.g. for me personally =)). Figuring out a comprehensive standard could be part of #280.

src/formats/XYZ.cpp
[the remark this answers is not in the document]
Luthaf: I'll double check what ASE & QUIP are doing, but if ASE is setting these to True, I am fine with doing the same =)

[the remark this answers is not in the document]
Luthaf: Actually, nowadays QUIP uses ASE directly to read files instead of there own code, so setting properties without values to true should be enough

Luthaf:
Not sure what's the state of this PR. Is it good to go for you @ezavod?

ezavod:
:+1: yes, this looks good to me.
Wasn't sure if there was anything going on in the ase repo that we were waiting for. I currently have not much time to follow the discussion over there (finishing my master and starting a phd position).

Luthaf:
Ok, let's merge this and revisit it later as needed.

> finishing my master and starting a phd position

Great news and good luck for the end of your master! Where are are you doing your PhD, if it is not indiscrete?

ezavod:
>Great news and good luck for the end of your master! Where are are you doing your PhD, if it is not indiscrete?

Thank you, I'm very happy to continue working in the Heuer group at the University of Münster.

PR #281, opened 2019-11-12
Update to a new version of emscripten, supporting 64-bit FS operations

Luthaf:


Luthaf:
This should be straightforward, so I'll merge it in a few days if no-one disagree!

Issue #287, opened 2019-11-28
MMTF writer fails if the topology contains no residues


Luthaf:
For example, converting xyz to MMTF fails with 

```
inconsistent numAtoms size
mmtf EncoderError, StructureData does not have Consistent data... exiting!
```

This seems related to missing residues (see below), but adding 'dummy' residues (one residue per atom, no resid or resname) manually tranforms the above error into a segfault, so there might be more to it.

https://github.com/chemfiles/mmtf-cpp/blob/136834e757aa95e544be68cb8d486ec2d748c169/include/mmtf/structure_data.hpp#L1018-L1021

https://github.com/chemfiles/mmtf-cpp/blob/136834e757aa95e544be68cb8d486ec2d748c169/include/mmtf/structure_data.hpp#L961-L963

PR #289, opened 2019-12-01
Fix MMTF writer for non-biological systems

Luthaf:
The code made a few assumption on what data was present in the frame, assumptions only matched by files containing biomolecular systems.

In particular, atoms are not guaranteed by chemfiles to be ordered inside residues, and not all atoms have associated residues.

@frodofine I would appreciate if you could give this a look! If you don't have time, that's fine, just let me know 😃

### DIFF
--- a/include/chemfiles/formats/MMTF.hpp
+++ b/include/chemfiles/formats/MMTF.hpp
@@ -55,6 +55,9 @@ private:
     /// Number of atoms read before the current model being read.
     /// Used as an offset for adding bonds when reading models.
     size_t atomSkip_ = 0;
+
+    /// did we already send the warning for non-contiguous atoms
+    bool warned_non_contiguous_ = false;
 };
 
 template<> FormatInfo format_information<MMTFFormat>();
--- a/include/chemfiles/formats/MMTF.hpp
+++ b/include/chemfiles/formats/MMTF.hpp
@@ -10,6 +10,7 @@
 
 #include "chemfiles/File.hpp"
 #include "chemfiles/Format.hpp"
+#include "chemfiles/Residue.hpp"
 
 namespace chemfiles {
 class Frame;
@@ -32,6 +33,8 @@ public:
     size_t nsteps() override;
 
 private:
+    // add a single residue to the structure_, using the data from the frame
+    void add_residue_to_structure(const Frame& frame, const Residue& residue);
 
     /// MMTF-CPP object holding all MacroMolecular information.
     /// This can be read and modified.
@@ -54,6 +57,15 @@ private:
 
     /// Number of atoms before the current model.
     size_t atomSkip_ = 0;
+
+    // Since MMTF uses model->chain->residue->atom as storage model, and
+    // chemfiles do not enforce that residues contains contiguous atoms, the
+    // atoms can be re-ordered when adding them to a MMTF structure. This vector
+    // stores the correspondance chemfiles index => MMTF index to be able to add
+    // the right bonds.
+    //
+    // This is only used when writing
+    std::vector<int32_t> new_atom_indexes_;
 };
 
 template<> FormatInfo format_information<MMTFFormat>();
--- a/include/chemfiles/formats/MMTF.hpp
+++ b/include/chemfiles/formats/MMTF.hpp
@@ -52,12 +52,8 @@ private:
     /// Current atom being read. Ranges from [0, structure.numAtoms)
     size_t atomIndex_ = 0;
 
-    /// Number of atoms read before the current model being read.
-    /// Used as an offset for adding bonds when reading models.
+    /// Number of atoms before the current model.
     size_t atomSkip_ = 0;
-
-    /// did we already send the warning for non-contiguous atoms
-    bool warned_non_contiguous_ = false;
 };
 
 template<> FormatInfo format_information<MMTFFormat>();
--- a/include/chemfiles/formats/MMTF.hpp
+++ b/include/chemfiles/formats/MMTF.hpp
@@ -33,8 +33,13 @@ public:
     size_t nsteps() override;
 
 private:
-    // add a single residue to the structure_, using the data from the frame
+    /// add a single residue to the structure_, using the data from the frame
     void add_residue_to_structure(const Frame& frame, const Residue& residue);
+
+    /// A function to translate from the index in MMTF lists to an atom id
+    /// suitable for chemfiles: starts at 0 for each model, and correspond to
+    /// the initial atom index if it exists. This is used when reading.
+    size_t atom_id(size_t mmtf_id);
 
     /// MMTF-CPP object holding all MacroMolecular information.
     /// This can be read and modified.
--- a/src/formats/MMTF.cpp
+++ b/src/formats/MMTF.cpp
@@ -293,7 +293,6 @@ void MMTFFormat::write(const Frame& frame) {
     structure_.numAtoms += static_cast<int32_t>(frame.size());
 
     const auto& topology = frame.topology();
-    const auto& positions = frame.positions();
 
     // pre-allocate some memory
     auto size = static_cast<size_t>(structure_.numAtoms);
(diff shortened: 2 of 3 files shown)


Luthaf:
And fix #287 !

frodofine:
I'll try today!

include/chemfiles/formats/MMTF.hpp
```diff
@@ -52,9 +55,17 @@ class MMTFFormat final: public Format {
...
+    /// Number of atoms before the current model.
     size_t atomSkip_ = 0;
+
+    // Since MMTF uses model->chain->residue->atom as storage model, and
+    // chemfiles do not enforce that residues contains contiguous atoms, the
+    // atoms can be re-ordered when adding them to a MMTF structure. This vector
+    // stores the correspondance chemfiles index => MMTF index to be able to add
+    // the right bonds.
+    //
+    // This is only used when writing
```
frodofine: Ok, good. I've a few performance improvements that I should be posting soon and I need to reading to work as before.

src/formats/MMTF.cpp
```diff
@@ -238,172 +248,259 @@ void MMTFFormat::read(Frame& frame) {
...
+    case Bond::BondOrder::UNKNOWN:
+        return -1;
+    case Bond::BondOrder::QINTUPLET:
+    case Bond::BondOrder::AMIDE:
+    case Bond::BondOrder::AROMATIC:
+    case Bond::BondOrder::UP:
+    case Bond::BondOrder::DOWN:
+    case Bond::BondOrder::DATIVE_L:
+    case Bond::BondOrder::DATIVE_R:
+        warning("MMTF Writer", "bond order '{}' can not be represented in MMTF, defaulting to single bond");
```
frodofine: Is this missing an argument? Possibly `order`?

Also, I see that `default` isn't included here, but maybe it shouldn't be as it is better to explicitly list the other cases.

src/formats/MMTF.cpp
```diff
@@ -111,7 +128,45 @@ void MMTFFormat::read(Frame& frame) {
         frame.set("deposition_date", structure_.depositionDate);
     }
 
+    // A function to translate from the index in MMTF lists to an atom id
+    // suitable for chemfiles: starts at 0 for each model, and correspond to
+    // the initial atom index if it exists.
+    std::function<size_t(size_t)> atom_id;
+    if (!mmtf::isDefaultValue(structure_.atomIdList)) {
+        atom_id = [&](size_t global_id){
```
frodofine: Nitpick: I prefer explicitly listing values captured by a lambda.

src/formats/MMTF.cpp
```diff
@@ -238,172 +248,259 @@ void MMTFFormat::read(Frame& frame) {
...
+            }
             continue;
         }
 
-        size_t atom_idx1 = atom1 - atomSkip_;
-        size_t atom_idx2 = atom2 - atomSkip_;
-
-        frame.add_bond(atom_idx1, atom_idx2);
+        frame.add_bond(atom_id(atom1), atom_id(atom2));
     }
```
frodofine: Note that this is not efficient. I've reworked this code, but never pushed upstream. I'll try to do this shortly.

src/formats/MMTF.cpp
```diff
@@ -238,172 +248,259 @@ void MMTFFormat::read(Frame& frame) {
...
     atomSkip_ = atomIndex_;
 }
 
 void MMTFFormat::write(const Frame& frame) {
-    // Used to add bonds  bonds
-    auto writeAtomLoc = static_cast<size_t>(structure_.numAtoms);
-
     structure_.numModels++;
     structure_.chainsPerModel.emplace_back(0);
+    structure_.numAtoms += static_cast<int32_t>(frame.size());
```
frodofine: Nitpick: Should a check for exceding the possible size of an MMTF file be placed here? I know that it is unlikely to happen, but this is done for the **PDB** and **GRO** formats already.

src/formats/MMTF.cpp
```diff
@@ -238,172 +248,259 @@ void MMTFFormat::read(Frame& frame) {
...
+    new_atom_indexes_.resize(frame.size(), -1);
+
+    std::string previous_chainId;
+    std::string previous_chainName;
+    bool got_at_least_one_chain_in_this_frame = false;
+    for (const auto& residue: topology.residues()) {
+        // WARNING: this assumes that residues in the same chain are contiguous
+        //
+        // [residue[chainid="A"], residue[chainid="B"], residue[chainid="A"]]
+        // will result in three chains
```
frodofine: Technically, this should be three chains anyway (different IDs, but maybe not names).

src/formats/MMTF.cpp
```diff
@@ -238,172 +248,259 @@ void MMTFFormat::read(Frame& frame) {
...
 
-        const auto& current_chainId_opt = res_opt->get("chainid");
-        const auto& current_chainName_opt = res_opt->get("chainname");
-        std::string current_chainId;
-        std::string current_chainName;
-
-        if (current_chainId_opt && current_chainId_opt->kind() == Property::STRING) {
-            current_chainId = current_chainId_opt->as_string();
+    if (!not_in_residue.empty()) {
+        // add all of these new articficial residues in a separate chain
```
frodofine: This is the major portion of this commit, I approve!

tests/formats/mmtf.cpp
```diff
@@ -250,4 +250,67 @@ TEST_CASE("Write files in MMTF format") {
         CHECK(topology.are_linked(topology.residue(0), topology.residue(1)));
         CHECK(!topology.are_linked(topology.residue(0), topology.residue(2)));
     }
+
+    SECTION("Structure not from RCSB") {
+        auto tmpfile = NamedTempPath(".mmtf");
```
frodofine: It would be interesting to see what an application like PyMol does with this file (it uses the same MMTF backend). My guess is that it wouldn't open, but I don't have the time to check.

src/formats/MMTF.cpp
Luthaf: Is the inefficiency due to the fact that all bonds are iterated over, instead of just the ones related to the current model?

src/formats/MMTF.cpp
Luthaf: You mean exceeding `int32_t`? Or is there another size limit for MMTF?

tests/formats/mmtf.cpp
Luthaf: I converted the water.xyz test file to mmtf with this branch, and PyMol seems to load it fine.

src/formats/MMTF.cpp
Luthaf: Whoop, you are right.

src/formats/MMTF.cpp
Luthaf: I can not since the lambda is capturing `this`. But I moved it to a separate private member function.

src/formats/MMTF.cpp
Luthaf: I am not sure I understand. How are chains determined? I though that residues with the same chainid and chainname should be part of the same chain.

Luthaf:
> I've a few performance improvements that I should be posting soon and I need to reading to work as before.

This is nice to hear, I'll be waiting for them! I want to do a 0.9.2 release very soon, do you want your changes to be in or can they wait for the next one?

src/formats/MMTF.cpp
frodofine: True, I was reading fast.

src/formats/MMTF.cpp
frodofine: No, it has to with adding peptide bonds in an odd order. Since the bonds are stored in a `sorted_array`, this causes a lot of moving around memory.

src/formats/MMTF.cpp
frodofine: Exceding `int32_t`. Probably not needed.

tests/formats/mmtf.cpp
frodofine: Great!

src/formats/MMTF.cpp
frodofine: Chains with the same chainid and chainname should be part of the same chain, but in some PDB files the chain names alternate. This causes the chainids to differ from the chainname.

frodofine:
They can wait. I'm swamped until the 15th.

src/formats/MMTF.cpp
Luthaf: Oh, OK. Out of curiosity, what are the gains you have? And if you have an example file where this is an issue, I am also interested. We might be able to change the way we do the sorting in `Connectivity` to help too.

Luthaf:
Ok, thanks for the review, and good luck until the 15th!

PR #293, opened 2019-12-12
Use codecs.open to read files with python

Luthaf:
This is the only way to specify the encoding parmeter with Python2.

The encoding parameter is needed else Python uses the locale to guess the encoding. For LANG=C, this means using ascii encoding and thus having "UnicodeDecodeError: 'ascii' codec can't decode byte ..." exceptions.

### DIFF
--- a/tests/lints/check-capi-docs.py
+++ b/tests/lints/check-capi-docs.py
@@ -4,8 +4,10 @@
 A small script checking that all the C API functions are documented, and have
 an example.
 """
+from __future__ import print_function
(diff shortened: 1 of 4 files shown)


Luthaf:
Unfortunately, we will have to keep Python 2 support for a bit more time, since it is still the default on CentOS 7 (used on a lot of super-computing facilities), which will reach EOL in 2024 ...

Fortunately, keeping Python 2 compatibility is not too hard!

PR #295, opened 2019-12-13
Improve speed of MMTF parsing for global bonds

frodofine:
Before:

```
$ time ./mmtf
[chemfiles] MMTF Writer: atom name 'C23456' is too long for MMTF format, it will be truncated
[chemfiles] MMTF Writer: atom type 'HuuuuH' is too long for MMTF format, it will be truncated
===============================================================================
All tests passed (117 assertions in 2 test cases)

./mmtf  785.46s user 0.25s system 99% cpu 13:05.78 total
```

After:
 ```
$ time ./mmtf
[chemfiles] MMTF Writer: atom name 'C23456' is too long for MMTF format, it will be truncated
[chemfiles] MMTF Writer: atom type 'HuuuuH' is too long for MMTF format, it will be truncated
===============================================================================
All tests passed (117 assertions in 2 test cases)

./mmtf  1.79s user 0.28s system 99% cpu 2.067 total
```

### DIFF
--- a/src/formats/MMTF.cpp
+++ b/src/formats/MMTF.cpp
@@ -128,6 +128,21 @@ void MMTFFormat::read(Frame& frame) {
         frame.set("deposition_date", structure_.depositionDate);
     }
 
+    auto inter_residue_bond_count = structure_.bondAtomList.size() / 2;
+    size_t bond_index = 0;
+    while (bond_index < inter_residue_bond_count) {
+        auto atom1 = structure_.bondAtomList[bond_index * 2 + 0];
+        auto atom2 = structure_.bondAtomList[bond_index * 2 + 1];
+
+        // We are below the atoms we care about
+        if ((atom1 < atomSkip_) || (atom2 < atomSkip_)) {
+            bond_index++;
+            continue;
+        }
+
+        break;
+    }
+
     // count the number of atoms in this frame/model
     size_t natoms = 0;
     auto modelChainCount = static_cast<size_t>(structure_.chainsPerModel[modelIndex_]);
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by frodofine** [comment]
This is ready for review.

File: src/formats/MMTF.cpp [hunk 1]
```diff
@@ -217,6 +232,28 @@ void MMTFFormat::read(Frame& frame) {
...
+                if (atom1 < atomSkip_ || atom2 < atomSkip_) {
+                    if (!(atom1 < atomSkip_ && atom2 < atomSkip_)) {
+                        warning("MMTF Reader", "chemfiles can not represent bonds between different models");
+                    }
+                    continue;
+                }
+
+                // We are above the atoms we care about
+                if (atom1 > atomIndex_ || atom2 > atomIndex_) {
+                    break;
```
Luthaf: Why did you keep the warning above and removed it here?

Do you know if MMTF allow for bonds between models? (Or if that even make sense?)

tests/formats/mmtf.cpp
```diff
@@ -189,6 +189,22 @@ TEST_CASE("Read files in MMTF format") {
         CHECK(!topology.are_linked(topology.residue(0), topology.residue(2)));
     }
 
+    SECTION("Large MMTF file") {
+    // Test fails on Windows due to timing of MSVC debug builds
+    #ifndef CHEMFILES_WINDOWS
```
Luthaf: It would make sense to have benchmarks separated from tests. This can stay as is now, I'll check how we can have benchmarks when I find the time

Luthaf: Looks like Catch (which we are using for tests) supports benchmarks too: https://github.com/catchorg/Catch2/blob/master/docs/benchmarks.md

frodofine: Yeah, that's probably a good idea. We should be benchmarking the release builds, not the debug ones.

src/formats/MMTF.cpp
frodofine: Technically speaking MMTF could store that, but it would be highly unlikely and probably due to an encoding mistake, not biology. I removed the warning in any case because it doesn't make sense. I think most MMTF encoders don't allow for it anyway.

frodofine:
The speed up is due to optimizing the order bonds are inserted. Previously, bonds we're inserted for local residues, then global bonds. Since many of the global bonds are indexed lower than the residue bonds, inserting later causes the slow down. This is because the sorted array needs to move all the later local bonds, then insert the new global bond. This is fine if you are moving only a few bonds, but when you are moving 1000000s of bonds it causes a significant problems. The scaling is probably O(n!), where n is the number of global bonds.

frodofine:
The speed up is due to optimizing the order bonds are inserted. Previously, bonds we're inserted for local residues, then global bonds. Since many of the global bonds are indexed lower than the residue bonds, inserting later causes the slow down. This is because the sorted array needs to move all the later local bonds, then insert the new global bond. This is fine if you are moving only a few bonds, but when you are moving 1000000s of bonds it causes a significant problems. The scaling is probably O(n!), where n is the number of global bonds. If this looks good, I will merge.

PR #297, opened 2020-01-09
Fix PDB reader not reading atomic names when ATOM record are too short

Luthaf:


tests/formats/pdb.cpp
```diff
@@ -246,51 +287,34 @@ TEST_CASE("Read files in PDB format") {
...
-        CHECK(topology.residue_for_atom(1)->get("insertion_code")->as_string() == "c");
-        CHECK(topology.residue_for_atom(2)->get("insertion_code")->as_string() == "x");
-        CHECK_FALSE(frame[3].get("insertion_code"));
+        CHECK(approx_eq(frame.positions()[0], {0.417, 8.303, 11.737}, 1e-5));
+        CHECK(approx_eq(frame.positions()[5], {8.922, 9.426, 5.320}, 1e-5));
     }
 }
 
-TEST_CASE("Problematic PDB files") {
-    auto file = Trajectory("data/pdb/bad/atomid.pdb");
```
frodofine: Is this test no longer valid?

tests/formats/pdb.cpp
```diff
@@ -246,51 +287,34 @@ TEST_CASE("Read files in PDB format") {
...
+    SECTION("Short CRYST1 record") {
+        auto file = Trajectory("data/pdb/short-cryst1.pdb");
+        auto frame = file.read();
     }
 
-    SECTION("Read Nucleic Residues") {
-        auto frame = Trajectory("data/pdb/2hkb.pdb").read();
-        auto topology = frame.topology();
+    SECTION("Short ATOM record") {
+        auto file = Trajectory("data/pdb/short-atom.pdb");
```
frodofine: Does this supersede the below test?

frodofine:
It looks Travis is having issues, but the build is fine otherwise.

tests/formats/pdb.cpp
Luthaf: The file moved to `data/pdb/atom-id-0.pdb`, and is still tested

tests/formats/pdb.cpp
Luthaf: No, this is a new test for the bug fixed by this PR.

Luthaf:
> It looks Travis is having issues, but the build is fine otherwise. 

Seems spurious, I restarted the build and everything is fine.

PR #298, opened 2020-01-10
Improve XYZ format output

Luthaf:
Using `{:g}` instead of `{}` produces nicer output:

```
# with {:g}
0.4569 1.34 4
# with {}
0.45689999999999999973423 1.34 4.0
```


tests/formats/gro.cpp
```diff
@@ -143,7 +143,7 @@ TEST_CASE("Write files in GRO format") {
     "    6XXXXX    E    5   0.400   0.500   0.600  0.9000  1.0000  1.1000\n"
     "    7XXXXX    F    6   0.400   0.500   0.600  0.9000  1.0000  1.1000\n"
     "    8XXXXX    G    7   0.400   0.500   0.600  0.9000  1.0000  1.1000\n"
-    "   2.20000   1.90526   4.40000 0.0 0.0  -1.10000 0.0  0.000000  0.000000\n";
+    "   2.20000   1.90526   4.40000 0.0 0.0  -1.10000 0.0   0.00000   0.00000\n";
```
frodofine: I'm guessing this is due to the new version of fmt? If so, I'm surprised only this changed.

Luthaf: Yes, I was surprised too. I guess this comes from beta & gamma being computed as `arccos`, and having values of 1e-16 (not zero), which get formatted slightly differently with the new version.

PR #299, opened 2020-01-10
Update netcdf to v4.7.3

Luthaf:
While trying to apply [this patch](https://github.com/JuliaPackaging/Yggdrasil/pull/366/files/dee35d1e93021bb4966665cd90b4511571fa1451..71f2d68e1ccfcd4316df3117037580f550dc6b57#diff-d492fa4a88a52553c127373c22f25bb3) to our vendored netcdf, I realized that https://github.com/chemfiles/netcdf-c and `external/netcdf.tar.gz` had drifted apart, and I could not find the changes I made to create `external/netcdf.tar.gz` locally.

So I re-made these changes in https://github.com/chemfiles/netcdf, and this is the corresponding tarball. At the same time it looks like the freebsd patch is no longer needed, which I'll double check before merging this.


Luthaf:
> which I'll double check before merging this.

Yep, it cross-compiles fine to freebsd

Luthaf:
Looks like an additional patch was needed on windows: https://github.com/chemfiles/netcdf/commit/c139bf51ecab580eabafac5d796a09125d7433c7

PR #302, opened 2020-01-24
Fix build with CMake 2.8.12 and add it to CI

Luthaf:


.travis.yml
```diff
@@ -54,10 +54,15 @@ matrix:
...
+    - name: cmake-2.8.12
+      os: linux
+      compiler: gcc
+      env: ALTERNATIVE_CMAKE=2.8.12-Linux-i386
+      addons: *linux64
     - name: 32-bits
       os: linux
       compiler: gcc
-      env: ARCH=x86 CMAKE_EXTRA=-DHAVE_LIBM=/lib32/libm.so.6
+      env: ARCH=x86 CMAKE_EXTRA=-DHAVE_LIBM=/lib32/libm.so.6 ALTERNATIVE_CMAKE=2.8.12-Linux-i386
```
frodofine: Is there a reason that the 32-bit build is only done with the old CMake?

scripts/ci/setup-travis.sh
```diff
@@ -39,18 +39,20 @@ if [[ "$EMSCRIPTEN" == "ON" ]]; then
...
 
     return
 fi
 
+# Install a different version of cmake
+if [[ "${ALTERNATIVE_CMAKE}" != "" ]]; then
+    cd $HOME
+    wget https://cmake.org/files/v${ALTERNATIVE_CMAKE%.*}/cmake-${ALTERNATIVE_CMAKE}.tar.gz
+    tar xf cmake-${ALTERNATIVE_CMAKE}.tar.gz
+    export PATH=$HOME/cmake-${ALTERNATIVE_CMAKE}/bin:$PATH
```
frodofine: It looks like the build is having difficulties finding this version of CMake in the 64-bit build.

.travis.yml
Luthaf: No, just tried to use cmake-2.8 with an existing build. Since it was failing, I moved it to its own build, but now both are failing ...

Luthaf: OK, I got confused here. The 32-bit builder should not have used cmake 2.8.12

scripts/ci/setup-travis.sh
Luthaf: The issue was with a missing 32-bit ld.so (since the cmake binary is 32-bit)

Luthaf:
> unable to find Boost

.travis.yml
```diff
@@ -37,7 +37,7 @@ matrix:
   include:
     - name: emscripten
       os: linux
-      env: STATIC_LIBS=ON EMSCRIPTEN=ON
+      env: STATIC_LIBS=ON EMSCRIPTEN=ON ALTERNATIVE_CMAKE=3.9.3-Linux-x86_64
```
frodofine: I assume this is because you can now remove the `Install modern CMake` section?

Luthaf: Yes, both emscripten and cmake-2.8.12 builders need a different cmake version, so they both use the same mechanism

PR #305, opened 2020-01-28
Prevent inclusion of .git directories in external/<XXX>.tar.gz

Luthaf:
Such directories can get very large, and in turn increase greatly the size of the chemfiles repository.


Luthaf:
This removes around 1Mb in the xdrfile archive.

Issue #303, opened 2020-01-24
A few errors when reading SMILES


Luthaf:
Using the database at http://www.lipidmaps.org/data/structure/download.php, and extracting the associated SMILES strings, I found a few bugs in out SMILES parser

- [X] ~crash of the Python interpreter with exception `'fmt::v5::format_error':  invalid format string`~ #308 

- [x] These SMILES now fail with `SMI Reader: unknown symbol: '{'.`
```
C([C@H](C)CO)C{-}/C=C(\C{+n}C/C=C(/CC/C=C(\C)/CC/C=C(\C)/C)\C)/C
C([C@H](C)C(=O)O)C{-}/C=C(\C{+n}C/C=C(/CC/C=C(\C)/CC/C=C(\C)/C)\C)/C
C([C@H](C)CC(O)=O)C{-}/C=C(\C{+n}C/C=C(/CC/C=C(\C)/CC/C=C(\C)/C)\C)/C
```

---

- [x] error while trying to generate a new Bond (`can not have a bond between an atom and itself`)

```
C1(O)=CC2O[C@H](C3C=CC(O)=C(O)C=3)[C@H](O)[C@@H](C3=C(O)C=C(O)C4[C@H](C5=C(O)C=C(O)C6[C@H](C7=C(O)C=C(O)C8[C@H](C9=C(O)C=C(O)C%10[C@H](C%11=C(O)C=C(O)C%12C[C@@H](O)[C@@H](C%13C=CC(O)=C(O)C=%13)OC=%12%11)[C@@H](O)[C@@H](C%11C=CC(O)=C(O)C=%11)OC=%109)[C@@H](O)[C@@H](C9C=CC(O)=C(O)C=9)OC=87)[C@@H](O)[C@@H](C7C=CC(O)=C(O)C=7)OC=65)[C@@H](O)[C@@H](C5C=CC(O)=C(O)C=5)OC=43)C=2C(O)=C1
```

---

- [x] Different number of atoms between the SMILES structure and the SDF structure. This might be related to implicit hydrogens, I am not sure.

```
C[C@@]12CC(=O)[C@@H]3[C@H](CC[C@H]4C[C@@H](O)CC[C@]43C)[C@H]1CC[C@@]2(O)[C@H](O)CO
structure 30634: different number of atoms: smi 26 / sdf 30

[...]

PR #308, opened 2020-01-29
Small fixes related to exceptions at the C APi boundary

Luthaf:
Related to #303. The first examples now properly fails with `SMI Reader: unknown symbol: '{'`.


Luthaf:
Thanks!

PR #312, opened 2020-02-07
Fix Issues #303 and #255

frodofine:
This PR will address issues #303 and #255. The first issue is described by the inability of the SMILES parser to read CurlySMILES as well as an issue reading bonds. The second issue is addressed by a single line. For #255, the SMILES writer has been reworked to longer be recursive and new test cases have been added to ensure that it functions properly.

### DIFF
--- a/src/formats/SMI.cpp
+++ b/src/formats/SMI.cpp
@@ -383,53 +383,6 @@ void SMIFormat::read_next(Frame& frame) {
     }
 }
 
-/// Depth first search to find all rings.
-/// Not guaranteed to fast, nor efficient, but just to work
-/// Also not guaranteed to find the SSSR, but it will find all rings
-/// needed for a given structure to be writen as SMILES
-/// \param adj_list The adjacentcy list graph
-/// \param hit_atoms Atoms already encountered
-/// \param ring_bonds Bonds already known to ring forming bonds
-/// \param ring_atoms Map from an atom to the number of rings it forms
-/// \param current_atom Atom to be checked in this iteration
-/// \param previous_atom Atom which was bound to the current atom
-static void find_rings_helper(
-    const std::vector<std::vector<size_t>>& adj_list,
-    std::vector<bool>& hit_atoms, std::set<Bond>& ring_bonds,
-    std::unordered_map<size_t, size_t>& ring_atoms,
-    size_t current_atom, size_t previous_atom) {
-
-    auto& current_atom_bonds = adj_list[current_atom];
-    hit_atoms[current_atom] = true;
-
-    for (auto neighbor : current_atom_bonds) {
-        // prevent back tracking
-        if (neighbor == previous_atom) {
-            continue;
-        }
-
-        // We've seen this neighbor before! Ring found
-        if (hit_atoms[neighbor]) {
-            // Don't process the same ring connection twice
-            if (ring_bonds.count(Bond(neighbor, current_atom))) {
-                continue;
-            }
-            ring_bonds.insert(Bond(neighbor, current_atom));
-
-            auto neigh_ring_iter = ring_atoms.find(neighbor);
-            if (neigh_ring_iter == ring_atoms.end()) {
-                neigh_ring_iter = ring_atoms.insert({neighbor, 0}).first;
-            }
-            neigh_ring_iter->second++;
-
-            continue; // No need to see this atom again
-        }
-
-        // Continue the search
-        find_rings_helper(adj_list, hit_atoms, ring_bonds, ring_atoms, neighbor, current_atom);
-    }
-}
-
 static void find_rings(
     const std::vector<std::vector<size_t>>& adj_list,
     std::unordered_map<size_t, size_t>& ring_atoms) {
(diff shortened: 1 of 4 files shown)


src/formats/SMI.cpp
```diff
@@ -333,6 +347,7 @@ void SMIFormat::read_next(Frame& frame) {
                 throw format_error("SMI Reader: rings defined with '%' must be double digits");
             }
             check_ring(parse<size_t>(smiles.substr(i + 1, 2)));
+            i += 2; // this line alone fixes most of issue 303 :)
```
Luthaf: :smile: what was the problem?

src/formats/SMI.cpp
```diff
@@ -441,13 +394,88 @@ static void find_rings(
...
+        // ring_atoms Map from an atom to the number of rings it forms
+
+        // The atoms to process consist of :
+        // current_atom Atom to be checked in this iteration
+        // previous_atom Atom which was bound to the current atom
+
+        std::stack<std::pair<size_t, size_t>> atoms_to_process;
+        atoms_to_process.push({ current_atom, current_atom });
+
+        while (atoms_to_process.size()) {
```
Luthaf: I would be explicit with `!atoms_to_process.empty()`

src/formats/SMI.cpp
```diff
@@ -441,13 +394,88 @@ static void find_rings(
...
+
+            size_t previous_atom;
+
+            std::tie(current_atom, previous_atom) = atoms_to_process.top();
+            atoms_to_process.pop();
+
+            auto& current_atom_bonds = adj_list[current_atom];
+
+            // We go through the neighbors backwards because we are using a
+            // stack instead and want to through them in the forward direction
```
Luthaf: I am not sure I understand this. Is a word missing after **instead**?

src/formats/SMI.cpp
```diff
@@ -441,13 +394,88 @@ static void find_rings(
...
+                    continue;
+                }
+
+                // We've seen this neighbor before! Ring found
+                // but only if we have NOT processed the current atom as this
+                // prevents rings from being created from mutliple directions.
+                if (hit_atoms[neighbor] && !hit_atoms[current_atom]) {
+
+                    // Don't process the same ring connection twice
+                    if (ring_bonds.count(Bond(neighbor, current_atom))) {
```
Luthaf: `ring_bonds.count(Bond(neighbor, current_atom)) != 0` or `ring_bonds.find(Bond(neighbor, current_atom)) != ring_bonds.end()` would be more explicit. (it is unfortunate that we have to wait for C++20 to get `std::map::contains`)

src/formats/SMI.cpp
```diff
@@ -441,13 +394,88 @@ static void find_rings(
...
+                        neigh_ring_iter = ring_atoms.insert({ neighbor, 0 }).first;
+                    }
+                    neigh_ring_iter->second++;
+
+                    continue; // No need to see this atom again
+                }
+
+                // Continue the search for new atoms
+                if (!hit_atoms[neighbor]) {
+                    //find_rings_helper(adj_list, hit_atoms, ring_bonds, ring_atoms, neighbor, current_atom);
```
Luthaf: left-over code?

src/formats/SMI.cpp
```diff
@@ -746,8 +686,120 @@ void SMIFormat::write_next(const Frame& frame) {
...
+        auto start_atom = static_cast<size_t>(std::distance(written.begin(), not_hit));
+
+        // We have found an atom that has not yet been printed! Now we must print it out
+        // along with all of its connections (the entire component).
+
+        // A structure to store a depth first like search through the component.
+        std::stack<std::tuple<size_t, size_t, bool>> atoms_to_process;
+        atoms_to_process.push(std::tuple<size_t,size_t,bool>( start_atom, start_atom, false ));
+
+        while (atoms_to_process.size()) {
```
Luthaf: `!atoms_to_process.empty()`

src/formats/SMI.cpp
```diff
@@ -746,8 +686,120 @@ void SMIFormat::write_next(const Frame& frame) {
...
+                if (written[neighbor]) {
+                    continue;
+                }
+
+                // To print a start bracket, we need to be branching (> 2 non-ring bonds)
+                // and we don't want to brank the last neighbor printed
+                auto needs_to_branch = neighbors_printed != 0 && neighbors_printed > ring_start;
+
+                // Depth First Search like recursion
+                //this->write_atom(frame, hit_atoms, neighbor, current_atom);
```
Luthaf: leftover?

Luthaf:
Looks very good overall, I have a few improvement suggestions and some questions regarding what is going on!

src/formats/SMI.cpp
frodofine: The variable `i` wasn't getting updated properly. So, on the next character, it would read the following digit character as its own ring. So, `%11` would create a ring id of `11` and then create a ring id `i` on the next character, then close that ring on the next character - causing the first error. A similar explanation caused the additional bonds with `%10` through ring id `0`.

Luthaf: Thanks!

src/formats/SMI.cpp
frodofine: It's meant to refer to later in the paragraph. Removing the word actually just makes it easier to read

src/formats/SMI.cpp
frodofine: Yeah, even **C++14** would be nice!

src/formats/SMI.cpp
frodofine: Yes, it is. When rewriting the function I wanted to keep track of where the recursion was.

src/formats/SMI.cpp
Luthaf: Yep =/. I think is it worth supporting old-ish CentOS & Debian releases (CentOS 7 & Debian 8), meaning we need to compile the code with gcc 4.8.5/4.9 respectively. Debian reach EOL in june of this year, but CentOS is still supported until 2024, and computing centers are usually very slow to update.

Do you have anything specific from C++14 you would like to be able to use? We can try to find a C++11 compatible implementation, like with span/string_view/optional.

frodofine: Yeah, I totally understand. I needed to support CentOS 5 for the longest time for an old project, it didn't even support C++11! I don't know if Travis supports this as it is based on Ubuntu, but some CIs support the redhat devtoolset package which allows you to build with newer versions of GCC on old versions of CentOS (this is how I can build **Lemon** on older versions of CentOS). Problem is, many computing centers do not have this package anyway, making it a moot point.

Ideally `std::enable_if_t`, `make_tuple`, and `make_unique`.

frodofine:
Looks like Travis is having issues today (apt couldn't find python, something about MSBuild). I had to restart a few builds - just a warning!

frodofine:
I'll go ahead an merged this in a couple of hours if there's no complaints!

src/formats/SMI.cpp
Luthaf: I think we can add all of these in a `utils` header with a C++11 implementation. Do you want to open an issue about this?

PR #314, opened 2020-02-18
Speed up atom removal in large structures

Luthaf:
Instead of checking each residue to find if the removed atom was inside, directly remove the atom from the corresponding residue


Luthaf:
thanks for the review!

PR #315, opened 2020-02-18
Tweak criterion for bonds detection

### LINKED ISSUE #301: Guess_bonds create bivalent hydrogen
Using the following trajectory [pos.xyz](https://github.com/chemfiles/chemfiles/files/4054682/pos.xyz.txt), `Frame::guess_bonds` find two bonds involving the the hydrogen n°77: one to oxygen n°84 and one to oxygen n°89.

This is somehow surprising to users who expect "standard valence" bonds. We already clean extraneous H-H bonds https://github.com/chemfiles/chemfiles/blob/6b06c7583ba748dc51ccfb3581b03b2883678fc1/src/Frame.cpp#L88-L99

So we could extend the above check to only keep one bond per H atoms, the shortest one.

VMD (which uses a similar bonds detection heuristic) only shows one bond, so we may want to check what they do here

Luthaf:
Fix #301

### DIFF
--- a/src/Frame.cpp
+++ b/src/Frame.cpp
@@ -79,7 +79,9 @@ void Frame::guess_bonds() {
             }
             auto d = distance(i, j);
             auto radii = i_radius.value() + j_radius.value();
-            if (0.03 < d && d < 0.6 * radii && d < cutoff) {
+            // This criteria comes from VMD, updated with the -0.4 since we
+            // don't have the same VdW radii (see issue #301)
+            if (0.03 < d && d < 0.6 * (radii - 0.4) && d < cutoff) {
                 topology_.add_bond(i, j);
             }
         }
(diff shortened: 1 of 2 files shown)


src/Frame.cpp
```diff
@@ -79,7 +79,9 @@ void Frame::guess_bonds() {
             }
             auto d = distance(i, j);
             auto radii = i_radius.value() + j_radius.value();
-            if (0.03 < d && d < 0.6 * radii && d < cutoff) {
+            // This criteria comes from VMD, updated with the -0.4 since we
+            // don't have the same VdW radii (see issue #301)
+            if (0.03 < d && d < 0.6 * (radii - 0.4) && d < cutoff) {
```
frodofine: There's something off here. Some builds pass all tests, others fail. Maybe the `0.4` magic number needs to be tweaked? For the ones that fail, a bond is missing/no bonds are found.

Luthaf: The builder that pass the tests run with `CHFL_BUILD_DOCTESTS=OFF` since they are already very slow. The failing tests are doctests/examples that only run on the remaining ones. 

Since the doc tests use artificial positions, I would rather change them and keep the 0.4 here. See https://github.com/chemfiles/chemfiles/issues/301#issuecomment-574100048 for the rational.

Luthaf: Update: looks like the Cl-Cl bond (which is the one being tested here) is actually 1.98 A. So this fix will not be enough here. 

I'll check if we can use an algorithm based on covalent radii instead of VdW radii for bond detection, since it would make more sense.

Luthaf:
This is somehow harder than I expected ^^. I started digging on publications on the subject, I'll create a new PR with algorithm change later. If you have suggestions of related papers, please comment on the related issue!

PR #316, opened 2020-02-24
Fix a segfault in SMI reader

Luthaf:
While I was at, I also changed the few remaining tests using boost::fs to loop over 'bad' files, and found one case of broken error message.

### DIFF
--- a/src/formats/SMI.cpp
+++ b/src/formats/SMI.cpp
@@ -345,6 +345,9 @@ void SMIFormat::read_next(Frame& frame) {
             branch_point_.push(previous_atom_);
             break;
         case ')':
+            if (branch_point_.empty()) {
+                throw format_error("SMI Reader: unmatched ')'");
+            }
             previous_atom_ = branch_point_.top();
             branch_point_.pop();
             break;
--- a/src/formats/SMI.cpp
+++ b/src/formats/SMI.cpp
@@ -379,7 +379,7 @@ void SMIFormat::read_next(Frame& frame) {
     }
 
     if (!rings_ids_.empty()) {
-        throw format_error("SMI Reader: {} unclosed ringid '{}'", rings_ids_.begin()->first);
+        throw format_error("SMI Reader: unclosed ringid '{}'", rings_ids_.begin()->first);
     }
 
     frame.resize(topology.size());
--- a/tests/formats/sdf.cpp
+++ b/tests/formats/sdf.cpp
@@ -99,14 +99,20 @@ TEST_CASE("Errors in SDF format") {
     auto file = Trajectory("data/sdf/bad/bad_atom_line.sdf");
     CHECK_THROWS_WITH(file.read(), "atom line is too small for SDF: '    3.7320   -0.0600'");
 
-    file = Trajectory("data/sdf/bad/bad_counts_line.sdf");
-    CHECK_THROWS_WITH(file.read(), "can not read file 'data/sdf/bad/bad_counts_line.sdf' at step 0, it does not contain any step");
+    CHECK_THROWS_WITH(
+        Trajectory("data/sdf/bad/bad_counts_line.sdf"),
+        "could not parse counts line: ' 21aaa           '"
+    );
 
-    file = Trajectory("data/sdf/bad/bad_counts_line2.sdf");
-    CHECK_THROWS_WITH(file.read(), "can not read file 'data/sdf/bad/bad_counts_line2.sdf' at step 0, it does not contain any step");
+    CHECK_THROWS_WITH(
+        Trajectory("data/sdf/bad/bad_counts_line2.sdf"),
+        "could not parse counts line: '  0  0'"
+    );
 
-    file = Trajectory("data/sdf/bad/blank.sdf");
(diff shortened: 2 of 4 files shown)


tests/formats/sdf.cpp
```diff
@@ -98,27 +95,18 @@ TEST_CASE("Read files in SDF format") {
...
-        auto test = [=](){
-            // We can throw either when creating the trajectory, or when reading
-            // the frame, depending on the type of error
-            auto file = Trajectory(entry.path().string());
-            file.read();
-        };
-        CHECK_THROWS(test());
-    }
+    file = Trajectory("data/sdf/bad/blank.sdf");
+    CHECK_THROWS_WITH(file.read(), "can not read file 'data/sdf/bad/blank.sdf' at step 0, it does not contain any step");
```
Luthaf: @frodofine the three tests fails with the same error, I guess they are checking the `forward` implementation/step number counter? Or is there a way to get a more specific error message for them?

frodofine: You should be able to get different error messages, I'll take a look if I get a chance today.

frodofine: You should be able to get specific error messages. I'll take a look if I get the chance.

frodofine: SDF's forward should be the following:

```C++

optional<uint64_t> SDFFormat::forward() {
    auto position = file_.tellpos();
    size_t natoms = 0;
    size_t nbonds = 0;
    auto counts_line = string_view("");
    try {
        // Ignore junk lines
        for (size_t i=0; i<3; i++) {
            file_.readline();
        }

        if (file_.eof()) {
            return nullopt;
        }

        counts_line = file_.readline();
        if (counts_line.length() < 10) {
            throw format_error("counts line must have at least 10 digits, it has {}", counts_line.length());
        }

        natoms = parse<size_t>(counts_line.substr(0,3));
        nbonds = parse<size_t>(counts_line.substr(3,3));
    } catch (const Error&) {
        // We could not read an integer, so give up here
        throw format_error("could not parse counts line: '{}'", counts_line);
    }

    try {
        for (size_t i=0; i<(natoms + nbonds); i++) {
            file_.readline();
        }
    } catch (const FileError&) {
        // We could not read the lines from the file
        throw format_error(
            "not enough lines in '{}' for SDF format", file_.path()
        );
    }

    // Search for ending character, updating the position in the file only
    while (!file_.eof()) {
        if (file_.readline() == "$$$$") {
            break;
        }
    }

    // We have enough data to parse an entire molecule.
    // So, even if the file may not have an ending string,
    // return the start of this step
    return position;
}
```

and the test should be:

```C++
TEST_CASE("Errors in SDF format") {
    auto file = Trajectory("data/sdf/bad/bad_atom_line.sdf");
    CHECK_THROWS_WITH(file.read(), "atom line is too small for SDF: '    3.7320   -0.0600'");

    CHECK_THROWS_WITH(
        Trajectory("data/sdf/bad/bad_counts_line.sdf"),
        "could not parse counts line: ' 21aaa           '"
    );

    CHECK_THROWS_WITH(
        Trajectory("data/sdf/bad/bad_counts_line2.sdf"),
        "could not parse counts line: '  0  0'"
    );

    CHECK_THROWS_WITH(
        Trajectory("data/sdf/bad/blank.sdf"),
        "could not parse counts line: 'asdf'"
    );
}
```

I tried to update the pull with the new code, but I couldn't figure out how.

Luthaf: > I tried to update the pull with the new code, but I couldn't figure out how.

You should b able to add https://github.com/luthaf/chemfiles/ as a remote, checkout to the right branch (smi-segfault) commit there and push directly.

I'll put that code in tomorrow if you don't beat me to it!

src/formats/SMI.cpp
```diff
@@ -345,6 +345,9 @@ void SMIFormat::read_next(Frame& frame) {
             branch_point_.push(previous_atom_);
             break;
         case ')':
+            if (branch_point_.empty()) {
+                throw format_error("SMI Reader: unmatched ')'");
+            }
```
frodofine: Good fix!

tests/formats/smi.cpp
```diff
@@ -255,27 +251,28 @@ TEST_CASE("Check parsing results") {
...
+
+    file = Trajectory("data/smi/bad/bad_symbol.smi");
+    CHECK_THROWS_WITH(file.read(), "SMI Reader: unknown symbol: '`'");
+
+    file = Trajectory("data/smi/bad/misplaced_property.smi");
+    CHECK_THROWS_WITH(file.read(), "SMI Reader: symbol not allowed outside of property: '@'");
+
+    auto bad = std::string("C)");
+    file = Trajectory::memory_reader(bad.c_str(), bad.size(), "SMI");
+    CHECK_THROWS_WITH(file.read(), "SMI Reader: unmatched ')'");
```
frodofine: Off-topic, but would an additional overload of `Trajectory::memory_reader` with `std::string` be useful?

Luthaf: That, or with a `span<const char>`. Even better if we can get automatic conversions from both `std::string` and `std::vector<char>` =)

PR #318, opened 2020-02-26
feat: NetCDF scale factor

### LINKED ISSUE #268: Support scale_factor in NetCDF
See http://ambermd.org/netcdf/nctraj.xhtml:

> Creators may define a scale_factor attribute of type float for each variable. Creators shall ensure that the units of data values, after being multiplied by the value of scale_factor (if it exists) are equal to that described by the units attribute. If a scale_factor attribute exists for a variable, readers shall multiply data values by the value of the scale_factor attribute before interpreting the data. This scaling burden is placed on the reader rather than the creator, as writing data is expected to be a more time-sensitive operation than reading it.

ezavod:
Found some old script that uses `scale_factor` in NetCDF extensively. This PR implements scaling of cell lengths, angles, coordinates and velocities by `scale_factor`.
Fixes #268.

### DIFF
--- a/include/chemfiles/files/NcFile.hpp
+++ b/include/chemfiles/files/NcFile.hpp
@@ -53,12 +53,12 @@ namespace nc {
         /// Get the dimensions size for this variable
         std::vector<size_t> dimmensions() const;
 
-        /// Get the attribute `name`.
-        std::string attribute(const std::string& name) const;
+        /// Get the string attribute `name`.
+        std::string string_attribute(const std::string& name) const;
         /// Get the float attribute `name`.
-        float attribute_float(const std::string& name) const;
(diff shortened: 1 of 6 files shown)


include/chemfiles/files/NcFile.hpp
```diff
@@ -55,9 +55,14 @@ namespace nc {
 
         /// Get the attribute `name`.
         std::string attribute(const std::string& name) const;
```
Luthaf: How about renaming this one to `string_attribute`, and the one below to `float_attribute`? (float_attribute feels more natural than attribute_float, and if the float one has the type in the name, the string one should too)

ezavod: That's a good idea. It's more descriptive and consistent.

PR #322, opened 2020-03-24
Expose {Frame,Topology}::clear_bonds to the C API

Luthaf:
They are part of the public C++ API, so I think they should be part of the public C API too. No idea why they were left out initially.

### DIFF
--- a/include/chemfiles/capi/frame.h
+++ b/include/chemfiles/capi/frame.h
@@ -322,6 +322,14 @@ CHFL_EXPORT chfl_status chfl_frame_remove_bond(
     CHFL_FRAME* frame, uint64_t i, uint64_t j
 );
 
+/// Remove all existing bonds, angles, dihedral angles and improper dihedral
+/// angles in the `frame`.
+///
+/// @example{capi/chfl_frame/clear_bonds.c}
+/// @return The operation status code. You can use `chfl_last_error` to learn
+///         about the error if the status code is not `CHFL_SUCCESS`.
+CHFL_EXPORT chfl_status chfl_frame_clear_bonds(CHFL_FRAME* frame);
+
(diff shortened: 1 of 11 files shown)


include/chemfiles/capi/frame.h
```diff
@@ -322,6 +322,14 @@ CHFL_EXPORT chfl_status chfl_frame_remove_bond(
     CHFL_FRAME* frame, uint64_t i, uint64_t j
 );
 
+/// Remove all existing bond, angles, dihedral angles and improper dihedral
```
ezavod, line 325: Sorry for being pedantic, but bond*s* (plural) sounds better :P

include/chemfiles/capi/topology.h
```diff
@@ -198,6 +198,14 @@ CHFL_EXPORT chfl_status chfl_topology_remove_bond(
     CHFL_TOPOLOGY* topology, uint64_t i, uint64_t j
 );
 
+/// Remove all existing bond, angles, dihedral angles and improper dihedral
```
ezavod, line 201: Same as in frame.

Luthaf:
> Those functions should also be accessible from rust, python and other language bindings. Maybe one can open an issue there so that it won't be forgotten?

I don't plan on backporting these to 0.9, so they will be wrapped when we release 0.10. Each binding repo already have a small script checking that all C API functions are exposed (https://github.com/chemfiles/chemfiles.py/blob/d795fd6b2df67dcb7120569d5224d0b5dd9c398d/scripts/check-used-functions.py for example), so if we forget to expose these the tests should fail 😃

include/chemfiles/capi/frame.h
Luthaf: thanks! that's the point of review too =)

PR #323, opened 2020-03-25
Unify chfl_trajectory_path with other functions getting string values

Luthaf:
All the other functions make a copy of the string (chfl_atom_name, chfl_selection_string), this one did it differently for no good reason (it is not a performance sensitive function)


Luthaf:
Thanks! The link should be fixed now!

PR #324, opened 2020-03-28
Add "atom_type" propertie in Tinker format

sguionni:


src/formats/Tinker.cpp
```diff
@@ -70,6 +70,7 @@ void TinkerFormat::read_next(Frame& frame) {
             std::string name;
             auto count = scan(line, id, name, x, y, z, atom_type);
 
+            frame[i].set("atom_type", atom_type);
```
Luthaf: This will fail, since you are trying to access atom `i` in the frame before it is added to the frame (see the line below). You need to set the property after adding the atom.

PR #326, opened 2020-04-02
feat: allow appending to gz files

ezavod:
Currently it is not allowed to append to .gz files.
This change feels too easy, but tests work :D

### DIFF
--- a/src/Format.cpp
+++ b/src/Format.cpp
@@ -64,7 +64,7 @@ void TextFormat::write_next(const Frame& /*unused*/) {
 #pragma GCC diagnostic pop
 #endif
 
-TextFormat::TextFormat(std::string path, File::Mode mode, File::Compression compression):
+TextFormat::TextFormat(std::string path, File::Mode mode, File::Compression compression) :
     file_(std::move(path), mode, compression) {}
 
 TextFormat::TextFormat(std::shared_ptr<MemoryBuffer> memory, File::Mode mode, File::Compression compression) :
@@ -73,6 +73,13 @@ TextFormat::TextFormat(std::shared_ptr<MemoryBuffer> memory, File::Mode mode, Fi
 void TextFormat::scan_all() {
     if (eof_found_) {
         return;
+    }
+
+    optional<TextFile> tmp_read_file = nullopt;
+    if (file_.mode() == File::Mode::APPEND && file_.compression() == File::Compression::GZIP) {
+        tmp_read_file = TextFile(file_.path(), File::Mode::READ, file_.compression());
+        // `forward()` needs a readable file
+        std::swap(*tmp_read_file, file_);
     }
 
     auto before = file_.tellpos();
@@ -87,6 +94,11 @@ void TextFormat::scan_all() {
     eof_found_ = true;
     // reset failbit in the file
     file_.clear();
+
+    if (tmp_read_file) {
+        // use original file for all further write operations
+        std::swap(file_, *tmp_read_file);
+    }
 
     if (before == 0 && !steps_positions_.empty()) {
         file_.seekpos(steps_positions_[0]);
--- a/src/files/GzFile.cpp
+++ b/src/files/GzFile.cpp
@@ -34,7 +34,8 @@ GzFile::GzFile(const std::string& path, File::Mode mode): TextFileImpl(path) {
         openmode = "wb7";
         break;
     case File::APPEND:
-        throw file_error("appending (open mode 'a') is not supported with gziped files");
+        openmode = "ab7";
+        break;
     }
 
     file_ = gzopen64(path.c_str(), openmode);
--- a/tests/files/gz-file.cpp
+++ b/tests/files/gz-file.cpp
@@ -49,11 +49,6 @@ TEST_CASE("Read a text file") {
             GzFile("not existing", File::READ),
             "could not open the file at 'not existing'"
         );
-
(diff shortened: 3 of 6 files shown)


tests/files/gz-file.cpp
```diff
@@ -109,6 +104,55 @@ TEST_CASE("Write a gz file") {
...
+
+    auto expected = std::vector<uint8_t> {
+        0x1f, 0x8b, 0x08, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0xff, 0x73,
+        0x2c, 0x28, 0x48, 0xcd, 0x4b, 0x51, 0x30, 0xe4, 0x32, 0x37, 0x33,
+        0x31, 0xe5, 0x02, 0x00, 0xf8, 0x06, 0xaf, 0x8d, 0x0e, 0x00, 0x00,
+        0x00, 0x1f, 0x8b, 0x08, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x03,
+        0x73, 0x2c, 0x28, 0x48, 0xcd, 0x4b, 0x51, 0x30, 0xe2, 0x32, 0x33,
+        0x37, 0x35, 0xe1, 0x02, 0x00, 0xc6, 0x09, 0x42, 0x21, 0x0e, 0x00,
+        0x00, 0x00,
+    };
```
ezavod, line 143: I am not really sure on how to get the test file contents, so I simply run the test to produce a test file and check this file with `gzip -l <temp_file>.gz`. Since `gzip` doesn't result in any errors, I assume that the file contents are valid and used them as hard coded reference...

src/files/GzFile.cpp
```diff
@@ -34,7 +34,8 @@ GzFile::GzFile(const std::string& path, File::Mode mode): TextFileImpl(path) {
         openmode = "wb7";
         break;
     case File::APPEND:
-        throw file_error("appending (open mode 'a') is not supported with gziped files");
+        openmode = "ab";
```
Luthaf, line 37: We may want to open the file in 'ab7' mode, to ensure the same compression ratio is used as in writing mode.

tests/files/gz-file.cpp
Luthaf: Do you mean checking what the file content does mean? In the write test it is done mostly as a regression test, the exact content is not that important.

ezavod: I was wondering if there is a systematic way to derive the bytes that are expected. Technically it is possible that my file is garbage and I'm comparing against a broken file. But this does not matter for a regression test and I assume that the file is not broken because gzip can read it. Just wanted to be sure.

ezavod:
Thanks for the fast comments. Will edit the changelog soon.
I will also test on an actual trajectory to make sure seeking works.

src/files/GzFile.cpp
ezavod: Makes good sense.

ezavod:
I hope I got everything together now...
The main problem is that zlib does not support reading a file that is opened in write/append mode (see [docs of `gzopen`](https://www.zlib.net/manual.html#Gzip)). When appending to a file chemfiles needs to know the number of frames that are already present in order to give the total number of frames via `nframes()`. `scan_all()` fails when appending because it needs to read all previous frames. The new flag `steps_initialized_` causes chemfiles to open a readable file handle and get the number of already present frames (and positions) from there. This is similar to [the way we do this for xdr files](https://github.com/chemfiles/chemfiles/blob/93832d1f7c7bc38a6440394bfe506ddb4f54a887/src/files/XDRFile.cpp#L42-L44). If that's done, `eof_found_` is true because appending takes place at the end of the file.
As the [docs of `gzopen`](https://www.zlib.net/manual.html#Gzip) state as well, appending creates a new stream. This is no problem for reading and seeking (see test for multiple streams), but obviously decreases the compression efficiency. A new header is written for every appended data block (meaning if a file is newly opened in append mode) and the compression algorithm only conseideres data within this block. This would only be a problem if one would open a file, append just a few bytes and close the file again. Since chemfiles typically appends units of frames that are rather large and one typically appends multiple frames per file opening operation, this should not matter for us.

Luthaf:
Ok, thanks for the detailed explanation!

> When appending to a file chemfiles needs to know the number of frames that are already present in order to give the total number of frames via nframes(). scan_all() fails when appending because it needs to read all previous frames. The new flag steps_initialized_ causes chemfiles to open a readable file handle and get the number of already present frames (and positions) from there.

Instead of adding a separate function, do you think it is possible to re-open the file in read mode inside `scan_all` directly? Also, I would only re-open the file if needed, i.e. check on the `compression` parameter.

Luthaf:
I have one last improvement suggestion, and then this looks good to go!

src/Format.cpp
```diff
@@ -75,6 +75,13 @@ void TextFormat::scan_all() {
         return;
     }
 
+    TextFile* tmp_read_file = nullptr;
```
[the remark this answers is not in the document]
ezavod: It doesn't need no be a pointer. Thanks for hinting me to `optional`  =)

ezavod:
Is the failure of the doc test unrelated to the changes of this PR (seems to be a problem with sphinx)? Just want to make sure before merging.

Luthaf:
From some local testing it looks like an issue with breath 4.14.2, fortunately it looks like they already fixed it in 4.15. I restarted the build, it should work now

PR #329, opened 2020-04-22
Check to ensure end IDs are greater than start IDs in the PDB reader

frodofine:
Turns out the fix is really simple. The way to misrepresent PDB files will never cease to amaze me!

### DIFF
--- a/src/formats/PDB.cpp
+++ b/src/formats/PDB.cpp
@@ -232,6 +232,12 @@ void PDBFormat::read_HELIX(string_view line) {
         return;
     }
 
+    // Fix 1HTQ
+    if (start > end) {
+        warning("PDB reader", "HELIX starting residue {} is greater than ending id {}", start, end);
+        return;
+    }
+
     // Convert the code as a character to its numeric meaning.
     // See http://www.wwpdb.org/documentation/file-format-content/format23/sect5.html
     // for definitions of these numbers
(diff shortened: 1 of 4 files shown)


src/formats/PDB.cpp
```diff
@@ -232,6 +232,12 @@ void PDBFormat::read_HELIX(string_view line) {
         return;
     }
 
+    // Fix 1HTQ
+    if (start > end) {
```
Luthaf, line 236: how about inverting `start` and `end` in this case (and still emit the warning)? If the file contains an `HELIX` record, it is very likely to come from the PDB archive, and users might be interested in the secondary structure. 

Plus I could not find anything in the standard saying that `start` had to be lower than `end`

Luthaf:
Thanks a lot! Travis failure is unrelated, I'll give it a look before merging this if I can.

I think we should try to read the full PDB archive to check for other issues like this. Is this something you did with lemon?

src/formats/PDB.cpp
frodofine: The PDB standard is a bit of a mess, residue IDs can be negative! Also, it is a bit more complicated than swapping the `start` and `end` ids because the residue IDS in the PDB file are not sorted properly. For example, the `504` residue is between the `351` and `352` residues and the way it's implemented would cause more residues to be labeled than appropriate.  In fact, other software (e.g. PyMOL) doesn't interpret this residue at all.  So, I think emitting a warning and not assigning the group is the best solution for now.

frodofine:
I resubmitted the failed Travis job.

This can be done, but Lemon does this using the MMTF format, which avoids these issues.

src/formats/PDB.cpp
Luthaf: ok, sounds good

PR #332, opened 2020-05-04
Check residue properties for selections

### LINKED ISSUE #331: Expose Residue properties to selections
From @mdimura  on gitter:

> How do I properly specify a property in a selection string? I tried "[chainname] A", but it gives an empty selection. Sorry, could not find an example in the docs.

The issue here is that `chainname` is a residue property, and only atomic properties are checked when evaluating selection: https://github.com/chemfiles/chemfiles/blob/49442411ab9c31680e03918b151ac3fa26c1958b/src/selections/expr.cpp#L323-L338

I think the best way to solve this is to give the same semantics as `resname` selections: if no property is found on the atom, check if the current atom is part of a residue, and if the residue contains the property.

Any thoughts?

mdimura:
Resolve #331

### DIFF
--- a/doc/src/selections.rst
+++ b/doc/src/selections.rst
@@ -75,12 +75,14 @@ Constraints on atomic properties
 --------------------------------
 
 It is possible to use atomic :cpp:class:`properties <chemfiles::Property>`
-(unfortunately not frame or residue properties) in constraints, with the
+(unfortunately not frame properties) in constraints, with the
 ``[<property>]`` syntax. ``<property>`` should be replaced by the property name:
 ``[is_hetatm]``, possibly using quotes around the property name if it contains
 spaces: ``["my own property"]``. Depending on the context, a Boolean, string or
-numeric property will be searched. If none can be found, or if the property type
-does not match, a default value will be used instead.
+numeric property will be searched. If an atomic property with a given name cannot
+be found, residue property is searched instead, however, atomic properties take 
+precedence. If none can be found, or if the property type does not match, a 
+default value will be used instead.
 
 Selection language reference
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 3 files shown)


src/selections/expr.cpp
```diff
@@ -167,6 +167,13 @@ bool BoolProperty::is_match(const Frame& frame, const Match& match) const {
             );
         }
     } else {
+        auto residue = frame.topology().residue_for_atom(match[argument_]);
+        if (residue) {
+            const auto& resProperty = residue->get(property_);
+            if (resProperty && resProperty->kind() == Property::BOOL) {
+                return resProperty->as_bool();
+            }
+        }
```
Luthaf, line 176: I would add the same type checking as above, something like 

```cpp
throw selection_error(
    "invalid type for property [{}] on the residue containing atom {}: expected bool, got {}",
    property_, match[argument_], kind_as_string(property->kind())
);
```

Luthaf:
Thanks a lot for doing this! I forgot about documentation in my last comment, would you mind adding a note in `doc/src/selections.rst` that residues properties will be checked if no atomic property is found? Sorry I should have asked for this in the first review.

doc/src/selections.rst
```diff
@@ -118,7 +120,7 @@ String properties
 - ``resname``: gives the residue name. If an atom is not in a residue, this
   return the empty string;
 - ``[<property>]``: gives the value of the string property named `'property'`
-  for the atom. This will return an empty string (`""`) if the property is not
+  for the atom or the reside. This will return an empty string (`""`) if the property is not
```
Luthaf, line 123:
```suggestion
  for the atom or the residue containing the atom, if any. This will return an empty string (`""`) if the property is not
```

doc/src/selections.rst
```diff
@@ -133,7 +135,7 @@ Most of the numeric properties only apply to a single atom:
 - ``resid``: gives the atomic residue index. If an atom is not in a residue,
   this return -1;
 - ``[<property>]``: gives the value of the numeric property named `'property'`
-  for the atom. This will return 0 if the property is not set;
+  for the atom or the residue. This will return 0 if the property is not set;
```
Luthaf, line 138:
```suggestion
  for the atom or the residue containing the atom, if any. This will return 0 if the property is not set;
```

mdimura:
No problem, done!

mdimura:
Thank you for your help!

PR #335, opened 2020-05-08
Fix a bug in PDB::forward

Luthaf:
Without this chemfiles would only find 152 steps inside `data/pdb/ase.pdb` instead of the expected 156


src/formats/PDB.cpp
```diff
@@ -801,7 +801,13 @@ optional<uint64_t> PDBFormat::forward() {
     while (!file_.eof()) {
         auto line = file_.readline();
 
+        std::string saved_line;
         if (line.substr(0, 6) == "ENDMDL") {
+            // save the current line into an owned string, since the underlying
+            // buffer in file_ can change in file_->readline();
+            saved_line = line.to_string();
+            line = saved_line;
+
```
frodofine, line 810: I'm surprised this fixed the issue, but I believe the data!  if the statements are executed, does the location of `save` and `next` change by calling `to_save`? The `file_` object doesn't seem to change, but maybe it does? Or is it due to the next statement not executing (on line 822) because **ENDMDL** and **END** no longer have the same first 3 characters?

Luthaf: My understanding is that `line` points to `file_.buffer`, and `next = file_.readline()` do not find the end of the line in the current buffer, causing the buffer to be filled with more data in https://github.com/chemfiles/chemfiles/blob/99486730d4849656308516c6f7ed60b74fd575bc/src/File.cpp#L192-L194

Since we are close to the end of file, only the beginning of the buffer actually contains data, the end of it if filled with 0:
https://github.com/chemfiles/chemfiles/blob/99486730d4849656308516c6f7ed60b74fd575bc/src/File.cpp#L130-L134

Meaning `line` is now equal to `"\0\0\0\0\0\0"`, and no longer `ENDMDL`, and we exit the function at the bottom 

Overall, this bug is an unfortunate side-effect of using string_view (i.e. borrowing views into data), where the owned data can change below us. This is something that would have been caught by the compiler in Rust ¯\\_(ツ)_/¯.

PR #337, opened 2020-05-14
Allow to use sub-selection in numerical functions (distance/angle/...)

### LINKED ISSUE #327: Allow sub-selections in distance/angle/...
See the initial discussion at https://discourse.julialang.org/t/what-is-the-best-molecular-dynamics-analysis-library-in-julia/36865.

Most of the functionality is already here; the idea would be to allow [SubSelections](https://github.com/chemfiles/chemfiles/blob/93832d1f7c7bc38a6440394bfe506ddb4f54a887/include/chemfiles/selections/expr.hpp#L122) inside `distance`, `angle`, `dihedral` and `out_of_plane` in addition to `#1/#2/...`. If someone want to take on this, the hardest part for this would be updating the selection parser to allow selections inside the above functions.

Luthaf:
Fixes #327 

Ping @dkonstan, since you were interested in using this. You should be able to compile this branch and use the code with Julia if you want to test before the next release.

TODO:
- [x] update CHANGELOG
- [x] add documentation for this and sub-selections in boolean selectors

### DIFF
--- a/src/selections/parser.cpp
+++ b/src/selections/parser.cpp
@@ -58,6 +58,10 @@ static std::map<std::string, numeric_functions_creator_t> NUMERIC_FUNCTIONS = {
     {"asin", [](MathAst ast){ return chemfiles::make_unique<Function>(static_cast<double (*)(double)>(asin), "asin", std::move(ast));}},
     {"acos", [](MathAst ast){ return chemfiles::make_unique<Function>(static_cast<double (*)(double)>(acos), "acos", std::move(ast));}},
     {"sqrt", [](MathAst ast){ return chemfiles::make_unique<Function>(static_cast<double (*)(double)>(sqrt), "sqrt", std::move(ast));}},
+    {"exp", [](MathAst ast){ return chemfiles::make_unique<Function>(static_cast<double (*)(double)>(exp), "exp", std::move(ast));}},
+    {"log", [](MathAst ast){ return chemfiles::make_unique<Function>(static_cast<double (*)(double)>(sqrt), "log", std::move(ast));}},
+    {"log2", [](MathAst ast){ return chemfiles::make_unique<Function>(static_cast<double (*)(double)>(log2), "log2", std::move(ast));}},
+    {"log10", [](MathAst ast){ return chemfiles::make_unique<Function>(static_cast<double (*)(double)>(log10), "log10", std::move(ast));}},
     {"rad2deg", [](MathAst ast){ return chemfiles::make_unique<Function>([](double rad){ return rad * 180 / PI; }, "rad2deg", std::move(ast));}},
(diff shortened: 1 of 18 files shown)


src/selections/parser.cpp
```diff
@@ -13,93 +14,145 @@
...
 };
 
 static std::map<std::string, NumericVariableFunction> NUMERIC_VAR_FUNCTIONS = {
-    {"distance", {2, [](std::vector<Variable> args){ return MathAst(new Distance(args[0], args[1])); }}},
-    {"angle", {3, [](std::vector<Variable> args){ return MathAst(new Angle(args[0], args[1], args[2])); }}},
-    {"dihedral", {4, [](std::vector<Variable> args){ return MathAst(new Dihedral(args[0], args[1], args[2], args[3])); }}},
-    {"out_of_plane", {4, [](std::vector<Variable> args){ return MathAst(new OutOfPlane(args[0], args[1], args[2], args[3])); }}},
+    {"distance", {2, [](SelectionArguments args) {
+        assert(args.count == 2);
+        return make_unique<Distance>(
```
ezavod, line 78: Should one use `chemfiles::make_unique` as before?
This applies for the whole map.

src/selections/parser.cpp
```diff
@@ -13,93 +14,145 @@
...
+using bool_selector_creator_t = std::function<Ast(SelectionArguments)>;
 struct BooleanFunction {  // NOLINT: constructor does not initialize these fields: arity
     unsigned arity;
     bool_selector_creator_t creator;
 };
 
 static std::map<std::string, BooleanFunction> BOOLEAN_SELECTORS = {
-    {"all", {0, [](std::vector<SubSelection> /*unused*/) {  // NOLINT: can not make std::vector<SubSelection> a const reference
+    {"all", {0, [](SelectionArguments /*unused*/) {  // NOLINT: can not make std::vector<SubSelection> a const reference
         return Ast(new All());
```
ezavod, line 120: Is there any reason why we don't use `chemfiles::make_unique<All>()` here?
This applies for the whole map.

src/selections/parser.cpp
Luthaf: Do you mean explicitly using the `chemfiles::` namespace? That should not be required, but MSVC 15 fails to perform correct ADL. It do not hurt to use it though

src/selections/parser.cpp
Luthaf: No reason, and it make sense to use it. I'll change this!

src/selections/parser.cpp
ezavod: >Do you mean explicitly using the chemfiles:: namespace?

Yes, I was wondering because in your other commit "Resolve ambiguous call to make_unique for MSVC" the namespace was added in a bunch of places.

PR #339, opened 2020-05-16
Add symmetry operations to MMTF

frodofine:
Everything is implemented except for adding symmetric bonds.

### DIFF
--- a/src/formats/MMTF.cpp
+++ b/src/formats/MMTF.cpp
@@ -442,6 +442,22 @@ void MMTFFormat::write(const Frame& frame) {
     structure_.chainsPerModel.emplace_back(0);
     structure_.numAtoms += static_cast<int32_t>(frame.size());
 
+    if (mmtf::isDefaultValue(structure_.unitCell)) {
+        auto& cell = frame.cell();
+        structure_.unitCell.resize(6);
+        structure_.unitCell[0] = static_cast<float>(cell.a());
+        structure_.unitCell[1] = static_cast<float>(cell.b());
+        structure_.unitCell[2] = static_cast<float>(cell.c());
+        structure_.unitCell[3] = static_cast<float>(cell.alpha());
+        structure_.unitCell[4] = static_cast<float>(cell.beta());
+        structure_.unitCell[5] = static_cast<float>(cell.gamma());
+
+        unitcellForWrite_ = cell;
+    }
+    else if (unitcellForWrite_ != frame.cell()) {
+        warning("MMTF Writer", "the MMTF format only allows one unit cell to be defined for all models, using the first one defined");
+    }
+
     const auto& topology = frame.topology();
 
     // pre-allocate some memory
@@ -526,6 +542,7 @@ void MMTFFormat::write(const Frame& frame) {
 }
 
 void MMTFFormat::add_residue_to_structure(const Frame& frame, const Residue& residue) {
+
     structure_.numGroups++;
     structure_.groupsPerChain.back() += 1;
 
--- a/src/formats/MMTF.cpp
+++ b/src/formats/MMTF.cpp
@@ -103,6 +103,7 @@ void MMTFFormat::read_step(const size_t step, Frame& frame) {
     groupIndex_ = 0;
     atomIndex_  = 0;
     atomSkip_   = 0;
+    interBondIndex_ = 0;
 
     // Fast-forward, keeping all indexes updated
     while(modelIndex_ != step) {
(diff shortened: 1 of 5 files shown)


src/formats/MMTF.cpp
```diff
@@ -301,6 +305,115 @@ void MMTFFormat::read(Frame& frame) {
     modelIndex_++;
 
     atomSkip_ = atomIndex_;
+
+    const auto original_size = frame.size();
+
```
Luthaf, line 310: Might be nice to split up `MMTF::read()` in multiple sub-functions (`MMTF::read_atoms`, `MMTF::read_secondary`, `MMTF::read_symmetries` ...), it is starting to grow quite large =)

src/formats/MMTF.cpp
```diff
@@ -301,6 +305,115 @@ void MMTFFormat::read(Frame& frame) {
...
+            for (auto&& residue : residues_to_add) {
+                frame.add_residue(std::move(residue));
+            }
+            
+            /*
+
+            The code below is way too slow, even in release mode. It is probably
+            due to the ordered nature of the bonds array. Keeping the first, but
+            not second loop causes the slow down.
```
Luthaf, line 391: so what you want here is the list of bonds containing a given atom? How about keeping such list inside the MMTF parser itself, and only inserting bonds in the frame when they have all been found, i.e. at the end of `read`?

src/formats/MMTF.cpp
frodofine: I agree! I've refactored it, some of the comments maybe drowned out as a result.

src/formats/MMTF.cpp
```diff
@@ -301,6 +305,115 @@ void MMTFFormat::read(Frame& frame) {
...
+
+                auto ss = residue.get("secondary_structure");
+                if (ss) {
+                    new_residue.set("secondary_structure", *ss);
+                }
+
+                for (auto atom_id : residue) {
+
+                    if (atom_id >= original_size) {
+                        continue;
```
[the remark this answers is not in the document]
frodofine: Done. It's just to ensure that we don't apply symmetry operations to other symmetry copies.  For example, there are 68 symmetry copies in 5IRE and they they added sequentially, so there is a possibility to apply a symmetry operation to an atom added through a previous symmetry operation, leading to additional unnecessary atoms being added.

src/formats/MMTF.cpp
```diff
@@ -301,6 +305,115 @@ void MMTFFormat::read(Frame& frame) {
...
+                if (!asmbl || asmbl->as_string() != "bio" + assembly.name) {
+                    continue;
+                }
+
+                auto chainindex = residue.get("chainindex");
+                if (!chainindex || chains_to_transform.count(chainindex->as_double()) == 0) {
+                    continue;
+                }
+
+                auto new_residue = Residue(residue.name(), *residue.id());
```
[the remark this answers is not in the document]
frodofine: Yes, that's intensional. The residue ID is only unique to the given chain. For example, the structure 4XUF has two chains (A and B), each is a complete copy of the other. The residue IDs for the two chains repeat (chain A has residues 600-947 **and** chain B has residues 600-947). Note that in this case, chain A and chain B are different biological assemblies. For cases where multiple chains make up the same biological assembly, the case is the same. Take 4HHB, which has chains A, B, C, and D and each chain has residues 1-141. Here, we are dealing with symmetry operations that take chains and 'copy' them to create what one would expect to see in a cell. The example here is 5IRE which has 6 chains (A, B, C, D, E, and F). In the actual crystallographic file, there is only one copy of each chain (note that A, C, and E have the same residue IDs and B, D, and F have the same residue IDs, but all 6 chains could be unique, it doesn't matter). In the biological assembly, these chains are copied multiple times to create the capsid for the Zika virus (see [this file](https://files.rcsb.org/download/5IRE.pdb1.gz) for a version of this file where the symetry operations have already been applied). Here, the residue IDs are only unique to the chain within a given biological assembly and it is this interpretation that I am following here.

src/formats/MMTF.cpp
frodofine: This worked!

src/formats/MMTF.cpp
```diff
@@ -392,6 +526,24 @@ void MMTFFormat::write(const Frame& frame) {
 }
 
 void MMTFFormat::add_residue_to_structure(const Frame& frame, const Residue& residue) {
+
+    if (frame.cell().shape() != UnitCell::INFINITE) {
```
Luthaf, line 530: I though I already left this comment, but I cant' find it now 🤷‍♂️ .

Instead of checking for infinite cell, we should check that the same cell is used for all frames, i.e. refuse `cell(10, 12, 10)` and then `cell(10, 10, 10)`.

frodofine: I have an email with a similar comment, but it never appeared here! Should be addressed now.

Luthaf:
Looks good to me, feel free to merge if you are done!

tests/formats/mmtf.cpp
```diff
@@ -176,6 +176,17 @@ TEST_CASE("Read files in MMTF format") {
         CHECK(frame[*edo.begin()].get("altloc")->as_string() == "B");
     }
 
+    SECTION("Check symmetry") {
+    #ifndef CHEMFILES_WINDOWS
+        if (!is_valgrind_and_travis()) {
+            auto file = Trajectory("data/mmtf/5IRE.mmtf");
+
+            auto frame = file.read();
+            CHECK(frame.size() == 661440);
```
Luthaf, line 185: It would be nice to have a smaller file to check symmetry operations. If there is none on the PDB archive, how about an "artificial" file, manually created?

frodofine: I purposefully chose an entry with a large number of symmetry operations, but I can replace the test with an entry with a smaller number.

frodofine: Turns out 5A1I is also an entry which reuires symmetry operations, so I just exanded the test. I'll merge this tomorrow unless there's a major objection.

PR #340, opened 2020-05-21
PDB Write changes

frodofine:
This PR stops the PDB writer from writing `CONECT` records for standard residues and writes `TER` records at the end of chains. It also attempts to convert some large residue IDs to see if they are actually negative numbers.

### DIFF
--- a/src/formats/GRO.cpp
+++ b/src/formats/GRO.cpp
@@ -64,9 +64,9 @@ void GROFormat::read_next(Frame& frame) {
             );
         }
 
-        size_t resid = SIZE_MAX;
+        int64_t resid = std::numeric_limits<int64_t>::max();
         try {
-            resid = parse<size_t>(line.substr(0, 5));
+            resid = parse<int64_t>(line.substr(0, 5));
         } catch (const Error&) {
             // Invalid residue, we'll skip it
         }
@@ -87,7 +87,7 @@ void GROFormat::read_next(Frame& frame) {
         }
         frame.add_atom(Atom(name), {x, y, z}, {vx, vy, vz});
 
-        if (resid != SIZE_MAX) {
+        if (resid != std::numeric_limits<int64_t>::max()) {
             if (residues_.find(resid) == residues_.end()) {
                 Residue residue(resname, resid);
                 residue.add_atom(frame.size() - 1);
@@ -159,7 +159,7 @@ void GROFormat::write_next(const Frame& frame) {
     // Only use numbers bigger than the biggest residue id as "resSeq" for
     // atoms without associated residue, and start generated residue id at
     // 1
-    uint64_t max_resid = 1;
+    int64_t max_resid = 1;
     for (const auto& residue: frame.topology().residues()) {
         auto resid = residue.id();
         if (resid && resid.value() > max_resid) {
@@ -185,7 +185,7 @@ void GROFormat::write_next(const Frame& frame) {
 
         if (residue && residue->id()) {
             auto value = residue->id().value();
-            if (value <= 99999) {
+            if (value <= 99999 && value >= 0) {
                 resid = std::to_string(value);
             } else {
                 warning("GRO writer", "too many residues, removing residue id");
--- a/src/formats/GRO.cpp
+++ b/src/formats/GRO.cpp
@@ -64,7 +64,7 @@ void GROFormat::read_next(Frame& frame) {
             );
         }
 
-        int64_t resid = std::numeric_limits<int64_t>::max();
+        optional<int64_t> resid = nullopt;
         try {
             resid = parse<int64_t>(line.substr(0, 5));
         } catch (const Error&) {
@@ -87,16 +87,18 @@ void GROFormat::read_next(Frame& frame) {
         }
         frame.add_atom(Atom(name), {x, y, z}, {vx, vy, vz});
 
-        if (resid != std::numeric_limits<int64_t>::max()) {
-            if (residues_.find(resid) == residues_.end()) {
-                Residue residue(resname, resid);
-                residue.add_atom(frame.size() - 1);
+        if (!resid) {
+            continue;
+        }
 
-                residues_.insert({resid, residue});
-            } else {
-                // Just add this atom to the residue
-                residues_.at(resid).add_atom(frame.size() - 1);
-            }
+        if (residues_.find(*resid) == residues_.end()) {
+            Residue residue(resname, *resid);
+            residue.add_atom(frame.size() - 1);
+
+            residues_.insert({*resid, residue});
+        } else {
+            // Just add this atom to the residue
+            residues_.at(*resid).add_atom(frame.size() - 1);
         }
     }
 
@@ -185,7 +187,13 @@ void GROFormat::write_next(const Frame& frame) {
 
         if (residue && residue->id()) {
             auto value = residue->id().value();
-            if (value <= 99999 && value >= 0) {
+            if (value <= 0) {
+                warning("GRO writer", "the residue id '{}' should not be negative or zero, treating it as blank", value);
+                auto value = max_resid++;
+                if (value <= 99999) {
+                    resid = std::to_string(value);
+                }
+            } else if (value <= 99999) {
                 resid = std::to_string(value);
             } else {
                 warning("GRO writer", "too many residues, removing residue id");
--- a/src/formats/PDB.cpp
+++ b/src/formats/PDB.cpp
@@ -661,6 +661,9 @@ void PDBFormat::write_next(const Frame& frame) {
         }
     }
 
+    // Used to skip writing unnecessary connect records
+    auto is_atom_record = std::vector<bool>(frame.size(), false);
+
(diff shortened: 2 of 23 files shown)


src/formats/PDB.cpp
```diff
@@ -639,6 +639,102 @@ static std::string to_pdb_index(uint64_t i) {
...
+
+    ResidueInformation res_info;
+
+    if (!residue_opt) {
+        auto value = max_resid++;
+        if (value < 9999) {
+            res_info.resid = to_pdb_index(value);
+        }
+        else {
+            res_info.resid = "  -1";
```
Luthaf, line 661: Should we emit a warning? 

Also, since we are touching this code, should we use some kind of hexadecimal numbering scheme if we have more than 9999 residues? I know that's what VMD is doing for such "non standard" files: https://www.ks.uiuc.edu/Research/vmd/mailing_list/vmd-l/7384.html

Also, do we want to use `-1` or another special value, such as `*****`/`0`/`    `?

src/formats/PDB.cpp
```diff
@@ -639,6 +639,102 @@ static std::string to_pdb_index(uint64_t i) {
...
+        warning("PDB writer", "residue '{}' name is too long, it will be truncated", res_info.resname);
+        res_info.resname = res_info.resname.substr(0, 3);
+    }
+
+    if (residue.id()) {
+        auto value = residue.id().value();
+        if (value > 9999) {
+            auto negative_conversion = std::to_string(static_cast<int64_t>(value));
+            if (negative_conversion.length() <= 4) {
+                warning("PDB writer", "it appears {} is actually negative, treating as {}", value, negative_conversion);
```
Luthaf, line 685: Are residue id potentially negative in standard PDB? We could change how we store the id if this is the case

src/formats/PDB.cpp
```diff
@@ -670,101 +774,76 @@ void PDBFormat::write_next(const Frame& frame) {
...
+            continue;
+        }
         if (bond[0] > 99999 || bond[1] > 99999) {
             warning("PDB writer", "atomic index is too big for CONNECT, removing the bond between {} and {}", bond[0], bond[1]);
             continue;
         }
-        connect[bond[0]].push_back(bond[1]);
-        connect[bond[1]].push_back(bond[0]);
+
+        auto lower_b0 = std::lower_bound(ter_serial_numbers.begin(),
```
Luthaf, line 821: This needs a comment explaining what is going on =)

src/formats/PDB.cpp
Luthaf: the hexadecimal thing have a standard: http://cci.lbl.gov/hybrid_36/

frodofine: I think this is a great idea! They provide sample code as well and it appears PHENIX (a popular program for manipulating PDBs) supports this standard. I'll check to see if PyMOL supports the standard, but it looks unlikely (maybe I'll add it there as well).

> Also, do we want to use -1 or another special value, such as `*****`/`0`/` `?

I'm not sure. I'll see what other software supports. My biggest worry is that non-integer characters will cause problems with other software.

src/formats/PDB.cpp
frodofine: Sure thing!

src/formats/PDB.cpp
Luthaf: Yes it will =). But that's fine, the next release will be a breaking one. I have other improvement idea for residues, I'll write an issue to discuss them!

Luthaf: Should negative resid also be handled by mmCIF & MMTF parsers?

frodofine: Sounds good! I think MMTF supports them already since it uses signed integers to store residue IDs, but mmCIF definetly should.

src/parse.cpp
```diff
@@ -216,3 +216,127 @@ template <> double chemfiles::parse(string_view input) {
...
+
+    // This function is only called within chemfiles for fixed format files.
+    // Therefore, the width should also be the length of the string as this is
+    // known at compile time.
+    if (s.length() > width) {
+        throw error("the length of '{}' is greater than the width '{}', this is a bug in chemfiles", s, width);
+    }
+
+    auto f = s[0];
+    if (f == '-' || f == ' ' || std::isdigit(f)) {
```
Luthaf, line 307: I try to not use `std::isdigit` or any functions from `<cctype>` or `<cclocale>` since they use a locale which might not correspond to the way we want to parse the values. 

There is a static `is_digit` function on the top of this file that you can use instead, and you can add `is_ascii_lower/is_ascii_upper` in the same way

src/parse.cpp
```diff
@@ -216,3 +216,127 @@ template <> double chemfiles::parse(string_view input) {
...
+    uint64_t result = 0;
+    auto n = digits_upper.length();
+    for (auto c : s) {
+        result *= n;
+        result += digit_to_value(c);
+    }
+    return result;
+}
+
+static int64_t pow_int(uint64_t base, uint64_t power) {
```
Luthaf, line 263: Not sure I understand the difference between this function and just `std::pow`, could you add a short comment?

src/formats/GRO.cpp
```diff
@@ -64,9 +64,9 @@ void GROFormat::read_next(Frame& frame) {
             );
         }
 
-        size_t resid = SIZE_MAX;
+        int64_t resid = std::numeric_limits<int64_t>::max();
```
Luthaf, line 67: Side note, but this looks like a good use case for `optional<int64_t>`. We should refactor at some point =)

src/formats/GRO.cpp
```diff
@@ -185,7 +185,7 @@ void GROFormat::write_next(const Frame& frame) {
 
         if (residue && residue->id()) {
             auto value = residue->id().value();
-            if (value <= 99999) {
+            if (value <= 99999 && value >= 0) {
```
Luthaf, line 188: Does the GRO format not support negative id? If so, I would emit a warning if we find one of these.

src/formats/PDB.cpp
```diff
@@ -625,18 +628,14 @@ Record get_record(string_view line) {
...
-            warning("PDB writer", "too many atoms, removing atomic id bigger than 100000");
-        }
-        return "*****";
-    } else {
-        return std::to_string(i + 1);
+static std::string to_pdb_index(int64_t value, uint64_t width) {
+    if (value == 10000 + 2 * 26 * static_cast<int64_t>(std::pow(36, width - 1) + 0.5)) {
+        auto type = width == 5 ?
+            "atom" : "residue";
+        warning("PDB writer", "the value for a {} serial/id is too large, using '*' instead", type);
```
Luthaf, line 634: Is this missing a `return "****";`?

src/formats/PDB.cpp
```diff
@@ -813,21 +804,13 @@ void PDBFormat::write_next(const Frame& frame) {
         if (is_atom_record[bond[0]] && is_atom_record[bond[1]]) { // both must be standard residue atoms
             continue;
         }
-        if (bond[0] > 99999 || bond[1] > 99999) {
+        if (bond[0] > 87440031 || bond[1] > 87440031) {
```
Luthaf, line 805: Instead of a magic number, how about making this a `const` or `constexpr` value? Like `constexpr size_t MAX_HYBRID36_NUMBER = 87440031;`?

tests/formats/pdb.cpp
```diff
@@ -329,7 +329,7 @@ TEST_CASE("Problematic PDB files") {
         REQUIRE(file.nsteps() == 156);
     }
 }
-
+#include <iostream>
```
Luthaf, line 332: Leftover debug code?

tests/formats/pdb.cpp
```diff
@@ -486,27 +491,42 @@ TEST_CASE("PDB files with big values") {
...
+        auto residue2 = frame.topology().residue_for_atom(1);
+        CHECK(*(residue2->id()) == 2436111);
+
+        auto residue3 = frame.topology().residue_for_atom(2);
+        CHECK(residue3 == nullopt);
     }
+    
+    /*
+
+    This test is too expensive with the limits of hybrid36...
```
Luthaf, line 510: Yeah, let's remove it =)

src/formats/PDB.cpp
frodofine: No, but the code is unclear - I'll refactor.

src/formats/GRO.cpp
frodofine: It probably should be addressed elsewhere, it's a really easy change!

src/formats/GRO.cpp
frodofine: It's rather unclear. I cannot find any examples where the ID is negative and GROMACS treats residues as signed integers. Here is a case where it is expected for the residue ID to be an increasing whole number. The spec is rather unclear:

http://manual.gromacs.org/documentation/2018/user-guide/file-formats.html#gro

PR #351, opened 2020-07-09
Fix the use of system libraries when they don't live in /usr

Luthaf:
System library refer to zlib/bzip/lzma/netcdf which can be taken directly from the system instead of using the chemfiles' bundled version.

Also remove the dependency on Boost, since it is more of a headache than something helping. In particular if a system library is located in `/usr` as well as a different Boost version from the bundled one, the compiler picked include from /usr/include and the build failed at link time.


Luthaf:
thanks for the review =)

PR #352, opened 2020-07-20
fix(file): update position of cursor when writing

ezavod:
Currently one is able to write multiple frames to CSSR files. Chemfiles is able to read back only one and I guess the standard allows only one frame anyway. This is because the checks rely on the cursor position which is currently not updated when writing. So one can write multiple frames without triggering those checks.
Technically this also applies for appending, but this would require an implementation of `tell` for all different files. Since CSSR does not support appending (and it would not be sensible to support appending for any other one-frame format) we can safely ignore appending for now.

### DIFF
--- a/tests/formats/cssr.cpp
+++ b/tests/formats/cssr.cpp
@@ -90,6 +90,9 @@ R"( REFERENCE STRUCTURE = 00000   A,B,C =  10.000  10.000  12.000
     frame.set_cell(UnitCell(10, 10, 12));
     auto t = Trajectory(tmpfile, 'w');
     t.write(frame);
+
+    CHECK_THROWS_WITH(t.write(frame), "CSSR format only supports writing one frame");
+
     t.close();
 
     std::ifstream checking(tmpfile);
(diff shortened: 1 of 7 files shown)


Luthaf:
Good catch, I guess I broke this in #271 =/

tests/formats/cssr.cpp
```diff
@@ -90,6 +90,10 @@ R"( REFERENCE STRUCTURE = 00000   A,B,C =  10.000  10.000  12.000
     frame.set_cell(UnitCell(10, 10, 12));
     auto t = Trajectory(tmpfile, 'w');
     t.write(frame);
+
+    // Format only supports writing one frame
+    CHECK_THROWS(t.write(frame));
```
Luthaf, line 95: Maybe use `CHECK_THROWS_WITH` instead of `CHECK_THROW` to ensure the expected exception is being thrown?

ezavod: Good idea.

PR #353, opened 2020-07-21
Implement Amber Restart Format

### LINKED ISSUE #47: Support restart convention for NetCDF
This is roughly the same as standard NetCDF, but using doubles most of the time instead of floats.

ezavod:
This enables chemfiles to read and write Amber Restart (.ncrst) files. Those files are pretty similar to Amber Trajectory files but use mostly doubles and contain only a single frame.
Closes #47.

Additional:
- ~fix link to specs of .gro files~
- added check in LAMMPS Data tests for writing multiple frames (see #352)
- read and write frame title property of Amber formats

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Changes in supported formats
 
+- Added read and write support for Amber Restart (.ncrst) files.
+
 ### Changes to the C API
 
 ## 0.10.0 (14 Feb 2021)
--- a/include/chemfiles/files/NcFile.hpp
+++ b/include/chemfiles/files/NcFile.hpp
@@ -134,6 +134,8 @@ public:
         return file_id_;
     }
 
+    /// Check if a global string attribut exists
+    bool global_attribute_exists(const std::string& name) const;
     /// Get a global string attribute from the file
     std::string global_attribute(const std::string& name) const;
     /// Create a global attribute in the file_
--- a/include/chemfiles/files/NcFile.hpp
+++ b/include/chemfiles/files/NcFile.hpp
@@ -73,7 +73,16 @@ namespace nc {
         /// Get `count` values starting at `start` from this variable
         std::vector<float> get(count_t start, count_t count) const;
         /// Add `cout` values from `data` starting at `start` in this variable
-        void add(count_t start, count_t count, std::vector<float> data);
+        void add(count_t start, count_t count, const std::vector<float>& data);
+    };
+
+    class NcDouble final: public NcVariable {
+    public:
+        NcDouble(NcFile& file, netcdf_id_t var) : NcVariable(file, var) {}
+        /// Get `count` values starting at `start` from this variable
+        std::vector<double> get(count_t start, count_t count) const;
+        /// Add `cout` values from `data` starting at `start` in this variable
+        void add(count_t start, count_t count, const std::vector<double>& data);
     };
 
     class NcChar final: public NcVariable {
@@ -87,6 +96,7 @@ namespace nc {
 
     template<typename NcType> struct nc_type;
     template<> struct nc_type<NcFloat> {static constexpr auto value = NC_FLOAT;};
+    template<> struct nc_type<NcDouble> {static constexpr auto value = NC_DOUBLE;};
     template<> struct nc_type<NcChar> {static constexpr auto value = NC_CHAR;};
 } // namespace nc
 
@@ -96,7 +106,7 @@ namespace nc {
 /// format. All the operation are guaranteed to return a valid value or throw an
 /// error.
 ///
-/// The template functions are manually specialized for float and char data types.
+/// The template functions are manually specialized for float, double and char data types.
 class NcFile final: public File {
 public:
     NcFile(std::string path, File::Mode mode);
--- a/include/chemfiles/formats/AmberNetCDF.hpp
+++ b/include/chemfiles/formats/AmberNetCDF.hpp
@@ -21,10 +21,16 @@ class FormatMetadata;
 
 template <class T> class span;
 
+enum AmberFormat {
+    AMBER_NC_RESTART,     ///< AMBER Restart
+    AMBER_NC_TRAJECTORY,  ///< AMBER NetCDF Trajectory
+};
+
 /// Amber NetCDF file format reader.
-class AmberNetCDFFormat final: public Format {
+template <AmberFormat F>
+class Amber final: public Format {
 public:
-    AmberNetCDFFormat(std::string path, File::Mode mode, File::Compression compression);
+    Amber(std::string path, File::Mode mode, File::Compression compression);
 
     void read_step(size_t step, Frame& frame) override;
     void read(Frame& frame) override;
@@ -32,6 +38,10 @@ public:
 
     size_t nsteps() override;
 private:
+    /// Generate the range of indices [start, stop] for a 3D vector at the current step.
+    std::array<std::vector<size_t>, 2> vec3d_range();
+    /// Generate the range of indices [start, stop] for multiple 3D vectors at the current step.
+    std::array<std::vector<size_t>, 2> vec3d_n_range(size_t n);
     /// Read the unit cell at the current internal step, the file is assumed to
     /// be valid.
     UnitCell read_cell();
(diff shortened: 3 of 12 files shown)


include/chemfiles/files/NcFile.hpp
```diff
@@ -76,6 +76,15 @@ namespace nc {
         void add(count_t start, count_t count, std::vector<float> data);
     };
 
+    class NcDouble final: public NcVariable {
```
Luthaf, line 79: Could you add some tests for this in `tests/files/netcdf-file.cpp`?

ezavod: Oh, I missed that there were some tests in place for variable types. Added tests for `NcDouble`.

ezavod:
> do you think we could merge both in a single class?

Yeah, you're right about that. Simply duplicating existing code was the easiest way, but I'll look into how it's done for Molfile. This might take some time because I have to get familiar with C++ templates first :D

> is the '.ncrst' extension standard?

Ambers [cpptraj seems to use this](https://github.com/Amber-MD/cpptraj/blob/dce68a30dcd22449d3e211278d1b9ebd15349a05/src/TrajectoryFile.cpp#L87-L88) as standard and [this websites](https://amberhub.chpc.utah.edu/cpptraj/trajectory-file-commands/) says the same.

ezavod:
Just had some time to rework this before it clocks a full year. This gets rid of duplicate code by using templates. I don't fully grasp every detail of C++ templates, so some parts might fell a little unidiomatic.

src/formats/AmberNetCDF.cpp
```diff
@@ -114,12 +126,45 @@ void AmberNetCDFFormat::read_step(const size_t step, Frame& frame) {
...
+    if (FormatSpec<F>::one_frame && step_ != 0) {
+        throw format_error("AMBER Restart format only supports reading one frame");
+    }
+
     read_step(step_, frame);
     step_++;
 }
 
-UnitCell AmberNetCDFFormat::read_cell() {
+namespace chemfiles {
```
ezavod, line 139: This prevents some error/bug(?) in gcc 4.8.4. Not sure if this is an ideal solution.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 - properties are now sorted, and iterating over properties will always yield
   them sorted by the associated key.
+- Added read and write support for Amber Restart (.ncrst) files.
```
Luthaf, line 17: this should be in the section below "changes in supported formats"

include/chemfiles/files/NcFile.hpp
```diff
@@ -76,6 +76,15 @@ namespace nc {
...
+    public:
+        NcDouble(NcFile& file, netcdf_id_t var) : NcVariable(file, var) {}
+        /// Get `count` values starting at `start` from this variable
+        std::vector<double> get(count_t start, count_t count) const;
+        /// Add `cout` values from `data` starting at `start` in this variable
+        void add(count_t start, count_t count, std::vector<double> data);
```
Luthaf, line 85: Should this use:
```suggestion
        void add(count_t start, count_t count, const std::vector<double>& data);
```
instead? There is no need to take the vector by value. The same for NcFloat.

include/chemfiles/formats/AmberNetCDF.hpp
```diff
@@ -21,17 +21,27 @@ class FormatMetadata;
 
 template <class T> class span;
 
+enum AmberFormat {
+    NCRST,                ///< AMBER Restart
+    NCTRJ,                ///< AMBER NetCDF Trajectory
```
Luthaf, line 26:
```suggestion
    AMBER_NC_RESTART,                       ///< AMBER Restart
    AMBER_NC_TRAJECTORY,                ///< AMBER NetCDF Trajectory
```

tests/files/netcdf-file.cpp
```diff
@@ -9,37 +9,72 @@ using namespace chemfiles;
...
+        CHECK(std::abs(positions[2] - 11.73717) < EPS);
+    }
 }
 
 TEST_CASE("Errors in NetCDF files") {
-    NcFile file("data/netcdf/water.nc", File::READ);
+    SECTION("Float variables") {
+        NcFile file("data/netcdf/water.nc", File::READ);
+
+        CHECK_THROWS_AS(file.global_attribute("FOO"), FileError);
```
Luthaf, line 64: could you change these to `CHECK_THROW_WITH`? It helps making sure the exception thrown is the one intended

src/formats/AmberNetCDF.cpp
Luthaf: no, that's the way to do it. The rules of namespaces & template specialization are a bit complex and changed for different standard version.

src/formats/AmberNetCDF.cpp
```diff
@@ -114,12 +126,45 @@ void AmberNetCDFFormat::read_step(const size_t step, Frame& frame) {
     }
 }
 
-void AmberNetCDFFormat::read(Frame& frame) {
+template <AmberFormat F>
+void Amber<F>::read(Frame& frame) {
+    if (FormatSpec<F>::one_frame && step_ != 0) {
```
Luthaf, line 131: Instead of having the `one_frame` member, you could do something like 
```suggestion
    if (F ==  NCRST && step_ != 0) {
```

src/formats/AmberNetCDF.cpp
```diff
@@ -97,15 +98,26 @@ AmberNetCDFFormat::AmberNetCDFFormat(std::string path, File::Mode mode, File::Co
...
+}
+
+template <>
+size_t Amber<NCTRJ>::nsteps() {
     return file_.dimension("frame");
 }
 
-void AmberNetCDFFormat::read_step(const size_t step, Frame& frame) {
+template <AmberFormat F>
+void Amber<F>::read_step(const size_t step, Frame& frame) {
```
Luthaf, line 112: this should check for `step != 0 && F == NCRST` as well, since the user could call this function directly.

src/formats/AmberNetCDF.cpp
```diff
@@ -221,19 +269,51 @@ static void initialize(NcFile& file, size_t natoms, bool with_velocities) {
...
 
     spatial.add("xyz");
     cell_spatial.add("abc");
     cell_angular.add({"alpha", "beta", "gamma"});
 }
 
-void AmberNetCDFFormat::write(const Frame& frame) {
+template <AmberFormat F>
+void Amber<F>::write(const Frame& frame) {
+    if (FormatSpec<F>::one_frame && step_ != 0) {
```
Luthaf, line 307:
```suggestion
    if (F == NCRST && step_ != 0) {
```

src/formats/AmberNetCDF.cpp
```diff
@@ -246,38 +326,83 @@ void AmberNetCDFFormat::write(const Frame& frame) {
...
+template <AmberFormat F>
+void Amber<F>::write_array(const std::vector<Vector3D>& array, const std::string& name) {
+    auto var = get_variable<F>(file_, name);
     auto natoms = array.size();
-    std::vector<size_t> start{step_, 0, 0};
-    std::vector<size_t> count{1, natoms, 3};
+    auto range = vec3d_n_range(natoms);
 
-    auto data = std::vector<float>(natoms * 3);
+    auto data = std::vector<typename FormatSpec<F>::type>(natoms * 3);
```
Luthaf, line 335: This is alright as-is, you could also add `using real_t = typename FormatSpec<F>::type` at the begining of the function and then use `real_t`  everywhere

Luthaf (commented):
Looks very good, I have a couple of minor improvements suggestions

CHANGELOG.md
ezavod: Oh, I missed that section.

include/chemfiles/files/NcFile.hpp
ezavod: Makes sense. Done.

include/chemfiles/formats/AmberNetCDF.hpp
ezavod: Done.

src/formats/AmberNetCDF.cpp
ezavod: That's too easy :-)

src/formats/AmberNetCDF.cpp
ezavod: The check should have been here in the first place. Moved to this position now.

src/formats/AmberNetCDF.cpp
ezavod: That's much easier to the eye. Done.

tests/files/netcdf-file.cpp
ezavod: Done.

Luthaf (approved):
Thanks a lot for doing this! I'll merge once CI is happy

PR #355, opened 2020-07-22
Fix spelling of “orthorhombic” Fixes #354

### LINKED ISSUE #354: “orthorhombic” takes two h's
See for example [wikipedia](https://en.wikipedia.org/wiki/Orthorhombic_crystal_system).
It should be “orthorhombic” not “orthorhombic” in several places, in particular in `UnitCell.cpp`: https://github.com/chemfiles/chemfiles/blob/7d744d67e8bc7245613534a2a303aac9d836eb5e/src/UnitCell.cpp#L252

fxcoudert:
Fixes https://github.com/chemfiles/chemfiles/issues/354

### DIFF
--- a/include/chemfiles/UnitCell.hpp
+++ b/include/chemfiles/UnitCell.hpp
@@ -57,7 +57,7 @@ public:
     /// @example{cell/cell-1.cpp}
     UnitCell(double a);
 
-    /// Construct an `ORTHOROMBIC` unit cell of side size `a`, `b`, `c`
+    /// Construct an `ORTHORHOMBIC` unit cell of side size `a`, `b`, `c`
     ///
     /// @example{cell/cell-3.cpp}
     UnitCell(double a, double b, double c);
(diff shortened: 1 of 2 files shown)


Luthaf:
By the way, since you may know that: do you have any idea why GCC 5.5 is so slow on OS X 10.13? (except because we are using old releases of everything ^^) It takes 30min to run the build, compared to 7min for clang on OS X or ~8min for gcc on Ubuntu 16.04. 

Could it be related to the flood of `section "__textcoal_nt" is deprecated` warnings we get from the linker/assembler?

Luthaf:
Wait it looks like we spend 15min updating homebrew and installing a bunch of dependencies such as qt? I'll look into this, and potentially update the CI to something more recent.

Luthaf:
Travis is happy at https://travis-ci.org/github/chemfiles/chemfiles/builds/710787958, looks like the github UI did not update.

fxcoudert:
> why GCC 5.5 is so slow on OS X 10.13?

There is no reason it should. But seriously, don't use it.

Luthaf:
I'm trying to stay conservative regarding which compiler version are supported (we still support GCC 4.8.5 since that's the default in CentOS 7, which will EOL in 2024), but we could update on macOS since way less people are using ancient GCC there. I'll try to look up version statistics.

PR #357, opened 2020-07-22
Speedup OS X gcc CI build by not re-installing everything

Luthaf:
It is much slower than other builders for an unknown reason


Luthaf:
Part of the reason was that a lot of dependencies were re-installed, using `HOMEBREW_NO_INSTALL_CLEANUP=1`
and `HOMEBREW_NO_AUTO_UPDATE=1` remove 17min from the build!

PR #360, opened 2020-07-27
Set higher precedence for 'not' than 'and'/'or' in selections

### LINKED ISSUE #349: not should have higher precendence than or/and in selections
From https://github.com/chemfiles/Chemfiles.jl/issues/37#issuecomment-647035920, by @amlimaye

`"not type == "16" or type == "17"` seems to be parsed as `"not (type == "16" or type == "17")`, while intuitively I would think it parses as `"(not type == "16") or (type == "17")`.

The current behavior means that `"not type == "16" or type == "17"` behaves differently from `"type == "16" or not type == "17"`, i.e. both sides of `or` are not equivalent. I believe the same is true of `and`, although I did not check.

What do you think, should we change the parsing order here?

Luthaf:
Fix #349

### DIFF
--- a/src/selections/parser.cpp
+++ b/src/selections/parser.cpp
@@ -246,7 +246,7 @@ Ast Parser::selector() {
         }
 
     } else if (match(Token::NOT)) {
-        auto ast = expression();
+        auto ast = selector();
         return chemfiles::make_unique<Not>(std::move(ast));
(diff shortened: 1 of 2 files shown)


Luthaf:
I'll merge this tomorrow if nobody opposes it!

PR #368, opened 2020-09-11
Change the representation of the unit cells

### LINKED ISSUE #69: Throw errors on negative cell lengths
The [`UnitCell::set_{a,b,c}`](https://github.com/chemfiles/chemfiles/blob/0aaa53d3ecaf07e8cd2a8a1dd0849f4690cdb928/src/UnitCell.cpp#L113) function should throw an error if the length is 0 (except for infinite cells) or negative.

On the same line, the [`UnitCell::set_{alpha,beta,gamma}`](https://github.com/chemfiles/chemfiles/blob/0aaa53d3ecaf07e8cd2a8a1dd0849f4690cdb928/src/UnitCell.cpp#L137) function should throw an error if the angles are not between 0 and 180.

Luthaf:
Instead of using lengths + angles, the full cell matrix is always used. This ensure that we can represent rotated unit cells without loosing data.

The previous implementation was loosing some data in NPT simulations where the unit cell was allowed to fluctuate. While it is always possible to represent cells as lengths + angles, doing so often requires rotating the cell in 3D to align it with the xyz axis; also changing positions. It is easier to always use a full cell matrix.

Fixes #69


src/UnitCell.cpp
```diff
@@ -9,261 +9,245 @@
...
-        a_ = b_ = c_ = 0;
-        alpha_ = beta_ = gamma_ = 90.0;
+static Matrix3D cell_matrix_from_lenths_angles(Vector3D lengths, Vector3D angles) {
+    if (lengths[0] < 0 || lengths[1] < 0 || lengths[2] < 0) {
+        throw error("can not create a unit cell with negative lengths");
+    }
 
-        update_matrix();
+    if (lengths != Vector3D(0, 0, 0) && (is_roughly_zero(lengths[0]) || is_roughly_zero(lengths[1]) || is_roughly_zero(lengths[2]))) {
+        warning("", "trying to create a unit cell with only one or two lengths set to zero");
```
Luthaf, line 61: @ezavod: this is a warning instead of an error in part because of the `scaled_traj.nc` test file introduced in https://github.com/chemfiles/tests-data/commit/e1c6a19ad774a4fc7c3e55e59c30834bdcf7066e, which have cell lengths of `60.9682, 60.9682, 0`. Is this by design?

ezavod: I needed scaling for some very antiquated script a while ago. Seems to be a problem with that script that I didn't thought about at that time because I replaced the box dimensions for subsequent simulation.
Should I update the test files to some reasonable value so that one can have an error instead?

Luthaf: I am not sure we want this to be an error: on one hand such cells do not make much sense, on the other hand they might be used to indicate partially periodic systems (for scaled_traj.nc, the system would only be periodic along x and y). Do you have a stronger opinion?

Luthaf:
@ezavod, do you think you will have time to review this? If not I can do it in some other way =)

ezavod (commented):
This looks really good and makes the tedious conversion in XTC/TRR files a lot cleaner. I just have some minor feedback.

src/UnitCell.cpp
```diff
@@ -9,261 +9,245 @@
...
-void UnitCell::set_gamma(double val) {
+void UnitCell::set_angles(Vector3D angles) {
     if (shape_ != TRICLINIC) {
-        throw error("can not set 'gamma' on non triclinic cell");
+        throw error("can not set angles for a non-triclinic cell");
+    }
+
+    if (angles[0] <= 0 || angles[1] <= 0 || angles[2] <= 0 ||
+        is_roughly_zero(angles[0]) || is_roughly_zero(angles[1]) || is_roughly_zero(angles[2])) {
+        throw error("can not set unit cell angles to negative values");
```
ezavod, line 225: The error message is somewhat inconsistent: [in the constructor](https://github.com/chemfiles/chemfiles/blob/ca56e4d2a922e05558c25132509977c804c4300a/src/UnitCell.cpp#L64-L70) values of zero and negative values are differentiated. Would it be better to define a (inlined) helper function something like `check_angles` that could be used everywhere to throw consistent errors? Same for lengths.

ezavod (commented):
Currently setting angles > 180° will result in an error: "invalid unit cell matrix with negative determinant". It might be a little bit confusing for novice users to identify the root cause.
Should one add an additional check for nonphysical angles >= 180°? This would then also fix #69 =)

include/chemfiles/capi/cell.h
```diff
@@ -21,37 +21,38 @@ typedef enum {  // NOLINT: this is both a C and C++ file
...
+/// Else a `TRICLINIC` cell is created. The matrix entries should be in
+/// Angstroms.
 ///
 /// The caller of this function should free the associated memory using
 /// `chfl_free`.
 ///
-/// @example{capi/chfl_cell/triclinic.c}
-/// @return A pointer to the unit cell, or NULL in case of error. You can use
+/// @example{capi/chfl_cell/from_matrix.c} 
+/// @return A pointer to the unit cell, or NULL in case of error. You can use
```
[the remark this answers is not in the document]
Luthaf: That's an issue with me changing my editor configuration to contribute to another project, thanks for catching it!

src/UnitCell.cpp
Luthaf: Sounds good!

src/formats/GRO.cpp
```diff
@@ -35,6 +35,16 @@ template<> FormatInfo chemfiles::format_information<GROFormat>() {
...
 
+static bool is_roughly_zero(double value) {
+    // We think that 0.00001 is close enough to 0
+    return fabs(value) < 1e-5;
+}
+
+static bool is_upper_triangular(const Matrix3D& matrix) {
+    return is_roughly_zero(matrix[1][0]) && is_roughly_zero(matrix[2][0]) &&
+           is_roughly_zero(matrix[2][1]);
+}
```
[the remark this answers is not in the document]
Luthaf: I agree, my initial justification for duplicating was that `UnitCell.hpp` is a public header and that's not something we want to expose to chemfiles users. I might move them to `chemfiles::__private_implementation_detail::is_upper_triangular` or something =)

src/UnitCell.cpp
Luthaf: Ok, let's keep it like this =)

src/UnitCell.cpp
```diff
@@ -9,261 +9,254 @@
...
-        b_ = matrix[1][1];
-        c_ = matrix[2][2];
+    if (is_roughly_zero(angles[0]) || is_roughly_zero(angles[1]) || is_roughly_zero(angles[2])) {
+        throw error("a unit cell can not have 0° angles");
+    }
 
-        alpha_ = beta_ = gamma_ = 90.0;
+    if (angles[0] > 180 || angles[1] > 180 || angles[2] > 180) {
+        throw error("a unit cell can not have angles larger than 180°");
+    }
```
ezavod, line 80: This would allow nonphysical angles of exactly 180°. Would `>=` be better? I'm pretty indifferent about that, so fell free to decide however you want.

ezavod (approved):
Looks good to me.

src/UnitCell.cpp
Luthaf: Sure, that might help a bit =)

Luthaf:
Fixes #69

src/UnitCell.cpp
ezavod: One very last nitpick: it would probably be nice to have the error say "... larger than or equal to 180°".
But then this is good to go :D

PR #373, opened 2020-09-21
Clarify static lib and header installation from binaries

jmintser:
Copy/pasting from Gitter


Luthaf:
Thanks for doing this! I've updated the syntax to follow restructured text, I'll merge when CI is happy =)

Issue #280, opened 2019-11-05
Extend formats documentations


Luthaf:
Currently, the supported formats all have a link pointing to an external definition of the format. These external links can unfortunately go away, so they should be replaced by a documentation of the format inside chemfiles docs.

Having our own documentation would also allow to document how property data is read/written by a given format, as well as which extension to the format (if any) are supported.

Luthaf:
Somehow related, we could add more properties to [`FormatInfo`](https://github.com/chemfiles/chemfiles/blob/b63b14714106437b097e7b085eebd312732903c4/include/chemfiles/Format.hpp#L80), such as
- does this format support reading?
- does this format support writing?
- does this format support storing more than one frame?
- can this format store:
   - positions?
   - velocities?
   - unit cell?
   - atomic names?
   - bonding information?
   - residues information?
   - ...
- can different steps have different topology?

---

This information could be made available as part of the (C) API for external users of chemfiles to query it.

The documentation could also extract it as necessary to generate the big table in http://chemfiles.org/chemfiles/latest/formats.html

Luthaf:
Done in #377

PR #377, opened 2020-10-03
Add more format metadata, and make the formats list accessible

Luthaf:
The goal of these changes is two fold:

- make the list of formats and associated metadata available to bindings and all users
- reduce the duplication of information by using the format metadata to generate part of the documentation

Two of the drawbacks regarding the second part are that it is no longer possible to generate only the documentation, without building the code; and that building the documentation is not possible when cross-compiling. I feel like they are minor drawbacks, since users are unlikely to build the documentation when compiling chemfiles for use inside their own project.


src/FormatMetadata.cpp
```diff
@@ -0,0 +1,57 @@
+// Chemfiles, a modern library for chemistry file reading and writing
+// Copyright (C) Guillaume Fraux and contributors -- BSD license
+
+#include "chemfiles/FormatMetadata.hpp"
+
+#include "chemfiles/utils.hpp"
+#include "chemfiles/error_fmt.hpp"
+using namespace chemfiles;
+
+static void check_not_empty(const char* value, string_view field, const char* name) {
```
ezavod, line 10: At first glance the parameter names are rather non-descriptive and I got confused. Do you think `field_value`, `field_name`, and `format_name` are better suited?

src/FormatMetadata.cpp
```diff
@@ -0,0 +1,57 @@
...
+    if (std::string(value).empty()) {
+        auto for_name = name == nullptr || field == "name" ? "" : fmt::format(" for '{}'", name);
+        throw format_error("the format {} can not be an empty string{}", field, for_name);
```
ezavod, line 13: I fell the possessive form "format's" would be clearer, because currently one might misunderstand and expect the name of the format e.g. "the format XYZ ..." when skimming the error messages

ezavod (approved):
LGTM.
In the future it would be possible to employ a script to test if the external links are still alive as they regularly go away (see #280).

Luthaf:
Agreed on the script to check for dead links! I'll if I can add it to this PR. I think there are sphinx extensions that can do this.

src/FormatMetadata.cpp
Luthaf: I went with `the XXX can not be YYY for format ZZZ`, do you find it better?

ezavod (approved):
Awesome that sphinx is able to do this natively!
It is unfortunate that the Molden format specs seem to have perished. Even Wikipedia has a dead link.

PR #380, opened 2020-10-20
Miscelaneous cleanup before the release

Luthaf:


Luthaf:
@ezavod and @frodofine, does one of you have some time in the upcoming week to review this?

frodofine:
I can take a look tonight.

src/files/NcFile.cpp
```diff
@@ -147,6 +147,8 @@ NcFile::NcFile(std::string path, File::Mode mode)
 NcFile::~NcFile() {
     auto status = nc_close(file_id_);
     assert(status == NC_NOERR);
+    // silent "unused variable" when compiling without assertions
```
frodofine, line 150: Would it be better to combine the above statement to `nc_close(file_id_) == NC_NOERR`?

src/files/XzFile.cpp
```diff
@@ -140,6 +141,8 @@ void XzFile::seek(uint64_t position) {
 
     auto count = this->read(buffer, static_cast<size_t>(position));
     assert(count == position);
+    // silent "unused variable" when compiling without assertions
+    (void)count;
```
frodofine, line 145: Maybe it would be a good idea to add a macro like `CHFL_UNUSED` which expands to this?

src/formats/PDB.cpp
```diff
@@ -807,7 +808,8 @@ void PDBFormat::write_next(const Frame& frame) {
     }
 
     // Used to skip writing unnecessary connect records
-    auto is_atom_record = std::vector<bool>(frame.size(), false);
+    // use std::deque because std::vector<bool> is completly broken
```
frodofine, line 811: How so?

tests/lints/check-used-functions.py
```diff
@@ -35,6 +35,15 @@
     r"(?P<name>\b(std)?(::)?strtod\b)": "parse<double>",
     r"(?P<name>\b(std)?(::)?strtold\b)": "parse<double>",
     r"(?P<name>\b(std)?(::)?sscanf\b)": "scan",
+    r"(?P<name>\b(std)?(::)?isalnum\b)": "is_ascii_TODO",
```
frodofine, line 38: Is this used?

frodofine (commented):
Looks good, minor comments/questions. One thing I noticed is that a lot of this PR is a result of header file changes. Would it be a good idea to include a check in Travis to avoid a major commit like this again?

include/chemfiles/selections/expr.hpp
```diff
@@ -4,15 +4,16 @@
...
-#include <cstring>
+
 #include <string>
 #include <vector>
 #include <memory>
-#include <cassert>
 #include <functional>
 
 #include "chemfiles/external/optional.hpp"
+#include "chemfiles/selections/NumericValues.hpp"
```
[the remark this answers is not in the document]
Luthaf: Yes, I feel like it makes more sense since this is a fundamental data structure used in multiple expressions

include/chemfiles/utils.hpp
```diff
@@ -42,22 +41,58 @@ inline std::vector<string_view> split(const char* string, char delim) {
 // disallow temporary string
 std::vector<string_view> split(std::string&& string, char delim) = delete;
 
-// chemfiles' definition of whitespace only includes ' ', '\t' and new lines
-inline bool is_whitespace(char c) {
-    return c == ' ' || c == '\t' || c == '\r' || c == '\n';
+// Check whether the given character is an ASCII whitespace
```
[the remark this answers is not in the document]
Luthaf: Yes, and make it clearer with the name that we only support ASCII

src/files/Bz2File.cpp
```diff
@@ -10,11 +10,21 @@
 #include <vector>
 #include <functional>
 
+#include "chemfiles/config.h"
+
+// bzlib.h includes windows.h on Windows platforms, which defines `min` and
+// `max` symbols, failing compilation below.
+#ifdef CHEMFILES_WINDOWS
+#define WIN32_LEAN_AND_MEAN
+#define NOMINMAX
```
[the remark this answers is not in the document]
Luthaf: Yeah, that make sense. I'll change it!

src/files/Bz2File.cpp
```diff
@@ -143,6 +153,8 @@ void Bz2File::seek(uint64_t position) {
 
     auto count = this->read(buffer, static_cast<size_t>(position));
     assert(count == position);
+    // silent "unused variable" when compiling without assertions
+    (void)count;
```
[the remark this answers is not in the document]
Luthaf: yes, without this in release mode the assertion is removed and the compiler thinks `count` is unused. Casting to void is the usual trick to get around this.

src/files/NcFile.cpp
Luthaf: Inside the assert? No, since asserts should not have side effects. Or are you suggesting something else?

src/formats/PDB.cpp
Luthaf: std::vector<bool> is required by the standard to use bit packing, and this has non negligible performance issues: https://stackoverflow.com/a/36933356/4692076, https://stackoverflow.com/q/17794569/4692076.

"completly broken" might be a bit strong here, I'll rephrase it =)

tests/lints/check-used-functions.py
Luthaf: whoops, good catch. This is not yet used, I wanted to add it for completeness. I'll change this!

Luthaf:
>  One thing I noticed is that a lot of this PR is a result of header file changes. Would it be a good idea to include a check in Travis to avoid a major commit like this again?

I think you are mentioning the "run include-what-you-use" commit? I'm a bit hesitant to add it to Travis because it adds another tool that potential contributors must install and learn how to use, plus it makes the compilation very slow. 

Until now, I've treated it in a similar way as clang-tidy, i.e. something I manually run every so often, fixing the corresponding warnings as I go.

For now, I've found the balance between additional work on my side vs slower CI and higher barrier to contribute to be against adding it as an additional check. I do agree that is can result in large-ish commits that don't play too nicely with git blame.

include/chemfiles/capi/shared_allocator.hpp
```diff
@@ -2,9 +2,7 @@
 #define CHFL_SHARED_ALLOCATOR_HPP
 
 #include <vector>
-#include <memory>
 #include <functional>
-#include <algorithm>  // IWYU pragma: keep
```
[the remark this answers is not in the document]
Luthaf: I don't really know, this include was added the first time I ran include-what-you-use on the code, in db19ed8fbdaca209d1ca71b01a51a6386befe4e4. It might have been to workaround a bug with include-what-you-use, I remember it being pretty fiddly at the time.

include/chemfiles/FormatMetadata.hpp
```diff
@@ -67,32 +64,6 @@ class CHFL_EXPORT FormatMetadata {
...
-///     template<> const FormatMetadata& format_metadata<MyFormat>() {
-///         static FormatMetadata metadata;
-///         metadata.name = "MyFormat";
-///         metadata.extension = ".mtf";
-///         return metadata;
-///     }
-/// }
-/// ```
-template<class Format>
-const FormatMetadata& format_metadata() {
```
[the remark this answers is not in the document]
Luthaf: The idea was to make it so `FormatMetadata.hpp` do not have to be included in any of `include/chemfiles/format/*.hpp`. This gains us very little, so I can revert it

src/files/XzFile.cpp
Luthaf: We could, but I don't like having an extra global macro, and there are very few use of this pattern. Do you feel it would be worth it, at least from the code clarity point of view?

Luthaf:
@frodofine do you think you will have some time in the next days to have another look at this?

Luthaf:
@frodofine or @ezavod would one of you have a bit of time this week to give this a final look?

Luthaf:
I'll go on and merge this to start the release process, feel free to add comments, I'll address in a separated PR!

Issue #116, opened 2017-10-19
Speed-up bonds detection with an octree


Luthaf:
The idea comes from this github project: https://github.com/PorridgeBear/imvs/blob/master/IMVS/AtomCloud.swift.

Creating an octree will allow for faster lookup in bonds detection, replacing the brute-force 2D loop here: https://github.com/chemfiles/chemfiles/blob/3cbd6d78bc99ded786044eee6db7022ed2bf4c75/src/Frame.cpp#L60-L80

Instead of looping over all the atoms in the frame, we will only need to loop over the atoms in the surrounding cells from the tree.

richardjgowers:
@Luthaf fwiw, we had a GSOC student last year looking at domain decomposition methods for limited distance searches (ie distances up to a threshold, here bond lengths). The full details are [in his blogposts](https://ayushsuhane.github.io/), but ultimately we found using a rip off of Gromacs' neighbour list algorithm was faster than a kd-tree/octree.  We were using python bindings of these methods, so maybe your mileage may vary!

Luthaf:
Yeah, cells lists (that the blog post mention rather than neighbours lists) might be a good idea here. Thanks for the head up!

PR #381, opened 2020-11-09
Speed-up bonds detection using cell lists

ynaimi:
Implemented a cell list algorithm (inspired by the VMD code) to detect bonds based on interatomic distances (related to issue  #116)

### DIFF
--- a/include/chemfiles/Frame.hpp
+++ b/include/chemfiles/Frame.hpp
@@ -8,394 +8,410 @@
 #include <vector>
 
 #include "chemfiles/exports.h"
-#include "chemfiles/types.hpp"
-#include "chemfiles/external/span.hpp"
 #include "chemfiles/external/optional.hpp"
+#include "chemfiles/external/span.hpp"
+#include "chemfiles/types.hpp"
 
-#include "chemfiles/Topology.hpp"
-#include "chemfiles/UnitCell.hpp"
 #include "chemfiles/Connectivity.hpp"
 #include "chemfiles/Property.hpp"
 #include "chemfiles/Residue.hpp"
+#include "chemfiles/Topology.hpp"
+#include "chemfiles/UnitCell.hpp"
 
 namespace chemfiles {
-class Atom;
+	class Atom;
 
-/// A frame contains data from one simulation step The Frame class holds data
-/// from one step of a simulation: the current topology, the positions, and the
-/// velocities of the particles in the system. If some information is missing
-/// the corresponding data is filled with a default value. Specifically:
-///
-/// @verbatim embed:rst:leading-slashes
-///
-/// * ``cell`` is an infinite unit cell;
-/// * ``topology`` is empty, and contains no data;
-/// * ``positions`` is filled with zeros;
-/// * ``velocities`` is the ``nullopt`` variant of :cpp:class:`chemfiles::optional`.
-///
-/// @endverbatim
-///
-///Iterating over a `Frame` will yield all the atoms in the system.
-///
-/// @example{frame/iterate.cpp}
-class CHFL_EXPORT Frame final {
-public:
-    /// Create an empty frame with no atoms and the given cell.
-    ///
-    /// @example{frame/frame.cpp}
-    explicit Frame(UnitCell cell = UnitCell());
+	/// A frame contains data from one simulation step The Frame class holds data
+	/// from one step of a simulation: the current topology, the positions, and the
+	/// velocities of the particles in the system. If some information is missing
+	/// the corresponding data is filled with a default value. Specifically:
+	///
+	/// @verbatim embed:rst:leading-slashes
+	///
+	/// * ``cell`` is an infinite unit cell;
+	/// * ``topology`` is empty, and contains no data;
+	/// * ``positions`` is filled with zeros;
+	/// * ``velocities`` is the ``nullopt`` variant of :cpp:class:`chemfiles::optional`.
+	///
+	/// @endverbatim
+	///
+	///Iterating over a `Frame` will yield all the atoms in the system.
+	///
+	/// @example{frame/iterate.cpp}
+	class CHFL_EXPORT Frame final {
+	public:
+		/// Create an empty frame with no atoms and the given cell.
+		///
+		/// @example{frame/frame.cpp}
+		explicit Frame(UnitCell cell = UnitCell());
 
-    ~Frame() = default;
-    Frame(Frame&&) = default;
-    Frame& operator=(Frame&&) = default;
+		~Frame() = default;
+		Frame(Frame&&) = default;
+		Frame& operator=(Frame&&) = default;
 
-    /// Get a clone (exact copy) of this frame.
-    ///
-    /// This replace the implicit copy constructor (which is private) to
-    /// make an explicit copy of the frame.
-    ///
-    /// @example{frame/clone.cpp}
-    Frame clone() const {
-        return *this;
-    }
+		/// Get a clone (exact copy) of this frame.
+		///
+		/// This replace the implicit copy constructor (which is private) to
+		/// make an explicit copy of the frame.
+		///
+		/// @example{frame/clone.cpp}
+		Frame clone() const {
+			return *this;
+		}
 
-    /// Get a const reference to the topology of this frame
-    ///
-    /// It is not possible to get a modifiable reference to the topology,
-    /// because it would then be possible to remove/add atoms without changing
-    /// the actual positions and velocity storage. Instead, all the mutating
-    /// functionalities of the topology are mirrored on the frame (adding and
-    /// removing bonds, adding residues, *etc.*)
-    ///
-    /// @example{frame/topology.cpp}
-    const Topology& topology() const {
-        return topology_;
-    }
+		/// Get a const reference to the topology of this frame
+		///
+		/// It is not possible to get a modifiable reference to the topology,
+		/// because it would then be possible to remove/add atoms without changing
+		/// the actual positions and velocity storage. Instead, all the mutating
+		/// functionalities of the topology are mirrored on the frame (adding and
+		/// removing bonds, adding residues, *etc.*)
+		///
+		/// @example{frame/topology.cpp}
+		const Topology& topology() const {
+			return topology_;
+		}
 
-    /// Set the topology of this frame to `topology`
-    ///
-    /// @param topology the new Topology to use for this frame
-    ///
-    /// @throw Error if the topology size does not match the size of this frame
-    ///
-    /// @example{frame/topology.cpp}
-    void set_topology(Topology topology);
+		/// Set the topology of this frame to `topology`
+		///
+		/// @param topology the new Topology to use for this frame
+		///
+		/// @throw Error if the topology size does not match the size of this frame
+		///
+		/// @example{frame/topology.cpp}
+		void set_topology(Topology topology);
 
-    /// Get a const reference to the unit cell of this frame
-    ///
-    /// @example{frame/cell.cpp}
-    const UnitCell& cell() const {
-        return cell_;
-    }
+		/// Get a const reference to the unit cell of this frame
+		///
+		/// @example{frame/cell.cpp}
+		const UnitCell& cell() const {
+			return cell_;
+		}
 
-    /// Get a reference to the unit cell of this frame
-    ///
-    /// @example{frame/cell.cpp}
-    UnitCell& cell() {
-        return cell_;
-    }
+		/// Get a reference to the unit cell of this frame
+		///
+		/// @example{frame/cell.cpp}
+		UnitCell& cell() {
+			return cell_;
+		}
 
-    /// Set the unit cell for this frame to `cell`
-    ///
-    /// @param cell the new UnitCell to use for this frame
-    ///
-    /// @example{frame/cell.cpp}
-    void set_cell(UnitCell cell) {
-        cell_ = std::move(cell);  // NOLINT: std::move for trivially copiable type
-    }
+		/// Set the unit cell for this frame to `cell`
+		///
+		/// @param cell the new UnitCell to use for this frame
+		///
+		/// @example{frame/cell.cpp}
+		void set_cell(UnitCell cell) {
+			cell_ = std::move(cell);  // NOLINT: std::move for trivially copiable type
+		}
 
-    /// Get the number of atoms in this frame
-    ///
-    /// @example{frame/size.cpp}
-    size_t size() const;
+		/// Get the number of atoms in this frame
+		///
+		/// @example{frame/size.cpp}
+		size_t size() const;
 
-    /// Get the positions of the atoms in this frame.
-    ///
-    /// @example{frame/positions.cpp}
-    span<Vector3D> positions() {
-        return positions_;
-    }
+		/// Get the positions of the atoms in this frame.
+		///
+		/// @example{frame/positions.cpp}
+		span<Vector3D> positions() {
+			return positions_;
+		}
 
-    /// Get the positions in this frame as a const reference
-    ///
-    /// @example{frame/positions.cpp}
-    const std::vector<Vector3D>& positions() const {
-        return positions_;
-    }
+		/// Get the positions in this frame as a const reference
+		///
+		/// @example{frame/positions.cpp}
+		const std::vector<Vector3D>& positions() const {
+			return positions_;
+		}
 
-    /// Add velocities data storage to this frame.
-    ///
-    /// If velocities are already defined, this functions does nothing. The new
-    /// velocities are initialized to 0.
-    ///
-    /// @example{frame/add_velocities.cpp}
-    void add_velocities();
+		/// Add velocities data storage to this frame.
+		///
+		/// If velocities are already defined, this functions does nothing. The new
+		/// velocities are initialized to 0.
+		///
+		/// @example{frame/add_velocities.cpp}
+		void add_velocities();
 
-    /// Get an velocities of the atoms in this frame, if this frame contains
-    /// velocity data.
-    ///
-    /// @example{frame/velocities.cpp}
-    optional<span<Vector3D>> velocities() {
-        if (velocities_) {
-            return {*velocities_};
-        } else {
-            return nullopt;
-        }
-    }
+		/// Get an velocities of the atoms in this frame, if this frame contains
+		/// velocity data.
+		///
+		/// @example{frame/velocities.cpp}
+		optional<span<Vector3D>> velocities() {
+			if (velocities_) {
+				return { *velocities_ };
+			}
+			else {
+				return nullopt;
+			}
+		}
 
-    /// Get an velocities of the atoms in this frame as a const reference, if
-    /// this frame contains velocity data.
-    ///
-    /// @example{frame/velocities.cpp}
-    optional<const std::vector<Vector3D>&> velocities() const {
-        if (velocities_) {
-            return {*velocities_};
-        } else {
-            return nullopt;
-        }
-    }
+		/// Get an velocities of the atoms in this frame as a const reference, if
+		/// this frame contains velocity data.
+		///
+		/// @example{frame/velocities.cpp}
+		optional<const std::vector<Vector3D>&> velocities() const {
+			if (velocities_) {
+				return { *velocities_ };
+			}
+			else {
+				return nullopt;
+			}
+		}
 
-    /// Resize the frame to contain `size` atoms.
-    ///
-    /// If the new number of atoms is bigger than the old one, missing data is
-    /// initializd to 0. Pre-existing values are conserved.
-    ///
-    /// If the new size if smaller than the old one, all atoms and connectivity
-    /// elements after the new size are removed.
-    ///
-    /// @example{frame/resize.cpp}
-    void resize(size_t size);
+		/// Resize the frame to contain `size` atoms.
+		///
+		/// If the new number of atoms is bigger than the old one, missing data is
+		/// initializd to 0. Pre-existing values are conserved.
+		///
+		/// If the new size if smaller than the old one, all atoms and connectivity
+		/// elements after the new size are removed.
+		///
+		/// @example{frame/resize.cpp}
+		void resize(size_t size);
 
-    /// Allocate memory in the frame to have enough size for `size` atoms.
-    ///
-    /// This function does not change the actual number of atoms in the frame,
-    /// and should be used as an optimisation.
-    ///
-    /// @example{frame/reserve.cpp}
-    void reserve(size_t size);
+		/// Allocate memory in the frame to have enough size for `size` atoms.
+		///
+		/// This function does not change the actual number of atoms in the frame,
+		/// and should be used as an optimisation.
+		///
+		/// @example{frame/reserve.cpp}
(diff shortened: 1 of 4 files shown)


include/chemfiles/config.h
```diff
@@ -0,0 +1,34 @@
+// Chemfiles, a modern library for chemistry file reading and writing
```
Luthaf, line 1: This file should not be committed, it is generated by cmake as needed. Are you building inside the source directory instead of a separate build folder?

include/chemfiles/exports.h
```diff
@@ -0,0 +1,42 @@
+
```
Luthaf, line 1: This file should not be committed, it is generated by cmake as needed.

src/Frame.cpp
```diff
@@ -1,251 +1,626 @@
...
+			return it->second;
+		}
+	}
+	else {
+		// default to chemfiles provided atom type
+		return atom.vdw_radius();
+	}
+}
+
+int Frame::generateBonds(std::vector<int>* result, std::vector<Vector3D>* positions, int totalCells, int** cells,
```
Luthaf, line 552: Pointers is not the usual tool to reach for in C++, it is more of a C thing to do. For example, it seems that `result` is used as an output parameter, the usual way of doing this in C++ would be to pass a reference: `std::vector<int>& result`. Same for positions, cells and all the pointer parameters.

src/Frame.cpp
```diff
@@ -1,251 +1,626 @@
...
+
+	for (cellIndex = 0; (cellIndex < totalCells) && (bondsCount < maxBonds); cellIndex += 1) {
+
+		const int* tmpCell = cells[cellIndex];
+		int atomsInCell = numInCell[cellIndex];
+
+		if (((cellIndex & 255) == 0)) {
+			char tmpbuf[128];
+			sprintf(tmpbuf, "%6.2f", (100.0f * cellIndex) / (float)totalCells);
+			printf("Generating bonds...: %s %% complete\n", tmpbuf);
```
Luthaf, line 568: Could you remove this debugging code?

src/Frame.cpp
```diff
@@ -1,251 +1,626 @@
...
+				//Skip over self and already-tested atoms
+				int firstNeighborCell = (selfCell) ? i + 1 : 0;
+
+				for (j = firstNeighborCell; (j < atomNumInCell) && (bondsCount < maxBonds); j++) {
+					int atom2 = neighborCell[j];
+					Vector3D posAtom2 = positions->at(atom2);
+					float distanceX = posAtom1[0] - posAtom2[0];
+					float distanceY = posAtom1[1] - posAtom2[1];
+					float distanceZ = posAtom1[2] - posAtom2[2];
+					float interatomicDistance = distanceX * distanceX + distanceY * distanceY + distanceZ * distanceZ;
```
Luthaf, line 595: This does not account for periodic boundary conditions. Why not use `this->distance(atom1, atom2)` here?

Luthaf (changes requested):
Thanks a lot for taking the time of doing this! 

Unfortunately, the code has been reformatted (maybe with an automatic setting in your editor?), which makes the changes hard to review. The formatting also do not correspond to the overall style  used in chemfiles. 

Could you please remove the auto-formatting, at least for the part of the code you did not touch? After that I'll be able to do a more in-depth review =)

include/chemfiles/config.h
ynaimi: Sorry this was a mistake, I shouldn't have committed this. I do have a separate build folder but I was trying (in another project) to include chemfiles' sources in order to debug and test.

include/chemfiles/exports.h
ynaimi: Same here, a mistake I am removing these two files from here in my next commit.

src/Frame.cpp
ynaimi: Sure. I remove it in the next commit.

include/chemfiles/config.h
Luthaf: No worries, we should add these files to the gitignore to ensure this can not happen again =)

src/Frame.cpp
ynaimi: You are right, it clearly doesn't take into account PBC but does bond perception have to take PBC into account? If yes, it means that when one of a two bonded atom crosses the box limit and goes to the other side of the box, the bond will be displayed in the entire width of the box. I don't know how PBC are treated and displayed in chemfiles but I could definitely use Frame's distance function in order to take that into account.

src/Frame.cpp
ynaimi: You are right. Even if I make sure to initialize the result variable before passing it to the function and I know the other variables exist but it is better to pass a reference instead. It will be changed in the next commit

ynaimi:
Hi !

In fact, I think I have this "automatic formatting" enabled in my VS settings. I will try to reverse that in the next commit and incorporate all the changes you pointed to.

src/Frame.cpp
Luthaf: Yes, I think that bond perception should take pbc into account, because it is not used only for visualization. For example I used this code to generate LAMMPS input from XYZ files, and in this case we want to get all bonds in the system, even the ones crossing periodic boundary.

To prevent this code from generating ugly bonds in a visualization application, I could see two approaches: one is to set the cell to an infinite cell before this calculation, and reset the original cell afterward. This would make the distance calculation behave exactly as the code as currently written. The other approach is the one taken by [VESTA](http://jp-minerals.org/vesta/en/), where broken molecules get duplicated when a bond would cross the cell. See the image below, where the input only contains atoms inside the cell
paracetamol

ynaimi: Thank you, your input is interesting and it makes sense. I will modify the code accordingly and prefer `this->distance(atom1,` atom2)` instead of a standard distance calculation.

ynaimi:
Hello Guillaume,

I made the changes we discussed above, please tell me if formatting is better now and allows you to review more easily the new code.

Also, you can find attached a [benchmark ](https://github.com/chemfiles/chemfiles/files/5532017/benchmark.pdf)I made between the brute force and the cell lists algorithms for generating bonds based on interatomic distance.

Don't hesitate if you have further remarks.

Luthaf:
The benchmarked speed-up look amazing! The formatting is still weird though. 

Unfortunately, the code itself looks more like C code than C++: all the uses of malloc/free should be replaced by a standard container instead (std::vector, std::unique_ptr<>, etc.); and parameters should be passed without any pointers. Do you think you can refactor this to use C++11 style? Else I can try to have a go at it later.

ynaimi:
Sure! 
I will make these changes. I am sorry the formatting is making the code review harder for you (in addition to having a different format compared to the rest of the code). I tried to reverse VS auto-formatting but apparently didn't succeed everywhere. I will take another look into that as well.

src/Frame.cpp
```diff
@@ -1,227 +1,493 @@
...
+            }
+            if (xCell && yCell < (yCells - 1) && zCell) {
+               nbList[cellIndex].push_back(cellIndex - xyTotalCells + xCells - 1);
+            }
+            if (xCell < (xCells - 1) && yCell < (yCells - 1) && zCell) {
+               nbList[cellIndex].push_back(cellIndex - xyTotalCells + xCells + 1);
+            }
+
+            //End of the list
+            nbList[cellIndex].push_back(-1);
```
Luthaf, line 240: this is not needed for std::vector, since we already know where is the end of a sub-vector.

src/Frame.cpp
```diff
@@ -1,227 +1,493 @@
...
+
+   min[0] = x1;
+   max[0] = x2;
+   min[1] = y1;
+   max[1] = y2;
+   min[2] = z1;
+   max[2] = z2;
+}
+
+int Frame::createNeighborList(std::vector<std::vector<int>>& nbList, int xCells, int yCells, int zCells) {
```
Luthaf, line 150: `std::vector<std::vector<int>>` have horrible CPU cache behavior (cf https://stackoverflow.com/q/38244435/4692076), this might explain part of the slowdown you see.

src/Frame.cpp
```diff
@@ -1,227 +1,493 @@
...
+   getMinMaxBox(positions_, numAtoms, min, max);
+
+   //Box's xyz ranges
+   float xRange = max[0] - min[0];
+   float yRange = max[1] - min[1];
+   float zRange = max[2] - min[2];
+
+   //Box's number of cells in each dim, in x and y dimensions, and in total (in x, y and z dimensions)
+   const float inversedCutoff = 1.0 / cutoff;
+   int xCells = ((int)(xRange * inversedCutoff)) + 1;
```
Luthaf, line 277: please use `std::floor` or `std::ceil` explicitly instead of casting to int

src/Frame.cpp
```diff
@@ -1,227 +1,493 @@
...
+   int yCells = ((int)(yRange * inversedCutoff)) + 1;
+   int zCells = ((int)(zRange * inversedCutoff)) + 1;
+   int xyTotalCells = yCells * xCells;
+   int totalCells = xyTotalCells * zCells;
+
+   //Check if total cells is not negative
```
Luthaf, line 283: I'm not sure how this could even happen. Could we use `assert(xRange > 0)` and so on for y and z instead?

src/Frame.cpp
```diff
@@ -1,227 +1,493 @@
...
+   cells.resize(totalCells);
+   numInCell.resize(totalCells);
+   maxInCell.resize(totalCells);
+
+   for (int i = 0; i < numAtoms; i++) {
+      int atomXCellNum, atomYCellNum, atomZCellNum, atomCellIndex, numOfAtomsInCell;
+
+      //Compute atom's cell index
+      Vector3D atomCoord = positions_[i];
+      atomXCellNum = (int)((atomCoord[0] - min[0]) * inversedCutoff);
```
Luthaf, line 315: please use `std::floor` or `std::ceil` explicitly instead of casting to int

src/Frame.cpp
```diff
@@ -1,227 +1,493 @@
...
+      atomYCellNum = (int)((atomCoord[1] - min[1]) * inversedCutoff);
+      atomZCellNum = (int)((atomCoord[2] - min[2]) * inversedCutoff);
+
+      atomCellIndex = atomZCellNum * xyTotalCells + atomYCellNum * xCells + atomXCellNum;
+
+      //Grow cell
+      numOfAtomsInCell = numInCell[atomCellIndex];
+      if (numOfAtomsInCell == maxInCell[atomCellIndex]) {
+         cells[atomCellIndex].resize(numOfAtomsInCell + 4);
+         maxInCell[atomCellIndex] += 4;
```
Luthaf, line 325: where does the +4 comes from?

src/Frame.cpp
```diff
@@ -1,227 +1,493 @@
...
+
+   cells.clear();
+   nbList.clear();
+   numInCell.clear();
+
+   if (bondsCount == 0 || bondsCount > maxNumBonds)
+   {
+      if (bondsCount == 0)
+      {
+         throw error(
```
Luthaf, line 353: I don't see why this should be an error, it can happen if all atoms are far appart

src/Frame.cpp
```diff
@@ -1,227 +1,493 @@
...
+         cells[atomCellIndex].resize(numOfAtomsInCell + 4);
+         maxInCell[atomCellIndex] += 4;
+      }
+
+      //Store atom index in appropriate cell
+      cells[atomCellIndex][numOfAtomsInCell] = i;
+      numInCell[atomCellIndex]++;
+   }
+
+   maxInCell.clear();
```
Luthaf, line 333: Looks like maxInCell is not used anywhere

src/Frame.cpp
```diff
@@ -1,227 +1,493 @@
...
+      }
+   }
+
+   // Go through the bond results vector to add bonds to connectivity
+   for (int bond = 0; bond < bondResults.size() / 2; bond++) {
+
+      int atom1Index = bondResults[bond * 2];
+      int atom2Index = bondResults[(bond * 2) + 1];
+
+      Atom atom1 = topology_[atom1Index];
```
Luthaf, line 371: this is making copies. Could you do `const Atom& atom1 = ...` or  better `const auto& atom1 = ...`

src/Frame.cpp
```diff
@@ -1,227 +1,493 @@
...
+      Atom atom2 = topology_[atom2Index];
+
+      //Prevent hydrogen atoms from bonding with each other
+      std::string atom1Type = atom1.type();
+      std::string atom2Type = atom2.type();
+      if (atom1Type != "H" || atom2Type != "H") {
+         // Add a bond, bondorder defaults to 1, bond type to -1
```
Luthaf, line 378: there is no bond type in chemfiles

src/Frame.cpp
```diff
@@ -1,227 +1,493 @@
...
+         add_bond(atom1Index, atom2Index);
+      }
+   }
+
+   bondResults.clear();
```
Luthaf, line 383: no need to clear, this is a local variable so memory willl be released when the scope exit.

src/Frame.cpp
```diff
@@ -234,18 +500,81 @@ const std::unordered_map<std::string, double> BOND_GUESSING_RADII = {
...
+         return it->second;
+      }
+   }
+   else {
+      // default to chemfiles provided atom type
+      return atom.vdw_radius();
+   }
+}
+
+int Frame::generateBonds(std::vector<int>& result, int totalCells, std::vector<std::vector<int>>& cells,
```
Luthaf, line 521: same as above, `std::vector<std::vector>` have bad CPU cache behavior. 

For `cells`, each sub-vector size is fixed at 27, right? Then it could be `std::vector<std::array<int, 27>>`.

Luthaf: This function could also directly add bonds to the frame instead of creating an extra "result" vector

Luthaf (commented):
This is starting to look better, but there is still a lot of work to fully bring the code to modern C++.

I have a few idea of where the slowdown you see could come from, I've commented on them below.

include/chemfiles/Frame.hpp
```diff
@@ -40,362 +40,376 @@ class Atom;
...
+	/// files to provide it.
+	///
+	/// @throw Error if the Van der Waals radius in unknown for a given atom.
+	///
+	/// @example{frame/guess_bonds.cpp}
+	void guess_bonds();
+
+	//***********************************************************************************
+
+	void getMinMaxBox(std::vector<Vector3D> pos, int n, Vector3D& min, Vector3D& max);
```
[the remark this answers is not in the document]
ynaimi: Sure, I will add these descriptions in the next commit.

include/chemfiles/Frame.hpp
```diff
@@ -40,362 +40,376 @@ class Atom;
...
+
+	int createNeighborList(std::vector<std::vector<int>>& nbList, int xCells, int yCells, int zCells);
```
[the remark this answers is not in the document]
ynaimi: Same as above, I will be adding it in the next commit.

src/Frame.cpp
```diff
@@ -1,227 +1,493 @@
...
+         to_remove.push_back(bond);
+      }
+   }
+
+   for (auto& bond : to_remove) {
+      topology_.remove_bond(bond[0], bond[1]);
+   }
+}
+
+void Frame::getMinMaxBox(std::vector<Vector3D> pos, int n, Vector3D& min, Vector3D& max) {
```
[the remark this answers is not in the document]
ynaimi: I absolutely missed that one, thank you for pointing out! 
About possible skewed triclinic cells, this function compute the dimensions of a cubic box which contains all the atoms of the structure by using atoms located at the extremities of each dimension. Therefore, I can't see how a different cell type will make it not work.

src/Frame.cpp
ynaimi: Absolutely right. It was needed for the array used in the previous version but not anymore with std::vector. Thanks.

src/Frame.cpp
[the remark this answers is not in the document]
Luthaf: Actually, I could not pinpoint issues specific to skewed cells, but there are a few issues related to periodicity:

cell-list-ortho

Here the unit cell is in black, and images in grey. The boxes created by the cell list (I'll call them boxes to prevent confusion with the unit cell) are in red. Atoms are in bright colors, and atom images semi-transparent.

One issue with the current implementation is that it will not find the bond between atoms far from one another in the central cell image (red and green atoms), although there should be one. A second related issue is the one of atoms outside the central image, that should have bond with atoms inside the central image (orange and blue).

I'll continue looking around about this skewed cell question.

src/Frame.cpp
ynaimi: > same as above, `std::vector<std::vector>` have bad CPU cache behavior.
> 
> For `cells`, each sub-vector size is fixed at 27, right? Then it could be `std::vector<std::array<int, 27>>`.

No, here we don't know the size of the sub-vector where atoms are stored. The `cells` data structure stores the atoms in cells depending on their position. Maybe the name of the variable is not clear enough but with documentation it will be clearer for sure. Anyway, I will change the `std::vector<std::vector>` data structure to have a better CPU cache behavior.

src/Frame.cpp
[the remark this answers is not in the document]
ynaimi: Thanks for the illustration. Indeed, I can clearly see the issue regarding periodicity. But I am wondering why the unit cell (in black) doesn't superimpose to the boxes created by the cell list (in red) which is containing all the atoms of the structure. Is it because of the chemfiles' way of defining a unit cell in periodic boundary conditions?

src/Frame.cpp
ynaimi: You are right, I didn't check the performance of such data structure using `std::vector`. For the `nblist` variable (known size of sub-vector), I am changing this to `std::vector<std::array<int, 27>>`. From first runs, this modification led to a high decrease of the slowdown seen before.

src/Frame.cpp
ynaimi: Sure.

src/Frame.cpp
ynaimi: Sure.

src/Frame.cpp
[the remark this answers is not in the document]
Luthaf: >  I am wondering why the unit cell (in black) doesn't superimpose to the boxes created by the cell list (in red) 

Isn't the code in this PR using the min/max positions of atoms in the system to create the cell list? I agree that it should use the unit cell faces instead.

src/Frame.cpp
ynaimi: It was used like the `numInCell` vector but I will remove the two of them now that I am using std::vector I no more need these data structures.

src/Frame.cpp
ynaimi: I was an arbitrary increase of the array but now that I am using `std::vector` I don't need that anymore, I am removing it.

src/Frame.cpp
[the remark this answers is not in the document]
ynaimi: Yes it is using min/max positions to define the box containing the structure. I will check how to use the unit cell faces instead in order to better manage periodicity.

ezavod:
Hi @ynaimi, your benchmarks look really promising. However implementing a cell list that respects all weird edge cases is rather hard. Even large projects with deeply committed contributors like MDAnalysis struggle with a correct implementation. Some time ago I reported an issue regarding their search algorithm and later also some other people found [additional problematic cases](https://github.com/MDAnalysis/mdanalysis/issues?q=is%3Aissue+nsgrid). AFAIK their current course of action is to disable the cell list until somebody steps up to the task. So don't get disenchanted if it takes some time to iron out all weird bugs that might come up :smiley:
A while ago I gave implementing a cell list a try myself [(see here)](https://gitlab.com/len_k/nsgrid-rs). I don't claim that this is bug-free or optimal at all. It was mostly for fun and for my own analysis that I also check by hand.
Regarding your discussion about skewed cells: At the time I found it the easiest way to simply transform any given cell into an orthorhombic cell. If the cell is infinite simply use a bounding box as you have implemented already. For triclinic cells one can use the diagonal elements of the cell matrix to get the cartesian dimensions of an equivalent orthorhombic box (I think).
Then you can subdivide in smaller orthorhombic sub-cells as you already do. It's probably the best to avoid any "overhangs" at the sides. So the [lengths of the sub-cells must be an integer multiple](https://gitlab.com/len_k/nsgrid-rs/-/blob/ca962ace285ea7b6690ba842fe523b541f30b964/src/bounding_box.rs#L109-132) of the containing UnitCell. (E.g.: UnitCell length = 12 and you want sub-cell length = 2.5 -> choose 4 sub-cells with length 3 instead.) When calculating these sub-cell lengths it might be beneficial to always round up. Note that the sub-cells might have different lengths in x, y and z.

ynaimi:
Hi @ezavod,

Thanks a lot for your encouraging words and for your very interesting feedback.

> So don't get disenchanted if it takes some time to iron out all weird bugs that might come up 😃

Now that @luthaf has pointed out periodicity issues, there are indeed few cases that need to be properly handled. I am taking a look on how to do that maybe first for common orthorhombic cells then additional cases later in the next versions.

> Regarding your discussion about skewed cells: At the time I found it the easiest way to simply transform any given cell into an orthorhombic cell. If the cell is infinite simply use a bounding box as you have implemented already. For triclinic cells one can use the diagonal elements of the cell matrix to get the cartesian dimensions of an equivalent orthorhombic box (I think).

Thanks for the input, it is very intresting. Can you tell me more about how to extract the cartesian dimensions from the diagonal elements of triclinic cell matrix?

> Then you can subdivide in smaller orthorhombic sub-cells as you already do. It's probably the best to avoid any "overhangs" at the sides. So the [lengths of the sub-cells must be an integer multiple](https://gitlab.com/len_k/nsgrid-rs/-/blob/ca962ace285ea7b6690ba842fe523b541f30b964/src/bounding_box.rs#L109-132) of the containing UnitCell. (E.g.: UnitCell length = 12 and you want sub-cell length = 2.5 -> choose 4 sub-cells with length 3 instead.) When calculating these sub-cell lengths it might be beneficial to always round up. Note that the sub-cells might have different lengths in x, y and z.

Yes, I made the same implementation in my code :)

ezavod:
> Can you tell me more about how to extract the cartesian dimensions from the diagonal elements of triclinic cell matrix?

@ynaimi A naive implementation would something like this:
```c++
if (cell_.shape() == UnitCell::INFINITE) {
   // calculate MinMaxBox...
   xRange = max[0] - min[0];
   yRange = max[1] - min[1];
   zRange = max[2] - min[2];
} else {
   xRange = cell_.matrix()[0][0];
   yRange = cell_.matrix()[1][1];
   zRange = cell_.matrix()[2][2];
}
```
But I just realized that this has some issues :(
Chemfiles' latest addition allows constructing cells from arbitrary matrices. This makes life hard for this PR (diagonal for non--infinite cells might be all zeros) and might also have some issues in itself (e.g. `UnitCell({ 0, 0, 1, /**/ 1, 0, 0, /**/ 0, 1, 0 })` is erroneously(?) triclinic).

PR #382, opened 2020-11-19
fix(mmtf): chain and group indices are consecutive

### LINKED ISSUE #344: Error reading 1HTQ.mmtf with read_step
The code below fails with various errors (see the associated PR #345 to catch the error early)

```cpp
auto trajectory = Trajectory("1HTQ.mmtf");
auto frame = trajectory.read_step(7);
```

The file contains 10 models, so reading step 7 should be fine. Reading with repeated calls to `trajectory.read();` works fine, and I managed to narrow the issue to the code in `read_step` settings the right atom/chain/group indexes https://github.com/chemfiles/chemfiles/blob/0cca28aefc85ec8940e2c777edb7b2d2a95cc356/src/formats/MMTF.cpp#L108-L125

After the loops above, `atomIndex_` is 989527, but `structure_.atomIdList` only contains 978720 entries, meaning calls to `atom_id` read out of bounds. I could not find the origin of the issue in the loops above though 🤷‍♂️

ezavod:
I have not read the full spec but these examples [[1]](https://github.com/rcsb/mmtf/blob/master/spec.md#chainspermodel) [[2]](https://github.com/rcsb/mmtf/blob/master/spec.md#groupsperchain) indicate that chains and groups are indexed consecutively. This is also aparent when looking at [`StructureData::print`](https://github.com/rcsb/mmtf-cpp/blob/ee4b855cf53b4a3ee7a40cc62d57195875fe84a5/include/mmtf/structure_data.hpp#L1028). When skipping, chemfiles wrongfully starts at 0 again for every model.

Fixes #344.

### DIFF
--- a/src/formats/MMTF.cpp
+++ b/src/formats/MMTF.cpp
@@ -124,19 +124,17 @@ void MMTFFormat::read_step(const size_t step, Frame& frame) {
     // Fast-forward, keeping all indexes updated
     while(modelIndex_ != step) {
         auto chainsPerModel = static_cast<size_t>(structure_.chainsPerModel[modelIndex_]);
-        while(chainIndex_ != chainsPerModel) {
+        for (size_t j = 0; j < chainsPerModel; ++j) {
             auto groupsPerChain = static_cast<size_t>(structure_.groupsPerChain[chainIndex_]);
-            while(groupIndex_ != groupsPerChain) {
+            for (size_t k = 0; k < groupsPerChain; ++k) {
                 auto groupType = static_cast<size_t>(structure_.groupTypeList[groupIndex_]);
-                auto group = structure_.groupList[groupType];
+                const auto& group = structure_.groupList[groupType];
                 auto atomCount = group.atomNameList.size();
                 atomIndex_ += atomCount;
                 groupIndex_++;
             }
-            groupIndex_ = 0;
             chainIndex_++;
         }
-        chainIndex_ = 0;
         modelIndex_++;
     }
 
--- a/src/formats/MMTF.cpp
+++ b/src/formats/MMTF.cpp
@@ -68,6 +68,21 @@ MMTFFormat::MMTFFormat(std::string path, File::Mode mode, File::Compression comp
         auto file = TextFile(std::move(path), mode, compression);
         auto buffer = file.readall();
         decode(buffer.data(), buffer.size(), file.path());
+        if (!mmtf::isDefaultValue(structure_.atomIdList)) {
+            // If ids are not ordered or are missing consecutive values the atoms
+            // have to be re-ordered.
+            bool isValidIdOrder = false;
+            if (structure_.atomIdList[0] == 1) {
+                // hack `is_sorted` to check that all ids are consecutive and sorted
+                isValidIdOrder = std::is_sorted(
+                    new_atom_indexes_.begin(), new_atom_indexes_.end(),
+                    [](const int32_t& lhs, const int32_t& rhs) { return (lhs + 1) == rhs; });
+            }
+            if (!isValidIdOrder) {
+                new_atom_indexes_ = structure_.atomIdList;
+                std::sort(new_atom_indexes_.begin(), new_atom_indexes_.end());
+            }
+        }
     } else if (mode == File::WRITE) {
         filename_ = std::move(path); // We really don't need to do anything, yet
     } else if (mode == File::APPEND) {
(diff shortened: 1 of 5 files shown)


ezavod:
While doing this I noticed that there is a ["reduced" version](https://mmtf.rcsb.org/download.html) of MMTF. Do we support this format (then it's a bug)? If not it would probably nicer to have a more verbose error message and note this on the format table.

Luthaf:
The fix looks good to me, let's merge it once we resolve the question of "reduced" MMTF

> Do we support this format (then it's a bug)? 

I don't know, but I would think that we do support it. It looks to me that the file format is the same, the only difference is which atoms are included in the files. So it is more a question of what's in the MMTF version of the PBD database than two different version of the format itself.

It would be good to have a test reading one of these reduced MMTF files though.

ezavod:
It seems that `natoms` in `read_model` is the reduced number of atoms and we add more atoms [here](https://github.com/chemfiles/chemfiles/blob/ee078b20e849c93b91c4ca19d740c0aaf5020435/src/formats/MMTF.cpp#L301). I'll try to find an easy solution.

ezavod:
Ok, I'm now fairly confident that I know what's going on. The problem is the determination of ids/indices with the function `MMTFFormat::atom_id`. Normally this function uses the lookup table `structure_.atomIdList` to determine the actual order of atoms as intended by the writer. Additionally the global id over all frames is converted to the local id within a given frame. This is fine as long as all values from the lookup table are smaller than the number of atoms / frame size.
The reduced file format stores ids within the lookup table as if all atoms were present which results in converted ids larger than the actual number of atoms. I don't see an easy workaround because the ids from the lookup table are not equally spaced so one would have to keep track of the number of missing atoms.
One technically possible workaround would be to simply use the `else` branch in any case, but this would break the order of atoms (failing test case "Structure not from RCSB" because order of atoms 1 and 2 are switched). This is probably bad for people using MMTF more extensively than me.

<details>
<summary>This are the first entries of the reduced 1HTQ.mmtf (global ids in second column):</summary>

```
ATOM    000002  CA      .       THR     A       A       601     .       106.152 52.289  139.747 55.330  0.100   C       0
ATOM    000009  CA      .       GLU     A       A       602     .       108.791 50.000  138.117 39.830  0.100   C       0
ATOM    000018  CA      .       LYS     A       A       603     .       107.972 48.955  134.461 46.200  0.100   C       0
ATOM    000027  CA      .       THR     A       A       1       .       108.460 51.143  131.310 17.130  0.100   C       0
ATOM    000034  CA      .       PRO     A       A       2       .       107.805 51.072  127.473 16.890  0.100   C       0
ATOM    000041  CA      .       ASP     A       A       3       .       110.869 49.051  126.318 39.320  0.100   C       0
ATOM    000049  CA      .       ASP     A       A       4       .       109.928 46.227  128.709 37.650  0.100   C       0
ATOM    000057  CA      .       VAL     A       A       5       .       106.459 46.062  127.084 41.150  0.100   C       0
ATOM    000064  CA      .       PHE     A       A       6       .       108.202 45.573  123.709 38.220  0.100   C       0
ATOM    000075  CA      .       LYS     A       A       7       .       110.182 42.655  125.170 27.270  0.100   C       0
ATOM    000084  CA      .       LEU     A       A       8       .       106.997 41.050  126.573 18.130  0.100   C       0
ATOM    000092  CA      .       ALA     A       A       9       .       105.431 41.395  123.104 27.380  0.100   C       0
ATOM    000097  CA      .       LYS     A       A       10      .       108.697 40.352  121.411 28.330  0.100   C       0
ATOM    000106  CA      .       ASP     A       A       11      .       109.340 37.257  123.587 26.120  0.100   C       0
ATOM    000114  CA      .       GLU     A       A       12      .       105.726 36.077  123.330 41.630  0.100   C       0
ATOM    000123  CA      .       LYS     A       A       13      .       105.416 36.166  119.488 34.890  0.100   C       0
ATOM    000132  CA      .       VAL     A       A       14      .       102.510 38.607  119.986 31.580  0.100   C       0
ATOM    000139  CA      .       GLU     A       A       15      .       100.232 38.741  116.903 24.130  0.100   C       0
ATOM    000148  CA      .       TYR     A       A       16      .       97.840  41.484  118.075 24.500  0.100   C       0
ATOM    000160  CA      .       VAL     A       A       17      .       97.682  44.393  120.510 17.440  0.100   C       0
ATOM    000167  CA      .       ASP     A       A       18      .       94.356  45.154  122.178 20.350  0.100   C       0
```
</details>

Luthaf:
So what you are saying is that we currently don't support reduced MMTF files, am I right? If that's the case, I'm fine with merging this PR, and opening an issue to add support for reduced files. 

If there is a way to know whether a file is using full or reduced MMTF, it would be great to add an error if someone try to open a reduced file.

ezavod:
Well I just bodged a workaround together that isn't particular pretty but it works(TM) :D
This should also retain most of the performance for normal files but reduced one are a bit slower.

src/formats/MMTF.cpp
```diff
@@ -68,6 +68,20 @@ MMTFFormat::MMTFFormat(std::string path, File::Mode mode, File::Compression comp
         auto file = TextFile(std::move(path), mode, compression);
         auto buffer = file.readall();
         decode(buffer.data(), buffer.size(), file.path());
+        if (!mmtf::isDefaultValue(structure_.atomIdList)) {
+            // If ids are not ordered or are missing consecutive values the atoms
+            // have to be re-ordered.
+            bool isValidIdOrder = false;
+            if (structure_.atomIdList[0] == 1) {
+                isValidIdOrder = std::is_sorted(
```
Luthaf, line 76: is this hacking `std::is_sorted` to check if the vector contains consecutive values?

Luthaf: If so it might be worth a comment =)

ezavod: yep, could not think of a better way

Luthaf (approved):
Thanks a lot! This bug was really puzzling me.

Could you also create a PR to backport this to the release-0.10 branch?

ezavod:
Sure, I'll open a backport PR immediately.
The doc build on travis has failed, but the timeout seems unrelated to these changes.

Luthaf:
>  The doc build on travis has failed, but the timeout seems unrelated to these changes.

yeah, that's sphinx linkcheck failing to reach archive.org =/ I'll see if we can tell linkcheck to ignore some links

PR #386, opened 2020-12-08
fix: reject rotated orthorhombic cell matrices

### LINKED ISSUE #384: Deal with vector permutation in the cell matrix
Chemfiles' latest addition allows constructing cells from arbitrary matrices. This makes life hard for this PR (diagonal for non--infinite cells might be all zeros) and might also have some issues in itself (e.g. `UnitCell({ 0, 0, 1, /**/ 1, 0, 0, /**/ 0, 1, 0 })` is erroneously(?) triclinic).

_Originally posted by @ezavod in https://github.com/chemfiles/chemfiles/issues/381#issuecomment-738287888_

ezavod:
Fixes #384 by rejecting unit cells that are not diagonal and orthorhombic. This means orthorhombic UnitCells can be constructed from diagonal matrices only.

### DIFF
--- a/src/UnitCell.cpp
+++ b/src/UnitCell.cpp
@@ -110,6 +110,42 @@ static Matrix3D cell_matrix_from_lenths_angles(Vector3D lengths, Vector3D angles
     return matrix;
 }
 
+static Vector3D calc_lengths_from_cell_matrix(const Matrix3D& matrix) {
+    Vector3D v1 = {matrix[0][0], matrix[1][0], matrix[2][0]};
+    Vector3D v2 = {matrix[0][1], matrix[1][1], matrix[2][1]};
+    Vector3D v3 = {matrix[0][2], matrix[1][2], matrix[2][2]};
+
+    return {v1.norm(), v2.norm(), v3.norm()};
+}
+
+static Vector3D calc_angles_from_cell_matrix(const Matrix3D& matrix) {
+    Vector3D v1 = {matrix[0][0], matrix[1][0], matrix[2][0]};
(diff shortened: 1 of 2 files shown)


src/UnitCell.cpp
```diff
@@ -123,6 +159,12 @@ UnitCell::UnitCell(Matrix3D matrix): matrix_(std::move(matrix)), matrix_inv_(Mat
         throw error("invalid unit cell matrix with negative determinant");
     }
 
+    auto lengths = calc_lengths_from_cell_matrix(matrix_);
+    auto angles = calc_angles_from_cell_matrix(matrix_);
+    if (!is_diagonal(matrix_) && is_orthorhombic(lengths, angles)) {
+        throw error("rotated ORTHORHOMBIC cells are not supported");
```
Luthaf, line 165:
```suggestion
        throw error("orthorhombic cell must have their a vector along x axis, b vector along y axis and c vector along z axis");
```

Luthaf (approved):
Looks good. I'm trying to think of a similar issue with triclinic matrix, but I can not find one on the top of my head.

ezavod:
> I'm trying to think of a similar issue with triclinic matrix, but I can not find one on the top of my head.

The main issue is that we treat orthorhombic cells specially. In case we want arbitrary rotated orthorhombic cells we can simply use `calc_*_from_cell_matrix` in `lengths()` and `angles()`. This would mean we treat triclinic and orthorhombic cells the same except the initial shape assignment. But that could be added later when a use-case arises.

Luthaf (approved):

Issue #288, opened 2019-11-28
Improve atomic properties


Luthaf:
This is an idea I had to improve atom-level properties. The issues I can see with the current system are:
1. since properties are stored on a per-atom basis, we don't know if all atoms in a topology/frame will have a given property (say `is_standard_pdb`), which make it somehow painful to use properties (we always need to define a default if the property is not present)
2. some atomic level data feel as first-class citizen while other do not. Why are `charge`, `mass` and `vdw_radius` always defined, and not `partial_charge`, `spin`, `anisou`?

The idea is to completely get rid of the `Atom` class, and to represent atomic level properties as a `PropertyVector`, stored in the corresponding topology.  A property vector is basically a `std::vector<Property>`, where all properties have the same type. This means that the Topology would look like

```cpp
class Topology {
    // ...
    Connectivity connectivity_;
    std::vector<Residue> residues_;
    std::map<std::string, PropertyVec> properties_;
}
```

Specific properties such as the atoms names/types and potentially mass and atomic number and charge could still get somehow privileged treatment:

```cpp
class Topology {
    // ...
    std::map<std::string, PropertyVec> properties_;
    // these are references (or pointers) into the properties_ map, 
    // allowing the corresponding  data to be accessed without lookup
    std::vector<double>& masses_;
    std::vector<double>& masses_;
}
```

Incidentally, this would also give us a `std::vector<double>` property type, for #227.

---  

In some cases, properties can legitimately only be defined for a few selected atoms. Instead of picking a default value for all the other atoms, we could introduce a new `Property::Missing` type to represent these.

---

This would increase the complexity of dealing with properties, while removing the `Atom` class. We should check that the overall complexity of the code dealing with atoms do not increases significantly.

PR #391, opened 2021-02-17
feat: rewrite LAMMPS trajectory reader

### LINKED ISSUE #371: Rewrite LAMMPS trajectory reader
We are currently using the molfiles implementation, with all the related problems, cf https://github.com/chemfiles/chemfiles/issues/154.

ezavod:
The new implementation allows reading and writing of the LAMMPS dump/atom format. This format is customizable and we can handle the most common custom attributes: units, time, masses, charges, names, and velocities.
Writing is also relatively complex. This implementation writes unwrapped coordinates which is not part of the atom style, but rather the custom variant. My understanding is that all atoms in the `atom` dump _must_ be wrapped, so using the `custom` style is faster for us. This also writes always mass and charge because there is no reliable way to detect whether it is explicitly set or the default of 0. #288 might help here in the future.
The handling of scaled coordinates is taken from [this part in molfiles](https://github.com/chemfiles/molfiles/blob/df74add3f1dc0fdc0a06e9e5200425b4241a09ec/src/lammpsplugin.c#L1280-L1327). Haven't really checked from 'first principles' if this conversion makes sense.

This closes #371 and ticks a box in #154.


Luthaf:
Nice! I am a bit underwater right now, but I'll try to have a look at this next week

Luthaf (commented):
This looks quite good! One area of improvement would be the error messages/exceptions. I think these should contain the format name as much as possible, to ensure good error message are propagated even when the user is not in control of the format.

src/formats/LAMMPSAtom.cpp
```diff
@@ -0,0 +1,641 @@
...
+            double zlo = parse<double>(splitted[0]);
+            double zhi = parse<double>(splitted[1]);
+            matrix[2][2] = zhi - zlo;
+            origin[2] = zlo;
+            if (shape == UnitCell::TRICLINIC) {
+                matrix[1][2] = parse<double>(splitted[2]);
+            }
+
+            auto cell = UnitCell(matrix);
+            cell.set_shape(shape);
```
ezavod, line 115: Reminder to self: this overwrites infinite cells with orthorhombic.

src/formats/LAMMPSAtom.cpp
```diff
@@ -0,0 +1,641 @@
...
+        file_.print("\n");
+    }
+}
+
+optional<uint64_t> LAMMPSAtomFormat::forward() {
+    auto position = file_.tellpos();
+    size_t natoms = 0;
+    try {
+        auto line = file_.readline();
+        if (line.empty()) {
```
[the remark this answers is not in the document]
ezavod: Empty lines should not happen within a file, but my intention was to test for eof. It's probably best and clearer to solely test for eof.

src/formats/LAMMPSAtom.cpp
```diff
@@ -0,0 +1,641 @@
...
+        auto item = get_item(line);
+        while (!line.empty() && (!item || *item != "NUMBER OF ATOMS")) {
+            line = file_.readline();
+            item = get_item(line);
+        }
+        if (!item || *item != "NUMBER OF ATOMS") {
+            throw error("could not find the number of atoms header");
+        }
+        int64_t lmp_natoms = parse<int64_t>(trim(file_.readline()));
+        natoms = static_cast<size_t>(lmp_natoms);
```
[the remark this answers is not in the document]
ezavod: Oh, I forgot to remove parts of my debug code.

src/formats/LAMMPSAtom.cpp
```diff
@@ -0,0 +1,641 @@
...
+        return nullopt;
+    }
+    return trim(splitted[1]);
+}
+
+std::array<double, 3> LAMMPSAtomFormat::read_cell(Frame& frame) {
+    auto item = get_item(file_.readline());
+    try {
+        if (!item) {
+            throw "expected an ITEM entry";
```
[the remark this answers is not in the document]
ezavod: Done.

src/formats/LAMMPSAtom.cpp
```diff
@@ -0,0 +1,641 @@
...
+            std::array<double, 3> origin;
+            auto shape = UnitCell::ORTHORHOMBIC;
+            if (splitted.size() >= 5 && (*item).find("xy xz yz") != string_view::npos) {
+                shape = UnitCell::TRICLINIC;
+            }
+            auto line = file_.readline();
+            auto splitted = split(line, ' ');
+            if ((shape == UnitCell::ORTHORHOMBIC && splitted.size() != 2) ||
+                (shape == UnitCell::TRICLINIC && splitted.size() != 3)) {
+                throw error("incomplete box dimensions");
```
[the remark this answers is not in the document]
ezavod: Done.

src/formats/LAMMPSAtom.cpp
```diff
@@ -0,0 +1,641 @@
...
+            origin[0] = xlo;
+            if (shape == UnitCell::TRICLINIC) {
+                matrix[0][1] = parse<double>(splitted[2]);
+            }
+
+            line = file_.readline();
+            splitted = split(line, ' ');
+            if ((shape == UnitCell::ORTHORHOMBIC && splitted.size() != 2) ||
+                (shape == UnitCell::TRICLINIC && splitted.size() != 3)) {
+                throw error("incomplete box dimensions");
```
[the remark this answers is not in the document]
ezavod: Done.

src/formats/LAMMPSAtom.cpp
```diff
@@ -0,0 +1,641 @@
...
+    IMGY,
+    IMGZ,
+    VELX,
+    VELY,
+    VELZ,
+    CHARGE,
+};
+
+// possible representations of the atomic position
+enum l_pos_repr_t {
```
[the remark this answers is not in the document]
ezavod: Done.

src/formats/LAMMPSAtom.cpp
```diff
@@ -0,0 +1,641 @@
...
+        } else {
+            throw format_error("expected BOX BOUNDS item");
+        }
+    } catch (const Error& e) {
+        throw format_error("can not read box header: {}", e.what());
+    }
+}
+
+// posible per-atom attributes by dump command
+enum l_atm_attr_t {
```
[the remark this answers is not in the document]
ezavod: Finding nice names for things in code is hard :)

src/formats/LAMMPSAtom.cpp
```diff
@@ -0,0 +1,641 @@
...
+        }
+    }
+}
+
+void LAMMPSAtomFormat::write_next(const Frame& frame) {
+    if (frame.get("units")) {
+        file_.print("ITEM: UNITS\n{:s}\n", (*frame.get("units")).as_string());
+    }
+    if (frame.get("time")) {
+        file_.print("ITEM: TIME\n{:.16g}\n", (*frame.get("time")).as_double());
```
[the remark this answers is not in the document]
ezavod: Good point, [LAMMPS real units](https://lammps.sandia.gov/doc/units.html) correspond to Angstroms & fs. I was vaguely aware that chemfiles uses those units from the XDR formats, but couldn't find any conclusive docs about this.

src/formats/LAMMPSAtom.cpp
```diff
@@ -0,0 +1,641 @@
...
+    bool has_names = false;
+    for (size_t i = 0; i < frame.size(); ++i) {
+        auto& atom = frame[i];
+        if (!atom.type().empty() && !has_types) {
+            try {
+                int type = parse<int>(atom.type());
+                if (type > 0)
+                    has_types = true;
+            } catch (const Error&) {
+                // ignore types that are not a number
```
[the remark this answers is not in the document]
ezavod: Good suggestion. I've added some code that deals with atom types. It now uses the atom types from the frame until the first invalid type is encountered. From then on there are consecutive types generated, even for valid types that might follow later on.

src/formats/LAMMPSAtom.cpp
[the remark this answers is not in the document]
Luthaf: You are right, this is not really documented. I'll open an issue!

doc/src/properties/frame.toml
```diff
@@ -36,9 +36,15 @@ CML = "The text of the ``title`` attribute in a molecule node. Only set if the a
 [time]
 type = "number"
 
+LAMMPS = "The time of the frame in time units."
 TRR = "The time of the frame in pico seconds."
 XTC = "The time of the frame in pico seconds."
 
+[units]
```
Luthaf, line 43: Since this is specific to LAMMPS, we may want to call this property `lammp_units`. This way we don't end up using any `units` property defined by another format when writing a LAMMPS file.

src/formats/LAMMPSAtom.cpp
```diff
@@ -0,0 +1,665 @@
...
+            // no more data to read, so give up here
+            return nullopt;
+        }
+        auto item = get_item(line);
+        while (!file_.eof() && (!item || *item != "NUMBER OF ATOMS")) {
+            line = file_.readline();
+            item = get_item(line);
+        }
+        if (!item || *item != "NUMBER OF ATOMS") {
+            throw error("could not find the number of atoms header in LAMMPS format");
```
Luthaf, line 638:
```suggestion
            throw format_error("could not find the number of atoms header in LAMMPS format");
```

src/formats/LAMMPSAtom.cpp
```diff
@@ -0,0 +1,665 @@
...
+        }
+        natoms = parse<size_t>(trim(file_.readline()));
+        // read the box
+        for (size_t i = 0; i < 4; ++i) {
+            file_.readline();
+            if (file_.eof()) {
+                throw error("could not read box data in LAMMPS format");
```
Luthaf, line 645:
```suggestion
                throw format_error("could not read box data in LAMMPS format");
```

src/formats/LAMMPSAtom.cpp
```diff
@@ -0,0 +1,665 @@
...
+            }
+        }
+        file_.readline();
+        if (file_.eof()) {
+            throw error("could not read atom header in LAMMPS format");
```
Luthaf, line 650:
```suggestion
            throw format_error("could not read atom header in LAMMPS format");
```

src/formats/LAMMPSAtom.cpp
```diff
@@ -0,0 +1,665 @@
...
+        }
+
+        for (size_t i = 0; i < natoms; ++i) {
+            file_.readline();
+            if (file_.eof()) {
+                throw error("could not read enough atoms in LAMMPS format");
```
Luthaf, line 656:
```suggestion
                throw format_error("could not read enough atoms in LAMMPS format");
```

src/formats/LAMMPSAtom.cpp
```diff
@@ -0,0 +1,665 @@
...
+            throw format_error("missing 'BOX BOUNDS' item in LAMMPS format");
+        }
+    } catch (const Error& e) {
+        throw format_error("can not read box header in LAMMPS format: {}", e.what());
+    }
+}
+
+// posible per-atom attributes by dump command
+enum lammps_atom_attr_t {
+    UNKNOWN,
```
Luthaf, line 138: Could you add small comments on all attributes telling what they are for someone not familiar with LAMMPS trajectory files?

Luthaf (commented):
Before merging this, I would like to see more varied tests, in particular around the different ways we can read coordinates. Current tests files (`polymer.lammpstrj` and `nacl.lammpstrj`) use wrapped and scaled coordinates respectively. I think we should also have unwrapped and unwrapped scaled test files, as well as a file to check the selection logic in `detect_best_pos_representation`.

Do you have small-ish examples of these formats? If not I can help to produce them.

src/formats/LAMMPSAtom.cpp
ezavod: Added descriptions as provided by the [dump command (see `custom` style)](https://lammps.sandia.gov/doc/dump.html).

doc/src/properties/frame.toml
ezavod: Changed to `lammps_units`.

ezavod:
> Before merging this, I would like to see more varied tests, in particular around the different ways we can read coordinates. Current tests files (`polymer.lammpstrj` and `nacl.lammpstrj`) use wrapped and scaled coordinates respectively. I think we should also have unwrapped and unwrapped scaled test files, as well as a file to check the selection logic in `detect_best_pos_representation`.

Some basic tests are done by using the memory reader in test sections 'Best position representation' and 'Triclinic boxes'. But I can generate some additional text files as complementary tests.

> Do you have small-ish examples of these formats? If not I can help to produce them.

I don't currently have any at hand, but they should be easy to generate. Will look into this over the next few days. If you already have any suitable files, feel free to add those.

Luthaf:
> Some basic tests are done by using the memory reader in test sections 'Best position representation' and 'Triclinic boxes'. But I can generate some additional text files as complementary tests.

Sorry, I missed those for some reason. I'm happy with the amount of testing for the different coordinates styles in this case!

I could not restart the CI run with Travis web UI, so I'll close & reopen the PR to restart Travis!

ezavod:
No worries. I have already created some text files for tests and I think that having more varied tests won't hurt. Especially since this lammps format is rather flexible, it is nice to have a tight test set so that we won't break anything in the future. So expect some additions soonish :)

src/formats/LAMMPSAtom.cpp
```diff
@@ -0,0 +1,686 @@
...
+    }
+    return trim(splitted[1]);
+}
+
+std::array<double, 3> LAMMPSAtomFormat::read_cell(Frame& frame) {
+    auto line = file_.readline();
+    auto item = get_item(line);
+    try {
+        if (!item) {
+            format_error("expected an ITEM entry in LAMMPS format, got '{}'", line);
```
ezavod, line 64: Note to self: missing `throw`.

src/formats/LAMMPSAtom.cpp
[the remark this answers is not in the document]
ezavod: I changed my mind on this one: while editing the additional test files either I or my editor added superfluous new lines at the end of the file which causes cryptic parsing errors. So I thinks it's better to also abort on newlines just in case people have messy trajectories.

tests/formats/lammps-atom.cpp
```diff
@@ -0,0 +1,553 @@
...
+    }
+
+    SECTION("Errors") {
+        auto file = Trajectory("data/lammps/broken.lammpstrj");
+        std::string msg = "can not read box header in LAMMPS format: expected an ITEM entry in "
+                          "LAMMPS format, got 'DUMMY'";
+        CHECK_THROWS_WITH(file.read_step(0), msg);
+        msg =
+            "can not read box header in LAMMPS format: missing 'BOX BOUNDS' item in LAMMPS format";
+        CHECK_THROWS_WITH(file.read_step(1), msg);
```
ezavod, line 172: Not sure if it's intended or not, but using `file.read()` here does not work. The file cursor is at the position of the last error instead of the position of the next frame, this means that `file.read()` simply starts reading the next line. In contrast to that, `frame.read_step(1)` starts at the correct position for the next frame.
IMHO this is a bug (but not a serious one) because as a user one gets the impression that an read error should be recoverable because we throw an error (or return an result in Rust).

Luthaf: Oh, that's interesting. This should be not too hard to fix, I'll open an issue.

Luthaf (approved):
This looks good, thanks a lot!

Luthaf:
Sorry, I forgot to merge this 🤦

ezavod:
Thanks for your time to review this.

Luthaf:
And thank you for implementing this!

PR #393, opened 2021-02-28
Switch from Travis to github actions

Luthaf:
This is the final repository in the chemfiles project to switch, and the largest one! 

I'll think it is better to keep Travis for the release-0.10 branch, and only use GH Actions for master. This will take a while to debug, expect a lot of force push!


Luthaf:
Moved to a PR from a branch in this repo #394

PR #394, opened 2021-02-28
Switch from Travis to github actions

### LINKED ISSUE #282: Update travis to use ubuntu 16.04
We are currently using `dist: trusty` in travis CI config, which is Ubuntu 14.04, release in 2014 and reached end of life in April 2019.

We should update it to `dist: xenial`, i.e. ubtuntu 16.04.

The main thing to check for when doing this update is support for gcc 4.8/cmake 2.8.12, either by installing them in the new version, or by using a docker image.

Luthaf:
This is the final repository in the chemfiles project to switch, and the largest one!

I'll think it is better to keep Travis for the release-0.10 branch, and only use GH Actions for master. This will take a while to debug, expect a lot of force push!

Fix #282


Luthaf:
This will also fix #282!

Luthaf:
After more than 60 force pushes, this should now be ready!

PR #401, opened 2021-04-10
Remove configuration file

### LINKED ISSUE #400: Remove configuration files
The initial idea for the configuration file was to rename atomic types and configure atomic radii for LAMMPS simulations specifically, where atom types are numeric and bond guessing would fail.

This can be achieved in other ways (such as setting the topology of the frame after the fact), and now that we have read-write access to the atoms even manually scripted if needed. Since they introduce global state and coupling, and are not that usefull, I think we should remove the configuration files altogether.

I don't think configuration files are used by anyone, let me know if I'm wrong!

Luthaf:
It seems that they are not used that often, and can be replaced by setting a custom topology for a given trajectory. 

Not having configuration files also remove an implicit and mostly invisible dependency on filesystem state.

Fixes #400


Luthaf:
I'll merge this in a few days if nobody disagree

PR #405, opened 2021-06-26
Add a guess_format function returning the format we would use for a given filename

Luthaf:
This will be used in the javascript bindings for 0.10 to decide when to use "native" in memory IO or write the file to a temporary in-memory filesystem and read/write from there.

@frodofine or @ezavod do you have some time to give this a look in the next week?

### DIFF
--- a/include/chemfiles/misc.hpp
+++ b/include/chemfiles/misc.hpp
@@ -27,25 +27,6 @@ typedef std::function<void(const std::string& message)> warning_callback_t; // N
 /// @param callback callback function that will be called on each warning
 void CHFL_EXPORT set_warning_callback(warning_callback_t callback);
 
-/// Read configuration data from the file at `path`.
-///
-/// By default, chemfiles reads configuration from any file named
-/// `.chemfiles.toml` or `chemfiles.toml` in the current directory or any parent
-/// directory. This function can be used to add data from another configuration
-/// file.
-///
-/// This function will throw a `ConfigurationError` if there is no file at
-/// `path`, or if the file is incorrectly formatted. Data from the new
-/// configuration file will overwrite any existing data.
-///
-/// @example{add_configuration.cpp}
-///
-/// @param path path to the configuration file to add
-///
-/// @throws ConfigurationError if the file at `path` can not be read, or if it
-///                            is invalid.
-void CHFL_EXPORT add_configuration(const std::string& path);
-
 /// Get the list of formats chemfiles knows about, and all associated metadata
 ///
 /// @example{formats_list.cpp}
--- a/include/chemfiles/misc.hpp
+++ b/include/chemfiles/misc.hpp
@@ -32,6 +32,26 @@ void CHFL_EXPORT set_warning_callback(warning_callback_t callback);
 /// @example{formats_list.cpp}
 std::vector<std::reference_wrapper<const FormatMetadata>> CHFL_EXPORT formats_list();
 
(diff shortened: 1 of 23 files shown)


src/Trajectory.cpp
```diff
@@ -27,16 +29,65 @@ using namespace chemfiles;
...
+            path
+        );
+    }
+
+    auto registered_format = FormatFactory::get().by_extension(extension);
+    auto format = std::string(registered_format.metadata.name);
+
+    if (!compression.empty()) {
+        format += " / " + compression;
+    }
```
ezavod, line 73: AFAIK there is no way to check whether or not the format actually supports the compression. Previously, this wouldn't matter because trying to open something like "traj.xtc.gz" would [thrown an error in the constructor](https://github.com/chemfiles/chemfiles/blob/c87c3dca55888540430031ab6a3297cea1784e9c/src/formats/XTC.cpp#L58-L60) immediately. But `chemfiles::guess_format("traj.xtc.gz");` happily returns "XTC / GZ", which introduces the subtle distinction between formats that are recognized (but not necessarily supported) and actually supported formats. However, it might be only me who would misuse such a function to check if a certain file is supported :)
Would it be helpful to highlight this fact in the docs?

include/chemfiles/misc.hpp
```diff
@@ -27,29 +27,27 @@ typedef std::function<void(const std::string& message)> warning_callback_t; // N
...
+std::vector<std::reference_wrapper<const FormatMetadata>> CHFL_EXPORT formats_list();
+
+/// Get the format that chemfiles would use to read a file at the given
+/// path.
 ///
-/// This function will throw a `ConfigurationError` if there is no file at
-/// `path`, or if the file is incorrectly formatted. Data from the new
-/// configuration file will overwrite any existing data.
+/// The format is only guessed from the filename extension, chemfiles does not
+/// currently read the file to guess the format.
```
ezavod, line 39: Not opening also means that a successful guess does not mean that the format is actually supported (see discussion in Trajectory.cpp). Same wording also in the C API.

ezavod (commented):
It seems there are some more scattered mentions of configuration files. A quick search finds the following occurrences in the current branch:
```c
include/chemfiles/Atom.hpp
71:    /// To change the default value for a given type, you can use configuration
81:    /// can use configuration files.
113:    /// use configuration files.
125:    /// can use configuration files.
137:    /// can use configuration files.
149:    /// type, you can use configuration files.

include/chemfiles/Frame.hpp
213:    /// information is missing for a specific atoms, one can use configuration
```

frodofine:
I can review later this week if needed.

src/Trajectory.cpp
Luthaf: At the very least we should indeed explicitly document this behavior! 

My use case for this function mean that I will immediately call the trajectory constructor after using it, but I can see how it would be useful to check if a file would be supported. However, the fact that the extension of a file might be supported does not mean that the file is in the right format. For example, `.nc` files will be interpreted as following the Amber NetCDF convention, but such file could be any NetCDF file, meaning trying to open the file will fail. So for now I think keeping (and documenting) the current behavior (`guess_format` might return a format that later fails to open if the file is not formatted as expected, or uses unsupported compression scheme) should be fine.

Luthaf: Updated!

ezavod (approved):

PR #406, opened 2021-07-08
Backports to release-0.10

Luthaf:
This backports #353 and #405


Luthaf:
I'm not sure why the tests are not running, I'll investigate.

Luthaf:
0.10 is still using Travis, but it looks like the "please build this" event is no longer going through

Luthaf:
yay, it works! The chemfiles account on Travis had no plan selected, selecting the Free plan seems to fix the issue

PR #356, opened 2020-07-22
Initial CIF support

fxcoudert:
This is going to be, for now, work in progress. It does not compile. But I wanted to have first feedback on it, in particular if I understand correctly what are the files to modify.

The only function I have written for now is `CIFFormat::write` 😄

PR #407, opened 2021-07-09
CIF/mmCIF improvements

Luthaf:
This PR should fix all chemfiles' side issues reported in https://github.com/MDAnalysis/mdanalysis/issues/2367#issuecomment-876065924. I'll work on the MDAnalysis side for the remaining issues.

- make the mmCIF reader more strict, and move COD/CCSD files support to CIF reader
- make `guess_format` distinguish between CIF and mmCIF files if the extension is .cif

@frodofine since you wrote the mmCIF code, could you give this a look?

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -215,7 +215,7 @@ jobs:
       - name: run tests
         run: |
           cd build
-          ctest --build-config ${{ matrix.build-type }} --parallel 2
+          ctest --output-on-failure --build-config ${{ matrix.build-type }} --parallel 2
       - name: check the list of exported symbols
         if: matrix.check-exported-symbols
         run: ./scripts/ci/check-exported-symbols.py build/libchemfiles.so
--- a/include/chemfiles/formats/mmCIF.hpp
+++ b/include/chemfiles/formats/mmCIF.hpp
@@ -48,9 +48,6 @@ private:
     std::map<std::string, size_t> atom_site_map_;
     /// Map of residues, indexed by residue id and chainid.
     std::map<std::pair<std::string, int64_t>, Residue> residues_;
-    /// Set to true if the file is based on fractional coordinates.
-    /// Set to false if the file is based on cartesian coordinates.
-    bool uses_fract_;
     /// Storing the positions of all the steps in the file, so that we can
     /// just `seekpos` them instead of reading the whole step.
     std::vector<uint64_t> steps_positions_;
@@ -62,6 +59,7 @@ private:
     size_t atoms_;
     /// Frame properties need to be stored
     std::string name_;
+    /// The PDB icode, if any
     std::string pdb_idcode_;
 };
 
--- a/src/formats/mmCIF.cpp
+++ b/src/formats/mmCIF.cpp
@@ -122,16 +122,15 @@ void mmCIFFormat::init_() {
                     trimmed.substr(1, trimmed.size() - 2).to_string() : "";
         }
 
-        if (in_loop && line_split[0].find("_atom_site") != std::string::npos) {
+        if (in_loop && line_split[0].find("_atom_site.") != std::string::npos) {
             auto atom_label = line_split[0].substr(11).to_string();
-            to_ascii_lowercase(atom_label);
             atom_site_map_[atom_label] = current_index++;
             break;
         }
     }
 
     if (current_index == 0) {
-        throw format_error("no atom sites found in mmCIF file");
+        throw format_error("could not find _atom_site loop in '{}'", file_.path());
     }
 
     cell_ = UnitCell(lengths, angles);
@@ -142,7 +141,6 @@ void mmCIFFormat::init_() {
     do {
         if (line.find("_atom_site") != std::string::npos) {
             auto atom_label = trim(line).substr(11).to_string();
-            to_ascii_lowercase(atom_label);
             atom_site_map_[atom_label] = current_index++;
 
             position = file_.tellpos();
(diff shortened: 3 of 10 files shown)


frodofine:
I'll try to take a look next week

include/chemfiles/formats/mmCIF.hpp
```diff
@@ -48,9 +48,6 @@ class mmCIFFormat final: public Format {
     std::map<std::string, size_t> atom_site_map_;
     /// Map of residues, indexed by residue id and chainid.
     std::map<std::pair<std::string, int64_t>, Residue> residues_;
-    /// Set to true if the file is based on fractional coordinates.
-    /// Set to false if the file is based on cartesian coordinates.
-    bool uses_fract_;
```
frodofine, line 53: Why is this removed? I think in theory mmCIF can use fractional coordinates.

.github/workflows/tests.yml
```diff
@@ -215,7 +215,7 @@ jobs:
       - name: run tests
         run: |
           cd build
-          ctest --build-config ${{ matrix.build-type }} --parallel 2
+          ctest --output-on-failure --build-config ${{ matrix.build-type }} --parallel 2
```
frodofine, line 218: Do you intend to keep "--output-on-failure"?

src/formats/mmCIF.cpp
```diff
@@ -122,16 +122,15 @@ void mmCIFFormat::init_() {
                     trimmed.substr(1, trimmed.size() - 2).to_string() : "";
         }
 
-        if (in_loop && line_split[0].find("_atom_site") != std::string::npos) {
+        if (in_loop && line_split[0].find("_atom_site.") != std::string::npos) {
```
frodofine, line 125: Why have you added the dot? This this just to improve specificity?

src/formats/mmCIF.cpp
```diff
@@ -122,16 +122,15 @@ void mmCIFFormat::init_() {
...
             auto atom_label = line_split[0].substr(11).to_string();
-            to_ascii_lowercase(atom_label);
```
frodofine, line 127: Why is this removed?

src/formats/mmCIF.cpp
```diff
@@ -159,22 +157,14 @@ void mmCIFFormat::init_() {
...
-    // Attempt to guess how coordinates are stored
-    auto cartn_x = atom_site_map_.find("cartn_x");
-    auto fract_x = atom_site_map_.find("fract_x");
-    if (cartn_x != atom_site_map_.end()) {
-        uses_fract_ = false;
-    } else if (fract_x != atom_site_map_.end()) {
-        uses_fract_ = true;
-    } else {
-        throw format_error("CIF file does not define coordinates");
+    if (atom_site_map_.find("Cartn_x") == atom_site_map_.end()) {
```
frodofine, line 163: Okay, so this code no longer supports fractional coordinates?

src/formats/mmCIF.cpp
```diff
@@ -232,7 +222,7 @@ void mmCIFFormat::read(Frame& frame) {
     // and stored with optional. I don't know which solution is better.
 
     // These are required for atoms
-    auto type_symbol = atom_site_map_.find("type_symbol");
+    auto type_symbol = atom_site_map_.at("type_symbol");
```
frodofine, line 225: Because we are guaranteed to have it!

src/formats/mmCIF.cpp
```diff
@@ -309,20 +296,10 @@ void mmCIFFormat::read(Frame& frame) {
...
-            frame.add_atom(std::move(atom), Vector3D(x, y, z));
-        } else {
-            auto u = cif_to_double(line_split[fract_x->second].to_string());
-            auto v = cif_to_double(line_split[fract_y->second].to_string());
-            auto w = cif_to_double(line_split[fract_z->second].to_string());
-
-            auto vector = cell_.matrix() * Vector3D(u, v, w);
-            frame.add_atom(std::move(atom), vector);
-        }
+        auto x = cif_to_double(line_split[cartn_x].to_string());
```
frodofine, line 299: Fractional coordinates no longer supported

src/guess_format.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+    if (extension.empty()) {
+        throw file_error(
+            "file at '{}' does not have an extension, provide a format name to read it",
+            path
+        );
+    }
+
+    if (extension == ".cif" && (mode == 'r' || mode == 'a')) {
+        extension = distinguish_cif_variants(path, compression).value_or(extension);
```
frodofine, line 63: Everything makes sense now!

frodofine (approved):
Looks good to me, I have no changes. Most of my comments are for my understanding.

.github/workflows/tests.yml
Luthaf: Yes, this make debugging from CI much nicer

include/chemfiles/formats/mmCIF.hpp
Luthaf: No, mmCIF only defines `Cartn_x` and friends: https://mmcif.wwpdb.org/dictionaries/mmcif_pdbx_v40.dic/Categories/atom_site.html

CIF files define both `Cartn_x` and `fract_x`: https://www.iucr.org/__data/iucr/cifdic_html/1/cif_core.dic/Iatom_site_Cartn_.html / https://www.iucr.org/__data/iucr/cifdic_html/1/cif_core.dic/Iatom_site_fract_.html, but these files are handled by the GEMMI based CIF reader.

src/formats/mmCIF.cpp
Luthaf: Exactly. The issue that prompt these change was that `1AKE` contains data in the `_atom_sites.<XXX>` group, which this code would match. For mmCIF files everything related to atoms is inside the `_atom_site.<XXX>` group, hence this change.

src/formats/mmCIF.cpp
Luthaf: Just to make the code match the files. If the file uses `Cartn_x` or `pdbx_PDB_model_num`, I'd rather have the same strings in the code, without case change. Should also prevent reading data from the wrong entry if the file is malformed and contains both `cartn_x` and `Cartn_x`.

src/formats/mmCIF.cpp
Luthaf: Exactly. Support for fractional coordinates is now done with GEMMI/the CIF reader introduced in #356.

src/formats/mmCIF.cpp
Luthaf: yes

src/formats/mmCIF.cpp
Luthaf: yes

src/guess_format.cpp
Luthaf: ^-^ 🎉

orbeckst:
Thank you!

Issue #409, opened 2021-07-16
Lammps dump properties not found.


SamTov:
I am using chemfiles to read lammps dump files as a replacement for a reader that I have written as I would like to have a generic reader that can handle many formats. Lammps is however, the most important one at the moment. I find that chemfiles returns an empty list when I call `frame.list_properties ` and returns 0 when I call `frame.properties_count`. However, if I call `frame.velocities` or `frame.positions` it does return the velocities and positions. Even the `frame.has_velocities` returns true.  

Perhaps I have misunderstood how the properties work but when looking through the cpp code I noticed that chemfiles does in fact read additional properties from the data and I believe the charges on the atoms were also stored correctly. I would like to be able to access any of the properties in the dump, including custom ones. The code appears to do this but I cannot get the api to access them.

the header of the file is:

```bash
ITEM: TIMESTEP
100
ITEM: NUMBER OF ATOMS
1600
ITEM: BOX BOUNDS pp pp pp
0.0000000000000000e+00 3.1134167999999999e+01
0.0000000000000000e+00 3.1134167999999999e+01
0.0000000000000000e+00 3.1134167999999999e+01
ITEM: ATOMS id mol type element q x y z ix iy iz
```
I also frequently will dump forces, energy, or stresses along with the data.

My input script is quite simple.
```python
>>> from chemfiles import Trajectory

>>> data = Trajectory('test_data/dump.lammpstrj')
>>> frame = data.read()
>>> frame.nsteps
4
>>> frame.has_velocities()
True
>>> frame.list_properties()
[]
>>> frame.properties_count()
0
```

Luthaf:
So the handling of properties is not fully consistent currently, and that's something I'd like to improve before releasing version 1.0. In particular, velocities, mass and charge have a special treatment in the code.

With the current version, frame properties are used for single scalar values associated with the frame, and atom property are used for scalar values associated with each atom. To access the velocities for example you should do 

```python
data = Trajectory('test_data/dump.lammpstrj')
frame = data.read()

frame.velocities  # give a numpy ndarray with the velocities
```

and for the mass/charges

```python
data = Trajectory('test_data/dump.lammpstrj')
frame = data.read()

for atom in frame.atoms:
    atom.mass
    atom.charge
    # also not atomic "properties"
    atom.name
    atom.type
```

> I also frequently will dump forces, energy, or stresses along with the data.

[...]

PR #413, opened 2021-07-18
Read Arbitrary Per-Atom Properties in LAMMPS Atoms Format

ezavod:
Read per-atom properties (see discussion in #409). LAMMPS cannot store any frame properties, so nothing to do here. Also, all LAMMPS fixes, computes, or variables are either integer or double. Reading string properties is only a fallback if something went wrong, e.g. when dealing with incorrect user-generated data or upstream changes in LAMMPS.
Array data is written element-wise which means that a 3D property results in 3 scalar properties `myProp[I]` where `I = 1, 2, 3`.

Edit: Tests failing due to changes in #412. Waiting for these changes to be merged and then run tests again.
<details>
<summary>How to generate test input</summary>

```
variable x index 1
variable y index 1
variable z index 1
variable   t index 300

variable xx equal 10*$x
variable yy equal 10*$y
variable zz equal 10*$z

units         lj
atom_style    atomic

lattice       fcc 1.5
region        box block 0 ${xx} 0 ${yy} 0 ${zz}
create_box    1 box
create_atoms  1 box
mass     1 1.0

velocity all create 1.44 87287 loop geom

pair_style    lj/cut 2.5
pair_coeff    1 1 1.0 1.0 2.5

neighbor 0.3 bin
neigh_modify  delay 0 every 20 check no

fix      1 all nve

compute stress all stress/atom NULL
variable sq_pos atom "x^2+y^2+z^2"
fix 2 all property/atom i_flag

variable cluster atom (id%2)
set atom * i_flag v_cluster

dump 3 all custom 100 lj_dump.lammpstrj id type x y z c_stress[*] v_sq_pos i_flag
#dump 4 all netcdf 100 lj_dump.nc x y z c_stress[*]

#compute 6 all property/local ntype1 ntype2
#dump 5 all local 100 local.txt index c_6[1]

run      $t
```
</details>

### DIFF
--- a/src/formats/LAMMPSTrajectory.cpp
+++ b/src/formats/LAMMPSTrajectory.cpp
@@ -517,7 +517,7 @@ void LAMMPSTrajectoryFormat::read_next(Frame& frame) {
                 try {
                     // LAMMPS should always write double values
                     atom.set(fields[j].name, parse<double>(splitted[j]));
-                } catch (const Error& e) {
+                } catch (const Error&) {
                     // use the string value as fallback
                     atom.set(fields[j].name, splitted[j].to_string());
                 }
--- a/src/formats/LAMMPSTrajectory.cpp
+++ b/src/formats/LAMMPSTrajectory.cpp
@@ -137,7 +137,7 @@ std::array<double, 3> LAMMPSTrajectoryFormat::read_cell(Frame& frame) {
 /// variables
 enum lammps_atom_attr_t {
     // other possible attributes that are not important for chemfiles
-    UNKNOWN,
+    CUSTOM,
     // atom ID
     ATOMID,
     // atom type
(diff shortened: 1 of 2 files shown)


src/formats/LAMMPSAtom.cpp
```diff
@@ -505,7 +508,15 @@ void LAMMPSAtomFormat::read_next(Frame& frame) {
                 atom.set_charge(charge);
             } break;
             case ATOMID:
+                break;
             case UNKNOWN:
```
Luthaf, line 512: Could we rename `UNKNOWN` to `CUSTOM`?

src/formats/LAMMPSAtom.cpp
```diff
@@ -354,6 +354,8 @@ void LAMMPSAtomFormat::read_next(Frame& frame) {
     }
     std::vector<lammps_atom_attr_t> fields;
     fields.reserve(atoms_item.size() - 1);
+    std::vector<std::string> field_names;
+    field_names.reserve(atoms_item.size() - 1);
```
Luthaf, line 358: If you feel like it, these two vectors could be made into a single one by creating an `AtomField` struct

```cpp
struct AtomField {
    std::string name;
    lammps_atom_attr_t kind;
};

// ...

std::vector<AtomField> fields;

// ...
```

Luthaf (approved):
Looks good, I have two small improvements suggestions.

src/formats/LAMMPSAtom.cpp
ezavod: Agree, `CUSTOM` is more descriptive.

src/formats/LAMMPSAtom.cpp
ezavod: Done.

src/formats/LAMMPSTrajectory.cpp
```diff
@@ -507,7 +512,15 @@ void LAMMPSTrajectoryFormat::read_next(Frame& frame) {
                 atom.set_charge(charge);
             } break;
             case ATOMID:
-            case UNKNOWN:
+                break;
+            case CUSTOM:
+                try {
+                    // LAMMPS should always write double values
+                    atom.set(fields[j].name, parse<double>(splitted[j]));
+                } catch (const Error& e) {
```
Luthaf, line 520:
```suggestion
                } catch (const Error&) {
```

Luthaf (approved):

PR #420, opened 2021-08-28
Update Molfile.cpp

ShoubhikRaj:
The unit cell is correctly read from DCD file, so the information in supported formats page needs to be changed.

Tested with two different DCD files generated by NAMD 2.14 (attached). Chemfiles correctly reads the unit cell parameters, which can be accessed from `frame.cell().lengths()` and `frame.cell().angles()`.

[dcd_pbc_test.zip](https://github.com/chemfiles/chemfiles/files/7070129/dcd_pbc_test.zip)


Luthaf (approved):
Thanks for the pull request and for providing test files! I integrated them in our test suite, let me know if that's a problem!

PR #422, opened 2021-08-31
Add a test & fix for opening the same file from multiple threads

### LINKED ISSUE #418: Is chemfiles thread-safe (e.g. with OpenMP)?
I am using the C++ API of chemfiles in my code for analysis of a MD trajectory. There are a lot of data to process so it would be good if parallelisation can be done in the code. I **do not** mean that chemfiles should run in parallel, but that I wish to call chemfiles in a parallel code.

My preliminary test code, parallelised with OpenMP, seems to work correctly and produce that same results as the sequential code. However, I wish to know which part of chemfiles are thread safe, and which part are not (if there are any), so that I don't run into weird problems in future.

For example, is it safe to access different frames from different threads, from the same trajectory using the `trajectory.read_step(i)`? Is it safe to call `frame.angle(i,j,k)` with different values of `i`,`j` and `k`  from different threads for a common (shared) frame object?

I could not find anything on this in the documentation, so any advice would be very helpful.

Luthaf:
Fixes #418 

This PR contains a test that I hope would fail on Windows CI, I'll add the corresponding fix after seeing a failure.


Luthaf:
Here is the failure on CI: https://github.com/chemfiles/chemfiles/runs/3474270894?check_suite_focus=true

Luthaf:
And here are the changes to molfiles: https://github.com/chemfiles/molfiles/commit/e232a7c5649d9f925602ed4c0fb79e2af77e9302

Luthaf:
@ezavod do you think you could give this a quick look?

Luthaf:
I've also incorporated a [small fix](https://github.com/chemfiles/molfiles/commit/d571c0b45b0db84006a80a4e2998a7f50e5d1af7) to redirect the vmd plugin output to chemfiles' warning system through `vmdconio`. Ping @lmiq since you were one of the person bothered by this stdout/stderro pollution!

tests/trajectory.cpp
```diff
@@ -212,6 +213,75 @@ TEST_CASE("Guessing format") {
...
+    read_from_multiple_threads("data/mmtf/1HTQ_reduced.mmtf", 12336);
+    read_from_multiple_threads("data/dcd/water.dcd", 297);
+    read_from_multiple_threads("data/trr/1aki.trr", 38376);
+    read_from_multiple_threads("data/xtc/ubiquitin.xtc", 20455);
+
+    // FIXME: these fail, probably because the underlying library maintains a
+    // pool of ressources without synchronization
+
+    // read_from_multiple_threads("data/tng/1aki.tng", 38376);
+    // read_from_multiple_threads("data/netcdf/water.nc", 297);
```
ezavod, line 280: I vaguely remember something about netcdf-c not being thread safe at all. Not sure if anything has changed in recent versions.

ezavod (approved):
Looks good. This is awesome, but also fells dangerous at the same time.

tests/trajectory.cpp
Luthaf: Yeah, I would not be surprised here =). I'm exploring the possibility to remove the dependency on netcdf-c and instead directly read/write netcdf3 files, which seems to be doable (I have a reader mostly working). This would improve both this multi-threading behavior and the size of the binary (netcdf alone is 20% of the size of the JavaScript/WASM version of chemfiles)

Luthaf:
Thanks a lot for the review!

Issue #208, opened 2018-11-12
Implement PSF file format


Luthaf:
These files only contain topology information and seem to be widely used in bio-molecules simulations.

See here for a description: https://www.ks.uiuc.edu/Training/Tutorials/namd/namd-tutorial-unix-html/node23.html

The format is a simple text-based format that looks relatively free-form, but I could not find a specification of the format. Maybe looking at what VMD does would help.

frodofine:
It's also used by NAMD quite frequently. Which begs the question, should we also consider GROMACS `top` files?  These are also topology only.

Luthaf:
Yes, every format can be supported =) Would you mind opening an issue for `top` file, with some kind of reference for the format ?

For topology only format, I think we can just set all positions to 0 in the frame, and only parse the topology.

frodofine:
Sure, I'll do that.  That's what `InChI` does, so sounds good.

ShoubhikRaj:
Hi, I wanted to ask if you are planning to implement the psf reader using the psfplugin from VMD, or is anyone currently working on a parser for chemfiles?

I am asking because I need to read weights from psf files in the same C++ code I am developing. I was planning on writing my own code to read the psf myself, but if it is being implemented in chemfiles, then I would rather not "reinvent the wheel". (If no one is working on a parser, then I am happy to share my code with you)

Luthaf:
I don't think anyone is working on a PSF parser right now. If you want to try implementing PSF parsing within chemfiles I can help you get started. Looking at the example in the link in the first message, it should not be very hard to implement. By "weights" you mean the atomic mass, right?

If you prefer writing it inside your own code, sharing the result will be very helpful our implementation effort once we get there!

ShoubhikRaj:
@Luthaf Yes I meant the atomic masses.

There are actually several "flavours" of PSF files, (see [here](https://userguide.mdanalysis.org/stable/formats/reference/psf.html)). The molfile plugin contains a [psf reader](https://github.com/chemfiles/molfiles/blob/chemfiles/src/psfplugin.c) which is supposed to read all types of psf files. Since it is shipped with chemfiles already and it is tested by VMD so using it would be a stable solution. However, I don't know enough C or C++ to be able to write the interface between that and chemfiles.

Luthaf:
[...]

PR #423, opened 2021-09-05
PSF reader based on VMD molfile

Luthaf:
Temporary solution to #208.

~This is based on top of #422, which contains changes to the VMD molfile code, I'll rebase once it has been merged.~

### DIFF
--- a/doc/src/properties/residue.toml
+++ b/doc/src/properties/residue.toml
@@ -22,16 +22,20 @@ If the property is not set, ``HETATM`` is used."""
 type = "string"
 
 PDB = """The **chainname** defines the name assigned to a protein chain by
-biologists. It is is a single character used to group residues which are linked
+biologists. It is a single character used to group residues which are linked
 together in a crystallographic file where it may not be unique."""
 
+PSF = """The **chainname** is a single character used to group residues which
+are linked together. For PSF files, it is taken as the first letter of the
+``segname``."""
+
 mmCIF = """The **chainname** defines the name assigned to a protein chain by
-biologists. It is is a single character used to group residues which are linked
+biologists. It is a single character used to group residues which are linked
 together in a crystallographic file where it may *not* be unique. This name *is*
 unique to a given biological assembly, however."""
 
 MMTF = """The **chainname** defines the name assigned to a protein chain by
-biologists. It is is a single character used to group residues which are linked
+biologists. It is a single character used to group residues which are linked
 together in a crystallographic file where it may *not* be unique. This name *is*
 unique to a given biological assembly, however."""
 
@@ -39,6 +43,8 @@ unique to a given biological assembly, however."""
 type = "string"
 
 PDB = "For ``PDB`` files, the **chainid** is identical to **chainname**"
+
+PSF = "For ``PSF`` files, the **chainid** is identical to **chainname**"
 
 mmCIF = """The **chainname** defines the name assigned to a protein chain by
 biologists. It is is a single character used to group residues which are linked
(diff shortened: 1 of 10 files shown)


tests/formats/psf-molfile.cpp
```diff
@@ -0,0 +1,84 @@
...
+        CHECK(topology[1].charge() == static_cast<double>(0.556400f));
+        CHECK(topology[1].mass() == static_cast<double>(1.0080f));
+
+        CHECK(topology[3104].name() == "CAL");
+        CHECK(topology[3104].type() == "CAL");
+        CHECK(topology[3104].charge() == 2.0);
+        CHECK(topology[3104].mass() == static_cast<double>(40.0800f));
+
+        // FIXME: the molfile plugin does not read residue information, we
+        // should add it when re-implementing a PSF reader.
```
ezavod, line 34: Surely I missed something, but I think molfiles does read residues. Wouldn't the following three lines fix this?
The resids in the water test file don't seem sensible, but overall this works principally.
```patch
diff --git a/src/formats/Molfile.cpp b/src/formats/Molfile.cpp
index cd141548..0a81452b 100644
--- a/src/formats/Molfile.cpp
+++ b/src/formats/Molfile.cpp
@@ -298,6 +298,9 @@ template <MolfileFormat F> void Molfile<F>::read_topology() {
         }
         atom_id++;
     }
+    for (auto res : residues) {
+        topology_->add_residue(std::move(res.second));
+    }
 
     if (plugin_handle_->read_bonds == nullptr) {
         return;
```

Luthaf: Right, we are reading them but not adding them to the topology ... Good catch!!

doc/src/properties/residue.toml
```diff
@@ -25,6 +25,8 @@ PDB = """The **chainname** defines the name assigned to a protein chain by
 biologists. It is is a single character used to group residues which are linked
 together in a crystallographic file where it may not be unique."""
 
+PSF = """ TODO """
```
Luthaf, line 28: @ShoubhikRaj could you help me here? The chain name is set by the molfile plugin to be the first letter of the segment name, does it make sense to expose it to the users?

doc/src/properties/residue.toml
```diff
@@ -109,8 +113,10 @@ Examples include ``extended``, ``alpha helix``, ``pi helix``, ``turn`` and
...
 used for writing."""
 
-[resname]
+[segname]
 type = "string"
 PDB = """``PDB`` files used with CHARM, NAMD and a few other software package
 can contain a non-standard *segment* name associated with a residue. If set when
 writting a frame, this property is written to the file."""
+
+PSF = """ TODO """
```
Luthaf, line 122: @ShoubhikRaj what would be a good description of the segname property? This will be rendered as part of the documentation (http://chemfiles.org/chemfiles/latest/properties.html#residue-properties)

doc/src/properties/residue.toml
ShoubhikRaj: @Luthaf I don't do much work with biomolecules either. I have never needed to use the chain name. I think if molfiles reads it, there is no reason not to expose it to the users. If there are no biomolecules, then I don't know how much chainname is necessary. For example, the segName for water is TIP4P for all TIP4P type water systems, and getting chainname T is not particularly informative.

doc/src/properties/residue.toml
ShoubhikRaj: @Luthaf No idea sorry. What I do know is that the segName is used by VMD for some particular purposes. For example, if you want to solvate a system, the water model files supplied has to have the segName "QQQ" for the water molecules otherwise the VMD won't be able to determine which one is the solvent. (Using the `solvate` plugin)

doc/src/properties/residue.toml
Luthaf: >  Can it be different for other formats that chemfiles accesses via molfiles?

Not now, but maybe in the future if we add other molfiles format. This "the first letter of segname is chainname" behavior is specific to the PSF reader.

I've edited the documentation to mention that it is always the first letter, this way the format works in the same way as most biomolecules formats.

ezavod (approved):

PR #424, opened 2021-09-08
Store atom properties in extended XYZ as well

Luthaf:


src/formats/XYZ.cpp
```diff
@@ -252,6 +304,115 @@ std::string write_extended_comment_line(const Frame& frame) {
...
+                if (partially_defined_already_warned.count(property.first) == 0) {
+                    warning(
+                        "Extended XYZ",
+                        "property '{}' is only defined for a subset of atoms, it will not be saved",
+                        property.first
+                    );
+                    partially_defined_already_warned.insert(property.first);
+                }
+            }
+        }
```
ezavod, line 374: Might be a little too pedantic but I feel there is a lot of iterating going on.
Could this be moved after the next loop and only be done if `atom.properties().size() > all_properties` or if at least one property was present in all previous atoms but not this one?
So in most sane cases this would simply be skipped. Not sure if this has any real-world performance impacts.

src/formats/XYZ.cpp
```diff
@@ -252,6 +304,115 @@ std::string write_extended_comment_line(const Frame& frame) {
...
+                continue;
+            }
+        }
+
+        for (const auto& name: to_remove) {
+            all_properties.erase(name);
+        }
+    }
+
+    auto results = properties_list_t();
```
ezavod, line 409: Would this make things better?
```suggestion
    auto results = properties_list_t();
    results.reserve(all_properties.size());
```

src/formats/XYZ.cpp
```diff
@@ -252,6 +304,115 @@ std::string write_extended_comment_line(const Frame& frame) {
...
+    for (auto property: std::move(all_properties)) {
+        results.push_back({std::move(property.first), std::move(property.second)});
+    }
```
ezavod, line 412: Again, this might be too pedantic but can one not simply return the map `all_properties` instead if iterating over everything?
The only thing that changes is the use of `property.first/second` instead of `property.name/type` in just a few places.

ezavod (approved):
Looks good, just wondering if the looping might get too expensive for systems with very many properties.

Do I understand correctly that property names or values with spaces are supported (and automatically quoted) for frame level properties but not for atomic properties?

src/formats/XYZ.cpp
Luthaf: I'll try moving it, although I don't think this will have performance impact for any real world scenario. You would need to have a lot of atoms in the frame, all with very different properties.

src/formats/XYZ.cpp
Luthaf: I see your concern, but I think I prefer to have clearer code instead of optimizing this too early. I expect that in most cases this would contains 0-10 values, so this iteration is not very costly.

ezavod: Ok, I get the value of having clear code and you're right that one can reasonably expect a small number of properties :+1:

PR #426, opened 2021-09-10
fix(DCD): read frames with missing unit cell data

### LINKED ISSUE #419: Handling of unit cell parameters when they are missing from the file
I am reading a DCD file (generated with NAMD) with chemfiles. In NAMD, the simulation was run without any periodic boundary conditions, however, it seems that NAMD sets the unit cell lengths to 1 Angstrom and angles to 90 degrees each when there is no boundary.

The chemfiles [documentation](https://chemfiles.org/chemfiles/latest/formats.html) mentions that the unit cell information is not read from the DCD file format. So, I am not sure if chemfiles is setting those values by default, or whether it is actually reading them in from the DCD file. (The VMD molfile plugins do read the unit cell information from DCD.)

Either this is a bug in chemfiles, or the documentation has not been updated. 

Also, would it be possible/reasonable to recognize that the 1 Angstrom value of unit cell was set because there is no unit cell (because 1 Angstrom is unlikely to be the size of an unit cell in any practical simulation), and then set it to infinity (i.e. no unit cell)?

This would eliminate issues when reading a trajectory without PBC.

The DCD file, and the corresponding PSF and PDB files are attached: [testfiles.zip](https://github.com/chemfiles/chemfiles/files/7054282/testfiles.zip).

ezavod:
Molfile now defaults to all lengths set to zero. Did some quick tests and this does not seem to interfere with anything.

Depends on https://github.com/chemfiles/molfiles/pull/2.

Might fix #419.


Luthaf (approved):
Looks good!

ezavod:
Just updated the molfiles.tar.gz to the latest chemfiles branch just to be sure things are consistent. This should effectively contain the exact same files so I guess the hash change is due to timestamps.

Should be ready to merge.

PR #428, opened 2021-09-30
Update guess_format documentation

Luthaf:
We can now read the file in guess_format when trying to distinguish between formats with the same extension.

This is a follow-up to #407.


Luthaf:
I'll merge this in a couple of days if no-one disagree in the mean time!

PR #430, opened 2021-10-06
Read correct frame indices from TNG

### LINKED ISSUE #425: TNG: cannot read last frame of 1aki.tng
After playing a little bit more with #422 I noticed that the failing tests for TNG are not directly a problem of multi threading. Instead it's not possible to read the last step of `1aki.tng`.  Reading the last step of `example.tng` works. The file is either corrupted or somehow different. I'll try to investigate more by running `gmx check` in the next few days.

<details>
<summary>Simple testcase</summary>

```patch
diff --git a/tests/formats/tng.cpp b/tests/formats/tng.cpp
index 2fda2d23..238e38a8 100644
--- a/tests/formats/tng.cpp
+++ b/tests/formats/tng.cpp
@@ -6,7 +6,21 @@
 #include "chemfiles.hpp"
 using namespace chemfiles;

+#include <iostream>
+
 TEST_CASE("Read files in TNG format") {
+    SECTION("Read last step") {
+        auto file = Trajectory("data/tng/1aki.tng"); // example.tng works (has also 10 frames)
+
+        std::cout << "nsteps: " << file.nsteps() << std::endl;
+        std::cout << "Reading step 8 ..." << std::endl;
+        auto frame = file.read_step(8);
+        std::cout << "Done." << std::endl;
+        std::cout << "Reading step 9 ..." << std::endl;
+        frame = file.read_step(9); // fails
+        std::cout << "Done." << std::endl;
+    }
+
     SECTION("Read trajectory") {
         auto file = Trajectory("data/tng/example.tng");
         CHECK(file.nsteps() == 10);
```
</details>

ezavod:
Depends on https://github.com/chemfiles/tng/pull/1.

Updates the external TNG library to read the correct number of frames in a trajectory. Additionally, store the time of a frame as property, if present.

Fixes #425.


Luthaf:
The changes looks good. I don't know why emscripten is failing on CI, I'll look into it.

Luthaf:
Ok, let's merge this and we'll see later how we can have similar code upstreamed!

PR #433, opened 2021-10-08
Try using the new codecov uploader

### LINKED ISSUE #432: Update codecove uploader in CI
It seems we are using an deprecated uploader [for codecov]

> "The /bash endpoint is currently unavailable due to a scheduled brownout. The brownout schedule and more information can be seen found at https://about.codecov.io/blog/codecov-uploader-deprecation-plan/. It is recommended to update this repository to use the binary uploader, instructions can be found at https://docs.codecov.com/docs/codecov-uploader. This endpoint will permanently cease to function on Feb. 1, 2022."

[...] Simply copy and pasting the new uploader does not work.

_Originally posted by @ezavod in https://github.com/chemfiles/chemfiles/issues/430#issuecomment-938111213_

---

Although there is no urgency (we have until 2022 to do it), we should update to the new codecov uploader!

Luthaf:
Fixes #432 

Let's see how this goes! I hope the new uploader is able to discover & process gcov files.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -219,6 +219,12 @@ jobs:
       - name: check the list of exported symbols
         if: matrix.check-exported-symbols
         run: ./scripts/ci/check-exported-symbols.py build/libchemfiles.so
-      - name: upload coverage
+      - name: extract coverage data
         if: matrix.collect-coverage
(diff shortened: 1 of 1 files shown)


Luthaf:
Seems to be working, but we have to manually process the gcov files

PR #438, opened 2021-11-29
Use smaller archives for external dependencies

Luthaf:
This reduces the zlib and tng archives, bringing the size of chemfiles archive from 16MiB down to 6MiB. This was prompted by crates.io refusing to upload the crate for chemfiles-sys.

I have also added a test to catch any file size regression.

PR #439, opened 2021-11-30
Backports for release 0.10

Luthaf:
This backports #438 and #437


Luthaf:
Now also with #440 as well

PR #443, opened 2022-01-28
Replace netcdf-c with handwritten netcdf3 reader/writer

Luthaf:
This PR removes the dependency on the netcdf-c library, and replace it with a hand written reader/writer for netcdf3 files, the only format we use from netcdf-c to implement the Amber NetCDF convention.

There are three and a half reasons for this rewrite:
- reading files is much faster with the new code (see benchmarks example below, I can see a 2x increase);
- the `netcdf-c` library is quite painful to work with, and one of the main source of issues when distributing chemfiles through conda & other binary distribution;
- the code size went down from 5.1 MiB to 4.3 MiB in `-O3` mode (this is mainly interesting for the JS bindings to chemfiles);

The half reason is that netcdf3 is a relatively well documented format, and doing this allowed me to build a lot of infrastructure to be able to rewrite the DCD code and other binary formats later.

A lot of inspiration was taken from `scipy.io.netcdf`, which contains a pretty clear implementation of the netcdf3 format.

---

I'm opening this PR as a draft since I expect a couple of CI builder to fail with this code. I had to use mmap to get the same writing performance as `netcdf-c`, so this PR carries two implementations (mmap & cstdio) of the BinaryFile class.

## benchmarks

```cpp
// reading
auto file = Trajectory("data/netcdf/water.nc");
for (size_t i=0; i<file.nsteps(); i++) {
   auto frame = file.read();
}
```

Previous code took 2.1ms, the new code takes 0.8ms, more than 2x faster.

```cpp
// writing
auto frames = /* all frames from data/netcdf/water.nc */

auto file = Trajectory("tmp.nc", 'w');
for (const auto& frame: frames) {
   file.write(frame);
}
```
Previous code took 1.8ms, the new code takes 1.9ms, so there is no regression here.

----

Ping @pastewka, you were interested in part of this code. The NetCDF3 parser itself is mostly contained in `BinaryFile.hpp`, `BinaryFile.cpp`, `Netcdf3File.hpp` and `Netcdf3File.cpp`. Only the parts required for the AMBER NetCDF trajectory have been tested, but other datatypes for variables and attributes should work fine. Only files with 64-bit offset are supported as well, but 32-bit offset should not be too hard to support either. I don't have the time to support this as a standalone library, but feel free to copy files from here to your own project (this library is under the BSD licence). I'm happy to answer questions on the code if you want to try to adapt it.


Luthaf:
Alright, CI is happy, so this is ready for review! @ezavod do you think you'll have time to give this a look?

src/files/BinaryFile.cpp
```diff
@@ -0,0 +1,735 @@
...
+   || defined(_M_ALPHA) || defined(__amd64) \
+   || defined(__amd64__) || defined(_M_AMD64) \
+   || defined(__x86_64) || defined(__x86_64__) \
+   || defined(_M_X64) || defined(__bfin__)
+
+# define CHEMFILES_BYTE_ORDER CHEMFILES_LITTLE_ENDIAN
+#else
+# error Unknown target machine endianness
+#endif
+/***** adapted from boost ******/
```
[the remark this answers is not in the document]
Luthaf: That's interesting! I'll try to read a bit more on this.

tests/external/helpers.hpp
```diff
@@ -35,12 +37,19 @@ class NamedTempPath {
     const std::string& path() const {return path_;}
 
 private:
+    std::string base_;
```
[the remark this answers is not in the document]
Luthaf: leftover from development, I'll remove it.

tests/files/binary-files.cpp
```diff
@@ -0,0 +1,178 @@
...
+}
+
+TEST_CASE("Write binary files") {
+    SECTION("big endian") {
+        auto filename = NamedTempPath(".data");
+        {
+            auto file = BigEndianFile(filename, File::Mode::WRITE);
+            write_binary_file(file);
+        }
+
```
ezavod, line 113: While checking out how mmap works, I ran into a strange problem. The following trivial addition to the tests fails with error `mmap failed for '/tmp/chemfiles-tmp-BT1vac.data': Invalid argument`. The "append" section of `netcdf3-file` passes despite being seemingly the same thing. As I initially suspected a problem with continuous allocation, I tested this on a second machine with the same result.
```suggestion
        {
            auto file = BigEndianFile(filename, File::Mode::APPEND);
        }

```

Luthaf: I'll have a look, it might also be an issue with permissions when opening the file

include/chemfiles/files/Netcdf3File.hpp
```diff
@@ -0,0 +1,435 @@
...
+    constexpr int32_t NC_CHAR = 2;
+    constexpr int32_t NC_SHORT = 3;
+    constexpr int32_t NC_INT = 4;
+    constexpr int32_t NC_FLOAT = 5;
+    constexpr int32_t NC_DOUBLE = 6;
+    constexpr int32_t NC_DIMENSION = 10;
+    constexpr int32_t NC_VARIABLE = 11;
+    constexpr int32_t NC_ATTRIBUTE = 12;
+
+    static_assert(sizeof(char) == sizeof(int8_t), "what are you doing?");
```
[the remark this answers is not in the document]
Luthaf: I removed this one

tests/files/binary-files.cpp
Luthaf: ok, the implementation was broken for append mode (the file was resized to 0 since `total_written_size` was not initialized), I fixed it & added tests. That was a very good catch!

src/files/BinaryFile.cpp
```diff
@@ -0,0 +1,745 @@
...
+        ((value & 0x00000000000000FFULL) << 56) |
+        ((value & 0x000000000000FF00ULL) << 40) |
+        ((value & 0x0000000000FF0000ULL) << 24) |
+        ((value & 0x00000000FF000000ULL) <<  8) |
+        ((value & 0x000000FF00000000ULL) >>  8) |
+        ((value & 0x0000FF0000000000ULL) >> 24) |
+        ((value & 0x00FF000000000000ULL) >> 40) |
+        ((value & 0xFF00000000000000ULL) >> 56)
+    );
+}
```
ezavod, line 343: It just came to my mind that there might be builtin functions for swapping. And it turns out [there are](https://stackoverflow.com/a/46137633) (e.g. [glibc](https://sourceware.org/git/?p=glibc.git;a=blob;f=bits/byteswap.h;h=ca8417d59a80d3db8ac00a866b76cb3cd98a97f0;hb=HEAD)). This is also utilized by [libxtc](https://gitlab.com/impulse_md/libxtc/-/blob/58f40e1e9d392ec607e775772658eb1b748be549/libxtc/xtc_unpack.cpp#L54-96).
I'm not sure if it's worth bothering though because from my [preliminary fooling around](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGIM6SuADJ4DJgAcj4ARpjEEgCcpAAOqAqETgwe3r7%2ByanpAiFhkSwxcVyJdpgOGUIETMQEWT5%2BAVU1AnUNBEUR0bEJtvWNzTltwz2hfaUDFQCUtqhexMjsHOYAzKHI3lgA1CYbbsgKBPiCh9gmGgCC1zdeoQQAbJIA%2BgR7CgDuTElvXleEEegleHz2ADcxF5MHMDgB2Ky3PYovbETAEZYMPYQe6o/E4iBQ7yYA5mZ57DSqDQ02l0ukAMQZciCQThhzcHL2AFZnuz4ZzkQSUVBiTCyRSqfTpRomTSWWyDkcuZINPzBTdhSKidDSeZJdSZTS5bSFezlUc9mZJOq8VrRbqJZTDdKTXSzUrOZa9gAOW1C4UOklOqX0t30j2XS4ov0IjVawlivXk52MhlG%2BWs81XDbYK02uN2wM64P61Oy9MZzOKqO5vaq/2a%2B0l8VlqnhquR3PR3lzIt7PsbJF3eEAEXuE7MWwYOy8%2Bw5UQAngRMD8/gA6BBR24gl7vT7AWh4KLIN5RNf/IG7sGfYn84f49GY4jY8%2B/S%2BSEuD4cmMccBa0Jw3K8H4HBaKQqCcJyljWF8SwrMmGw8KQBCaP%2BCwANYgNyGj6JwkggWhEGcLwCggLhqFgf%2BpBwLASBoCwSR0LE5CUAxTH0HExAQq8AC0wDIMgXAaGYuE0LQK7EGREBRERUShA0i6cMhDFsIIADyDC0EpVGkFgLCGIeazgfg6I1BCq5EZgqjVF4K7KbwTyYIBulHlExCKR4WBEQQxB4CwDkLFQBjAAoABqeCYN86lJIwDkyIIIhiOwUgJfIShqERuhcPohkoNY1j6MeZGQAsqBJI4AhkRwvHqRspHOdUlV%2BBArijH4OXBFMJRlHoKRpM17V9fkzW9D1sy2I1HQMF0IyeC0ejtM1s2TMU/TlEM3RDTlpzdGN60SAsCjwasei%2BZgaw8ABQGEbpkEcAASgAkkIbi8WFaI8ZIAl7AJyB7MJ65mOuGg4tBViWKQey4IQJBkkhUMeIxzHEPDXBzLwlFaHMmHYbhLkEaQoHgfdpHkShaE43hHBmLdJMkRTVFUxZUkZCAkhAA%3D%3D), it seems there is only a difference in the assembler for MIPS and RISC architectures.

Luthaf: > It just came to my mind that there might be builtin functions for swapping. And it turns out [there are](https://stackoverflow.com/a/46137633) (e.g. [glibc](https://sourceware.org/git/?p=glibc.git;a=blob;f=bits/byteswap.h;h=ca8417d59a80d3db8ac00a866b76cb3cd98a97f0;hb=HEAD)).

I find it easier to just have the same implementation everywhere instead of using `#ifdef` everywhere to find the right one =)

> I'm not sure if it's worth bothering though because from my [preliminary fooling around](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGIM6SuADJ4DJgAcj4ARpjEEgCcpAAOqAqETgwe3r7%2ByanpAiFhkSwxcVyJdpgOGUIETMQEWT5%2BAVU1AnUNBEUR0bEJtvWNzTltwz2hfaUDFQCUtqhexMjsHOYAzKHI3lgA1CYbbsgKBPiCh9gmGgCC1zdeoQQAbJIA%2BgR7CgDuTElvXleEEegleHz2ADcxF5MHMDgB2Ky3PYovbETAEZYMPYQe6o/E4iBQ7yYA5mZ57DSqDQ02l0ukAMQZciCQThhzcHL2AFZnuz4ZzkQSUVBiTCyRSqfTpRomTSWWyDkcuZINPzBTdhSKidDSeZJdSZTS5bSFezlUc9mZJOq8VrRbqJZTDdKTXSzUrOZa9gAOW1C4UOklOqX0t30j2XS4ov0IjVawlivXk52MhlG%2BWs81XDbYK02uN2wM64P61Oy9MZzOKqO5vaq/2a%2B0l8VlqnhquR3PR3lzIt7PsbJF3eEAEXuE7MWwYOy8%2Bw5UQAngRMD8/gA6BBR24gl7vT7AWh4KLIN5RNf/IG7sGfYn84f49GY4jY8%2B/S%2BSEuD4cmMccBa0Jw3K8H4HBaKQqCcJyljWF8SwrMmGw8KQBCaP%2BCwANYgNyGj6JwkggWhEGcLwCggLhqFgf%2BpBwLASBoCwSR0LE5CUAxTH0HExAQq8AC0wDIMgXAaGYuE0LQK7EGREBRERUShA0i6cMhDFsIIADyDC0EpVGkFgLCGIeazgfg6I1BCq5EZgqjVF4K7KbwTyYIBulHlExCKR4WBEQQxB4CwDkLFQBjAAoABqeCYN86lJIwDkyIIIhiOwUgJfIShqERuhcPohkoNY1j6MeZGQAsqBJI4AhkRwvHqRspHOdUlV%2BBArijH4OXBFMJRlHoKRpM17V9fkzW9D1sy2I1HQMF0IyeC0ejtM1s2TMU/TlEM3RDTlpzdGN60SAsCjwasei%2BZgaw8ABQGEbpkEcAASgAkkIbi8WFaI8ZIAl7AJyB7MJ65mOuGg4tBViWKQey4IQJBkkhUMeIxzHEPDXBzLwlFaHMmHYbhLkEaQoHgfdpHkShaE43hHBmLdJMkRTVFUxZUkZCAkhAA%3D%3D), it seems there is only a difference in the assembler for MIPS and RISC architectures.

Both MIPS and RISC contains a direct call to the intrinsic function (`call    __bswapdi2`) instead of inlining it inside `glibc_bswap_64`, and this function should contain the same assembly as `swap_u64`.

Luthaf:
This should be ready for a new round of review!

src/files/BinaryFile.cpp
```diff
@@ -0,0 +1,751 @@
...
+#else
+    std::swap(this->file_, other.file_);
+#endif
+
+    return *this;
+}
+
+void BinaryFile::read_char(char* data, size_t count) {
+#if CHEMFILES_BINARY_FILE_USE_MMAP
+    std::memcpy(data, mmap_data_ + offset_, count);
```
ezavod, line 198: One can read beyond the end of a mmap which returns junk values. Might be better to throw an exception when reaching EOF. Should there be test case for that?
```suggestion
    if (offset_ + count > mmap_size_) {
        throw file_error(
            "failed to read {} bytes from the file at '{}': mmap out of bounds",
            count, this->path()
        );
    }
    std::memcpy(data, mmap_data_ + offset_, count);
```

tests/files/binary-files.cpp
```diff
@@ -0,0 +1,198 @@
...
+            auto file = LittleEndianFile(filename, File::Mode::APPEND);
+            write_binary_file(file);
+        }
+
+        content = read_binary_file(filename);
+        auto expected_2 = expected;
+        expected_2.insert(expected_2.end(), expected.begin(), expected.end());
+        CHECK(content == expected_2);
+    }
+}
```
ezavod, line 198: AFAIK chemfiles allows writing to newly created files in append mode like it's done [here](https://github.com/chemfiles/chemfiles/blob/6129ac7c06cb6261589818f4fd6b42276a6b9121/tests/formats/trr.cpp#L290).
This does not work for `BinaryFile`:
```cpp
TEST_CASE("Append on new trajectory") {
    SECTION("big endian") {
        auto filename = NamedTempPath(".data");
        {
            auto file = BigEndianFile(filename, File::Mode::APPEND);
            write_binary_file(file);
        }
    }
}
```

src/files/BinaryFile.cpp
ezavod: Having the same implementation everywhere makes sense:+1:

src/files/BinaryFile.cpp
Luthaf: reading past the end would probably segfault, but giving a good error message is a good idea. I'll triple check that it does not degrade performances.

tests/files/binary-files.cpp
Luthaf: I've added a bunch of tests for append mode, all the way to the final format implementation (BinaryFile, Netcdf3File, AmberNetCDF), hopefully everything should work now!

Luthaf:
CI is happy again, I think I'll merge this is a couple of days unless other issues are found!

PR #446, opened 2022-04-15
refactor(Atom): do not allocate map without atom properties

ezavod:
This makes the per-atom property map an optional field. Some (most?) formats don't make use of per-atom properties. Allocating the map for every atom nonetheless comes with quite some overhead. This is noticeable for frames with many atoms especially.

When reading 'ubiquitin.xtc' a ~11% speedup is observable (cf. [comparison to other implementations](https://github.com/chemfiles/benchmarks)).
For a `.nc` file with 50k atoms the speedup is ~19% for reading and ~7% for writing.

Ideally, this can be superseded by a more data-oriented approach as in #288.

This changes the user-facing API.

<details>
<summary>Benchmarking</summary>
R/W 1k steps with 50k atoms as `.nc` from C++ in a `tmpfs` ramdisk.

# master
```
Benchmark 1: ./bench_write
  Time (mean ± σ):     11.704 s ±  0.325 s    [User: 8.692 s, System: 3.004 s]
  Range (min … max):   10.672 s … 12.207 s    25 runs
 
Benchmark 1: ./bench_read
  Time (mean ± σ):      1.725 s ±  0.015 s    [User: 1.674 s, System: 0.050 s]
  Range (min … max):    1.697 s …  1.759 s    25 runs
```
# PR
```
Benchmark 1: ./bench_write
  Time (mean ± σ):     10.854 s ±  0.369 s    [User: 7.638 s, System: 3.209 s]
  Range (min … max):   10.102 s … 11.920 s    25 runs
 
Benchmark 1: ./bench_read
  Time (mean ± σ):      1.390 s ±  0.014 s    [User: 1.335 s, System: 0.054 s]
  Range (min … max):    1.370 s …  1.428 s    25 runs
```
</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - Added read support for PSF files using VMD molfile plugin.
 - Amber NetCDF files are now read/written with a custom netcdf parser (#443)
 
-### Changes to the C API
+### Changes to the C++ API
+- Per-atom properties are optional, i.e. `Atom::properties` returns an optional property map.
 
 ## 0.10.0 (14 Feb 2021)
 
--- a/include/chemfiles/Atom.hpp
+++ b/include/chemfiles/Atom.hpp
@@ -144,11 +144,12 @@ public:
     /// @example{atom/atomic_number.cpp}
     optional<uint64_t> atomic_number() const;
 
-    /// Get the map of properties associated with this atom. This map might be
-    /// iterated over to list the properties of the atom, or directly accessed.
+    /// Get the map of properties associated with this atom. If no properties
+    /// are set, this function returns `nullopt`. This map might be iterated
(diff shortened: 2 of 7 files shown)


Luthaf:
This is surprising, I would have expected that having an empty map would not allocate/cost us on run-time. I think this is a good solution until we overhaul atomic properties.

Luthaf (approved):

Luthaf:
Sorry, I forgot to merge this!

ezavod:
> Sorry, I forgot to merge this!

No worries, I wasn't expecting such a quick response over the holidays :-)

> expected that having an empty map would not allocate/cost us on run-time.

Just FYI: I'm using GCC 7.5.0 (libstdc++). [People on SO](https://stackoverflow.com/a/68561608) also say that there shouldn't be any allocations, so there is probably some other overhead. On first glance there are a lot of constructor/destructor calls for strings caused by the map.

PR #451, opened 2022-05-18
feat: rewrite TRR and XTC formats

ezavod:
This is a rewrite of the XDR-based files TRR and XTC which used GROMACS' xdrfile before.
Main point of the rewrite is the increased read performance that originates from the use of the new `BinaryFile` (cf. #443).
Reading TRR files is approx. 60% faster!
Writing speed of TRR and XTC files does not change significantly. Reading speed of XTC does not change significantly as well. Arguably, XTC is compute bound and does not profit from better IO.


Luthaf:
That looks nice! I should have time to do a first round of review this weekend.

include/chemfiles/formats/TRR.hpp
```diff
@@ -28,13 +26,45 @@ class TRRFormat final : public Format {
...
+        int ir_size;     /* Backward compatibility                                  */
+        int e_size;      /* Backward compatibility                                  */
+        int box_size;    /* Size in Bytes, non zero if a box is present             */
+        int vir_size;    /* Backward compatibility                                  */
+        int pres_size;   /* Backward compatibility                                  */
+        int top_size;    /* Backward compatibility                                  */
+        int sym_size;    /* Backward compatibility                                  */
+        int x_size;      /* Size in Bytes, non zero if coordinates are present      */
+        int v_size;      /* Size in Bytes, non zero if velocities are present       */
+        int f_size;      /* Size in Bytes, non zero if forces are present           */
```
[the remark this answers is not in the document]
ezavod: You're right, they are stored as 32bit integers so `int32_t` conveys the intent better.

src/formats/TRR.cpp
```diff
@@ -1,33 +1,50 @@
...
 #include "chemfiles/UnitCell.hpp"
-#include "chemfiles/FormatMetadata.hpp"
 
 #include "chemfiles/files/XDRFile.hpp"
 #include "chemfiles/formats/TRR.hpp"
 
+// definitions from the xdrfile library
+#define TRR_MAGIC 1993
+#define TRR_VERSION "GMX_trn_file"
+#define DIM 3
```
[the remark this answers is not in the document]
ezavod: Using this define made it easier to port the code from xdrfile. But keeping this obfuscation is not necessary.

include/chemfiles/files/XDRFile.hpp
```diff
@@ -4,65 +4,67 @@
...
+    /// Write an XDR unsigned integer
+    void write_uint(uint32_t value) { write_single_u32(value); }
+
+    /// Read an XDR double
+    double read_double() { return read_single_f64(); }
+
+    /// Read an XDR float
+    float read_float() { return read_single_f32(); }
+    /// Write an XDR float
+    void write_float(float value) { write_single_f32(value); }
```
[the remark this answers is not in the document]
ezavod: Yes, the functions of the underlying `BinaryFile` are now used instead.

include/chemfiles/files/XDRFile.hpp
```diff
@@ -4,65 +4,67 @@
...
-/// Check a return code from a xdrlib function, and throw a `FileError` if the
-/// status is not exdrOK.
-void check_xdr_error(int status, const std::string& function);
+    /// Read XDR doubles into an allocated vector
+    void read_double_array(std::vector<double>& data) { read_f64(data.data(), data.size()); }
+
+    /// Read XDR floats into an allocated vector
+    void read_float_array(std::vector<float>& data) { read_f32(data.data(), data.size()); }
+    /// Write XDR floats
+    void write_float_array(const std::vector<float> data) { write_f32(data.data(), data.size()); }
```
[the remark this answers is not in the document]
ezavod: They are now part of the `BinaryFile` and the tests have been moved accordingly. For now, I thinks it's best to keep the remaining few functions in the `XDRFile`.

include/chemfiles/files/BinaryFile.hpp
```diff
@@ -161,6 +161,8 @@ class BinaryFile: public File {
         this->read_f32(&value, 1);
         return value;
     }
+    /// Read exactly as many 32-bit floating point numbers as fit in the pre-allocated vector
+    void read_f32_array(std::vector<float>& data) { this->read_f32(data.data(), data.size()); }
```
Luthaf, line 165: I think we should provide these for all supported types, not only f32/f64. About the function name, It feels like `read_f32_array` and `read_f32` are similar enough to have the same name & use an overloaded function to distinguish between them. What do you think?

include/chemfiles/files/XDRFile.hpp
```diff
@@ -4,66 +4,45 @@
...
+/// Including additional helper routines for GROMACS
+class XDRFile final : public BigEndianFile {
+  public:
+    XDRFile(std::string path, File::Mode mode);
+    ~XDRFile() = default;
+
+    /// Read a non-compliant GROMACS string
+    /// A GROMACS string stores the length of the string including the NULL
+    /// terminator as int32 before the XDR compliant string data without the
+    /// terminator. An XDR string has the same represenmtation as opaque data.
```
Luthaf, line 25:
```suggestion
    /// terminator. An XDR string has the same representation as opaque data.
```

include/chemfiles/files/BinaryFile.hpp
ezavod: Agree, it makes sense to have overloaded functions and it cuts down on typing. I'm not to sure how to deal with overloading pure virtual functions though.

include/chemfiles/files/XDRFile.hpp
ezavod: Done.

include/chemfiles/files/XDRFile.hpp
```diff
@@ -19,6 +19,10 @@ class XDRFile final : public BigEndianFile {
     XDRFile(std::string path, File::Mode mode);
     ~XDRFile() = default;
 
+    using BinaryFile::read_f32;
+    using BinaryFile::read_f64;
+    using BinaryFile::write_f32;
```
ezavod, line 24: Not sure if this is the best way to expose the base functions in the derived class.

Luthaf: That's how I would do it as well =)

Luthaf (approved):

PR #471, opened 2023-03-14
Add check before fmt inclusion to avoid clash

PlathC:
Using chemfiles with other submodules can cause clashes like the following:
```
[CMake] CMake Error at <PathToChemfiles>/chemfiles/external/fmt/CMakeLists.txt:204 (add_library):
[CMake]   add_library cannot create target "fmt" because another target with the same
[CMake]   name already exists.  The existing target is a static library created in
[CMake]   source directory
[CMake]   "<PathToExternal>/fmt".  See
[CMake]   documentation for policy CMP0002 for more details.
```
This forces the user's project to rely on chemfiles' version of fmt which is not always possible.
To address this issue and allow more flexibility, I added a check before fmt inclusion avoiding clashes with a fallback to the previously defined target.

### DIFF
--- a/external/CMakeLists.txt
+++ b/external/CMakeLists.txt
@@ -135,19 +135,25 @@ if (${CMAKE_VERSION} VERSION_LESS "3.1")
     endfunction()
 endif()
 
-external_library(fmt)
-list(APPEND EXTERNAL_INCLUDES ${CMAKE_CURRENT_BINARY_DIR}/fmt/include/)
+if(NOT TARGET fmt)
+    external_library(fmt)
+    list(APPEND EXTERNAL_INCLUDES ${CMAKE_CURRENT_BINARY_DIR}/fmt/include/)
 
-# Hide FMT variables from CMake GUI
-mark_as_advanced(FORCE
-    FMT_CUDA_TEST
-    FMT_DOC
-    FMT_FUZZ
-    FMT_INSTALL
-    FMT_PEDANTIC
-    FMT_TEST
-    FMT_WERROR
-)
+    # Hide FMT variables from CMake GUI
+    mark_as_advanced(FORCE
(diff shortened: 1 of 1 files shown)


Luthaf:
I see your point here, but I'm not sure this is the best way to solve it for us: this is assuming that the `fmt` target matches the library we want. 

I think a better solution might be to prefix all libraries target in `external/` with `chemfiles::` (i.e. the target would be `chemfiles::fmt`), which should allow using chemfiles as a sub-module without target pollution. Since we are already vendoring all dependencies, this should be fully transparent from the user point of view.

PlathC:
I do agree that a prefix could be more elegant. 
However, I'm not aware of a CMake mechanism to inject a namespace into the target name to avoid redefinition during `add_subdirectory` apart from modifying `add_library` argument directly.
Would you recommend it or could you point me an alternative method?

Luthaf:
I don't know any such mechanism either 😄 I was thinking of either trying to redefine `add_library`, or modify all the vendored dependencies (they live in this organization as forks: https://github.com/chemfiles/fmt, ...). This is somehow documented here: https://github.com/chemfiles/chemfiles/blob/master/external/README.md

Luthaf:
This should be covered by #491 now

PlathC:
Just tested the following CMakeLists with your branch `update-tests`, and I do confirm that the issue seems to be fixed.

```
cmake_minimum_required(VERSION 3.21)
project(TestChemfiles)

add_subdirectory(extern/fmt)
add_subdirectory(extern/chemfiles)

add_executable(TestChemfiles main.cpp)
target_compile_features(TestChemfiles PRIVATE cxx_std_17)
target_link_libraries(TestChemfiles PRIVATE fmt chemfiles)
```

Thanks a lot for your work !

Issue #476, opened 2023-05-04
[feature request] Feasibility of reading angle dihedral and improper


Roy-Kid:
I would like to discuss the feasibility of reading angle, dihedral and improper from files(especially LAMMPS) in this issue. Currently, there is no straightforward way to read it. I have found that the Topology of chemfiles already has data structures for angle dihedral and improper, but no implementation and parser in LAMMPSData. However, adding this feature would be extremely beneficial for me when working with OPLS-AA or other all-atom forcefields.

I want to ask if is there a plan to add support for parse angle dihedral and improper. I can do this work in the next month or two.

Luthaf:
So currently chemfiles guesses angles/dihedrals/impropers from the list of bonds, and do not have a way to set these to arbitrary values. You can get the guessed list from the [`Topology`](https://chemfiles.org/chemfiles/latest/classes/topology.html#_CPPv4NK9chemfiles8Topology6anglesEv).

Do some LAMMPS file have list of angles/dihedrals/impropers that do not match the list of bonds? Could you share an example of such file?

Luthaf:
Right, so you are taking about angle/... **type**, not the indexes of the atoms inside the angle/.... 

That's not handled by chemfiles in any way currently; instead when writing LAMMPS file, we guess the angle types from the types of the atoms, so all H-O-H angles will have the same type (for example H-H-H would be 1, H-O-H would be 2, …). LAMMPS also recently added support for type labels, i.e. using strings instead of numbers. This is something that I'd like to support in chemfiles (#462). Would it help you if the output of chemfiles looked like `1 H-O-H 1 2 3` instead of `1 3 1 2 3`?

Adding the types inside chemfiles core data structures would require some large scale changes: tracking this information, determining a default when it is not present in the data format, exposing the data back to the user, but if this is the only solution we can think of what would be the best way to do it.

Roy-Kid:
Yes. Mapping the type from its original presentation to a specific format is tedious work. What if we designate some keywords by using `PropertyMap`. Let's say we define a `bond_properties = property_map()`, and we can read `bond_type` as an `int` or a `std::string` and store it. When users need the type info, they and check the `list_properties` and find out what key is. When write the data to file, users must manually use the keyword and set the type. Namely, we do not unify the representation type internally, but let users control it themselves.

Roy-Kid:
I need both **extract** angle/... types from LAMMPS file and write to new files. 

Sorry again for my vague words

[...]

Issue #462, opened 2022-09-12
Add support for symbolic Atom types in LAMMPS 


Luthaf:
LAMMPS very recently (https://github.com/lammps/lammps/issues/2002) added support to use strings as atomic types instead of numbers. This adds new optional sections in the LAMMPS data files, containing these symbolic types. We should add support for these in the parser!

PR #478, opened 2023-06-10
A draft for read and write bond type for lammps data

Roy-Kid:
Dear @Luthaf  and other developers,

According to the previous discussion #476 , I write an untested draft that read and write bond type from lammps data file. In this PR, I want to read and write bond/angle/dihedral data from lammps data and trajectory.

Since issue #462 is planning to support symbolic atom type, I choose std::string as the representation for bond type. 

In the following code, I completely complied with the storage method of BondOrder. But here I want to initiate a discussion about whether to use `bond type` or even `bond order` as a member attribute of `Bond`, like `type` in `Atom`. I think the benefits are as follows: The form is tidier, and the bond information can be obtained directly by iterating `topology.bonds`

I hope you can review my code in your free time so that I can complete these features in the future~ 

Thanks a lot!

### DIFF
--- a/include/chemfiles/Connectivity.hpp
+++ b/include/chemfiles/Connectivity.hpp
@@ -317,7 +317,7 @@ public:
     const sorted_set<Improper>& impropers() const;
 
     /// Add a bond between the atoms `i` and `j`
-    void add_bond(size_t i, size_t j, Bond::BondOrder bond_order = Bond::UNKNOWN);
+    void add_bond(size_t i, size_t j, Bond::BondOrder bond_order = Bond::UNKNOWN, std::string);
 
     /// Remove any bond between the atoms `i` and `j`
     void remove_bond(size_t i, size_t j);
@@ -330,6 +330,9 @@ public:
 
     /// Get the bond order of the bond between i and j
     Bond::BondOrder bond_order(size_t i, size_t j) const;
+
+    /// Get the bond type of the bond between i and j
+    std::string bond_type(size_t i, size_t j) const;
 private:
     /// Recalculate the angles and the dihedrals from the bond list
     void recalculate() const;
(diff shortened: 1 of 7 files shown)


Luthaf (commented):
Sorry for the very long delay for the review. This looks pretty good overall! 

My main concern is about the handling of bond types in the LAMMPS code. It looks like currently the code will create a set containing **all** bonds, where we would like to only have one entry per bond type (which should be a much smaller number).

I think the code should try to look for bond types in the topology, and if it can not find one, fallback to the old bond type generation.

---- 

The new bond type parameter to `add_bond` should also be added to the C API, to make it accessible from Python/Rust/Julia/...

Roy-Kid (commented):
Accept all suggestions and revise!

Roy-Kid:
@Luthaf Hi, sorry for bothering you. 
I re-wrote the code according to your suggestion. Now if it not find a custom bond type, it will degrade to integer bond type.
However, the custom bond type to integer bond type is one-to-one mapping, but the add_bond API is per-bond type mapping. For example, if 1-2 and 1-3 both bond type `1`, after I do `add_bond(1, 2, order, "typeA")` and `add_bond(1, 3, order, "typeB")`, the bond type in lammpsdata is both `typeB`. The later bond_type will override the former one. 
Should we use the rule: check if user defines a per-bond type, for example, add_bond(1, 2, order, "typeA"). If so, use "typeA" instead of the original `type 1`, if not, degrade it to `type 1`;

Luthaf:
> However, the custom bond type to integer bond type is one-to-one mapping, but the add_bond API is per-bond type mapping. For example, if 1-2 and 1-3 both bond type 1, after I do add_bond(1, 2, order, "typeA") and add_bond(1, 3, order, "typeB"), the bond type in lammpsdata is both typeB.

I'm not sure I understand here. `add_bond` is the chemfiles function, right? In this case, why would `add_bond(1, 2, order, "typeA")` followed by `add_bond(1, 3, order, "typeB")` result in both bonds having `typeB`? I would expect that this results in two bonds (1-2, with `typeA` and 1-3 with `typeB`). I'll re-read through the code as well to try to understand.

PR #480, opened 2023-07-29
Update to (more) recent CMake and C++ standard

Luthaf:


Luthaf:
This will be part of the 0.11 release, and not be backported to 0.10

PR #482, opened 2023-08-16
Faster XTC decoding

ezavod:
This PR adds faster decoding of integers which are stored as <= 64 bits. For these values, integer arithmetic can be used as proposed [here](https://doi.org/10.1186/s13104-021-05536-5) and implemented by [libxtc](https://gitlab.com/impulse_md/libxtc).

libxtc also has a special implementation for larger ints using 128 bit ints. I found that these lengths are rather unusual in "typical" MD trajectories.
Instead, the old algorithm is used to decode those values here eliminating special compile time checks for the presence of 128 bit types.

Also, a testcase for these large ints is added to test the old code paths.

The changes reduce the time to read the `ubuquitin.xtc` test file by more than 20%.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - DCD files are now read/written with a custom parser (#453)
 - Support TPR files up to version 2023 without a warning. Files from future
   versions are tried to be read but emit a warning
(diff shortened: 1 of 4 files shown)


Luthaf (approved):
This looks good! Could you also add this to the CHANGELOG?

Luthaf:
merged! thanks for doing this!

PR #484, opened 2023-09-07
mmtf optimization issue

sacraiou:
Writing a mmtf file of a big molecule is taking a lot of time abnormally.

On line 593 of src/formats/MMTF.cpp there is the line :
`auto positions = frame.positions();`

This affectation make a copy of the whole position vector instead of just got a reference.
Because this affectation is called for every residue, it dramatically slow down the writing process.

Changing it to
`const auto& positions = frame.positions();`

will use a reference assignment and correct the optimization issue.

### DIFF
--- a/src/formats/MMTF.cpp
+++ b/src/formats/MMTF.cpp
@@ -590,7 +590,7 @@ void MMTFFormat::add_residue_to_structure(const Frame& frame, const Residue& res
     group.atomNameList.reserve(residue.size());
     group.elementList.reserve(residue.size());
 
-    auto positions = frame.positions();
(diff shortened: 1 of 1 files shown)


Luthaf (approved):

Luthaf:
Good catch! Thanks for the PR. I'll merge once CI is happy.

Luthaf:
Let me know if you need this in a release! As far as I know, you are building VTX using your own fork of chemfiles, right?

sacraiou:
Indeed, we already made this modification in our fork, so we don't need a special release for it, thank you!

Issue #500, opened 2025-01-28
Amber NetCDF reader does not read `time`


Ladme:
Chemfiles does not provide the `time` information as an optional Frame property when reading Amber NetCDF trajectories.

But Amber NetCDF [can contain information about simulation time](https://ambermd.org/netcdf/nctraj.xhtml), although the presence of this property is optional:

```
6.2 Data variables
All data variables are optional. Some data variables have dependencies on other data variables, as described below. Creators shall define a units attribute of type character string for each variable as described below. Creators may define a scale_factor attribute of type float for each variable. Creators shall ensure that the units of data values, after being multiplied by the value of scale_factor (if it exists) are equal to that described by the units attribute. If a scale_factor attribute exists for a variable, readers shall multiply data values by the value of the scale_factor attribute before interpreting the data. This scaling burden is placed on the reader rather than the creator, as writing data is expected to be a more time-sensitive operation than reading it.
For trajectories, it is left as an implementation detail whether creators create a separate file for each variable grouping (e.g. coordinates and velocities) or a single file containing all variables. Some creators may allow the user to select the approach. Readers should support reading both styles, that is, combining data from multiple files or reading it all from a single file.
For restarts, any variable present should be written to one file.

time

Trajectory:
    float time(frame) units = “picosecond”
Restart:
    double time units = ”picosecond”
```

Other libraries, for instance, `mdtraj` or [`MDAnalysis`](https://docs.mdanalysis.org/2.8.0/documentation_pages/coordinates/TRJ.html) are able to read simulation time from NetCDF. It would be very nice to have the option to also read it with chemfiles.

Luthaf:
Yes, this makes sense. This should be set as a property on the frame. 

I only have limited time to work on chemfiles these days, so the fastest way to get the feature would likely be to contribute it yourself.

Ladme:
Understood. (#501)

Luthaf:
Fixed by #501

PR #501, opened 2025-01-28
Support for reading and writing simulation time in Amber Restart and Amber NetCFD trajectories

Ladme:
1. I’ve added support for reading simulation time from both Amber Restart and Amber NetCDF trajectories. I also added some checks for this in the existing tests. I also did some manual testing with a handful of other `.nc` files to confirm the reported times are accurate - though it wasn’t particularly exhaustive. Reading NetCDF trajectories without time info still works as expected too.

2. I’ve also added the ability to write simulation time for both Amber Restart and Amber NetCDF trajectories. For trajectories, time writing is only initialized if the first frame written includes time info. If a later frame includes time info but time writing wasn’t initialized, a warning is raised, and the time isn’t written. If time writing *is* initialized and a frame doesn’t have time info, nothing happens. I ran some manual tests to ensure the files generated by chemfiles can be read by VMD, and everything looks good. The files are also parsable by `netCDF4` and the time info is accessible. The writer assumes that frame time is in picoseconds. I’ve also added checks for time handling in the write tests.

3. I've also updated the documentation in `frame.toml`.

4. Finally, I've tested `chemfiles.rs` compiled with this new version of chemfiles and everything seems to be working fine as well there.

This addresses [Issue #500](https://github.com/chemfiles/chemfiles/issues/500).

I don’t have much experience with C++, so a thorough code review might be useful.

### DIFF
--- a/src/formats/AmberNetCDF.cpp
+++ b/src/formats/AmberNetCDF.cpp
@@ -288,7 +288,7 @@ void AmberNetCDFBase::write(const Frame& frame) {
                 throw format_error("invalid type for time variable");
             }
         } else {
-            warning("AMBER NetCDF", "this file does not contain space for time, it will not be saved");
+            warning("AMBER NetCDF", "this file does not have a 'time' attribute, it will not be saved");
         }
     }
 
--- a/src/formats/AmberNetCDF.cpp
+++ b/src/formats/AmberNetCDF.cpp
@@ -34,6 +34,7 @@ static size_t get_dimension_id(const netcdf3::Netcdf3Builder& builder, const std
 
 static double scale_for_distance(std::string unit);
 static double scale_for_velocity(std::string unit);
+static double scale_for_time(std::string unit);
 
 /******************************************************************************/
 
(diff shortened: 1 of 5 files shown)


src/formats/AmberNetCDF.cpp
```diff
@@ -245,6 +273,25 @@ void AmberNetCDFBase::write(const Frame& frame) {
...
+                variables_.time.var->write(step_, &time_f32, 1);
+            } else if (variables_.time.var->type() == netcdf3::constants::NC_DOUBLE) {
+                // Amber Restart
+                double time_f64 = *time_opt;
+                variables_.time.var->write(step_, &time_f64, 1);
+            } else {
+                throw format_error("invalid type for time variable");
+            }
+        } else {
+            warning("AMBER NetCDF", "this file does not contain space for time, it will not be saved");
```
Luthaf, line 291:
```suggestion
            warning("AMBER NetCDF", "this file does not have a 'time' attribute, it will not be saved");
```

Luthaf (commented):
This looks good to me, I have one small suggestion to improve the warning wording.

CI failure is unrelated, I'll update it separately ASAP

src/formats/AmberNetCDF.cpp
Ladme: Yes, that's better.

Luthaf (approved):

Luthaf:
Thanks a lot! I think I'll release this as v0.10.5, but it might also take some time

PR #506, opened 2025-03-08
Remove references to 'step' in `Frame` and `Trajectory` 

Luthaf:
The old frame `step` is now stored as a property called `"index"`, and the corresponding trajectory functions are `Trajectory::size` and `Trajectory::read_at`. When available, the simulation step is stored in another frame property, called `"simulation_step"`.

Fix https://github.com/chemfiles/Chemfiles.jl/issues/81. @lmiq, does the description above make sense to you? it should be a direct translation to the Julia API. I will keep the old function around in Julia/Python with a deprecation warning, for a couple of releases.


lmiq:
This seems nice. Is there a ˋindex(::Frame)ˋ function to retrieve the index of the frame from the frame object? This is useful in some situations.

Luthaf:
> Is there a ˋindex(::Frame)ˋ function to retrieve the index of the frame from the frame object? This is useful in some situations.

Not directly, but you can use `property(frame, "index")` to get this information. 

We could also define a `index(::Frame)` function that does that (and is type stable) as well, not sure if this will be used often enough to warrant a new function?

lmiq:
It is useful to compute time-dependent properties in parallel, where each task computes the property for a different frame and updates a vector of lengh(trajectory) at the appropriate position. Of course the frame index can be carried around, but if that is supported out of the box it makes the code slightly cleaner.

Luthaf:
I was not very clear here =) The data is carried around inside the Frame, as a `Property`; and you can always access it with `property(frame, "index")` in Julia. 

I was wondering if it was worth also having a convenience function `index(frame)` to do the same thing in fewer characters

lmiq:
Yes, I got that. I understood that we can already have it with `property`. That's ok, while in Julia having the possibility of restricting type-instabilities to a minimum is usually important, particularly for debugging. I mean, given the small cost of implementing that (I guess), I would probably do it. 

Another advantage of having getter functions for all properties is the possibility of easily developing software that uses Chemfiles or other backend and swithching only by defining dispatch of the appropriate structs.

Luthaf:
Ok, so in the end I kept the existing infrastructure for `step`, and renamed it to `index`. The simulation_step is still handled separately.

PR #510, opened 2025-04-17
Support UTF-8 paths on Windows (allow for non-ASCII characters in paths)

DmitriyMarin:
These changes make it possible to load files with special UTF-8 characters (e.g., é, à, ö) in their paths on Windows.

The tests pass.

### DIFF
--- a/src/files/BinaryFile.cpp
+++ b/src/files/BinaryFile.cpp
@@ -78,10 +78,10 @@ BinaryFile::BinaryFile(std::string path, File::Mode mode):
 #ifdef _WIN32
     // On Windows, allow for UTF-8 paths containing non-ASCII characters
 
-	// convert to a wide string (UTF-8) to take care of special characters
-	const int size_needed = MultiByteToWideChar(CP_UTF8, 0, &this->path()[0], (int)this->path().size(), NULL, 0);
-	std::wstring w_path = std::wstring(size_needed, 0);
-	MultiByteToWideChar(CP_UTF8, 0, &this->path()[0], (int)this->path().size(), &w_path[0], size_needed);
+    // convert to a wide string (UTF-8) to take care of special characters
+    const int size_needed = MultiByteToWideChar(CP_UTF8, 0, &this->path()[0], (int)this->path().size(), NULL, 0);
+    std::wstring w_path = std::wstring(size_needed, 0);
+    MultiByteToWideChar(CP_UTF8, 0, &this->path()[0], (int)this->path().size(), &w_path[0], size_needed);
 
     auto file_descriptor = _wopen(w_path.c_str(), open_mode, permissions);
 #else
--- a/src/files/BinaryFile.cpp
+++ b/src/files/BinaryFile.cpp
@@ -22,6 +22,11 @@
 
 #ifdef CHEMFILES_WINDOWS
 #include <io.h>
+
+#ifdef _WIN32
+#include <windows.h>
+#endif
+
 #endif
 
 #ifdef __CYGWIN__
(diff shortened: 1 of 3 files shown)


src/files/BinaryFile.cpp
```diff
@@ -70,7 +75,18 @@ BinaryFile::BinaryFile(std::string path, File::Mode mode):
     int permissions = S_IRWXU | S_IRWXG | S_IROTH;
 #endif
 
+#ifdef _WIN32
```
Luthaf, line 78: Any reason for using `_WIN32` instead of `CHEMFILES_WINDOWS`? The second one is defined by cmake and should cover the same kind of platforms.

src/files/BinaryFile.cpp
```diff
@@ -22,6 +22,11 @@
 
 #ifdef CHEMFILES_WINDOWS
 #include <io.h>
+
+#ifdef _WIN32
+#include <windows.h>
```
Luthaf, line 27: Do we need all of `windows.h` for path conversion functions? MSDN indicates `stringapiset.h` as a relevant header.

I find `windows.h` to be very large and painful on the compilation time

tests/lints/check-public-headers.py
```diff
@@ -35,6 +35,8 @@
     "new",
     "cstddef",
     "map",
+    # windows headers used for converting to a wide string
+    "windows.h",
```
Luthaf, line 39: I don't understand why this is needed? You only included this in `.cpp` files, and this script should only check chemfiles' headers

Luthaf (commented):
Thanks a lot for your contribution! It looks good overall, and I have a couple of question about why this is written the way it is

tests/lints/check-public-headers.py
DmitriyMarin: Indeed, I will remove it. I haven't checked and thought that it would also check the source files for allowed includes.

src/files/BinaryFile.cpp
DmitriyMarin: No particular reason. I will modify it to use CHEMFILES_WINDOWS.

src/files/BinaryFile.cpp
DmitriyMarin: `windows.h` is mainly required due to the usage of `CP_UTF8`.

src/files/PlainFile.cpp
```diff
@@ -36,6 +42,32 @@ PlainFile::PlainFile(const std::string& path, File::Mode mode): TextFileImpl(pat
...
+        break;
+    case File::WRITE:
+        openmode = L"wb";
+        break;
+    default:
+        unreachable();
+    }
+
+    // convert to a wide string (UTF-8) to take care of special characters
+    const int size_needed = MultiByteToWideChar(CP_UTF8, 0, &path[0], (int)path.size(), NULL, 0);
```
Luthaf, line 65: So one last question I have about this: would this code fail if the user system is set to use a different codepoint than UTF-8? And how frequent would this kind of system be nowadays?

DmitriyMarin: Great question!

Yes, it might fail in this case, but I would not expect it to be a frequent case. It would depend on how std::strings are stored in the system, and what the user has supplied as the path. I tested it locally, and I can open files with special characters in their names (I have en-US locale).

I haven't noticed specifying `UNICODE` or `_UNICODE` in the `CMakeLists.txt` - maybe this should be done as well?

Without this conversion into a wide string, it currently fails if the path contains some non-ASCII characters, whether in UTF-16/UTF-8 or others.

The better way would be to allow for builds with C++17 and, in case of the C++17-enabled build, use `std::filesystem::u8path`.

We can also modify the conversion with the use of `MultiByteToWideChar` to try to convert from UTF-8 into UTF-16, and, if it fails, then try to convert from ANSI/ACP:

```c++
  std::wstring w_path;
  // Try UTF-8
  int len = MultiByteToWideChar(CP_UTF8, MB_ERR_INVALID_CHARS, path.c_str(), -1, NULL, 0);
  if (len != 0) {
      w_path.resize(len);
      MultiByteToWideChar(CP_UTF8, 0, path.c_str(), -1, &w_path[0], len);
  } else {
      // Fallback to ANSI
      len = MultiByteToWideChar(CP_ACP, 0, path.c_str(), -1, NULL, 0);
      w_path.resize(len);
      MultiByteToWideChar(CP_ACP, 0, path.c_str(), -1, &w_path[0], len);
  }
```

The code above is just an example in which it is assumed that `path` is null-terminated and, hence, passing `-1`  instead of `(int)path.size()`. Note: I haven't tested this code.

Another option would be to use the following functionality that is deprecated in C++17 (but not yet removed!):
```c++
// Convert std::string to wide string
const std::wstring w_text = std::wstring_convert<std::codecvt_utf8<wchar_t>>().from_bytes(text);
```

Note: I am no expert in the encodings on Windows, so I might be missing something.

Luthaf:
Hey! Sorry for the long delay, work is a bit crazy right now. 

I just realized that we actually require C++17 (https://github.com/chemfiles/chemfiles/blob/c4553d2db739a8ba0c0094510db0be475d8242d8/CMakeLists.txt#L90) on master, and the next release will as well.

Let me know If you have time to update this PR using the C++17 tools, which from what I understand should be a lot nicer than the current hacks! Otherwise I'll have a go myself as soon as I can.

DmitriyMarin:
Hi! Great! I will update today using the C++17 approach then. I haven't noticed this and somehow thought there was a limitation with the usage of C++ not higher than C++11.

Luthaf (approved):
Thanks a lot!

PR #511, opened 2025-05-13
Read and write forces in Gromacs TRR files

### LINKED ISSUE #496: TRR parser does not read forces
The GROMACS TRR file format allows to save the coordinates, velocities, forces and energies (https://manual.gromacs.org/current/reference-manual/file-formats.html#trr). But the chemfiles TRR parser does not read the data of the forces or energies.

In `include/chemfiles/formats/TRR.hpp` the property `f_size` for the forces is created but forces are never read. There also does not exist a flag whether the forces exist.

ezavod:
Fixes #496.

This reads forces from a TRR trajectory, converts them from `kJ / (mol * nm)` to `kJ / (mol * Angstrom)`, and stores them in the atomic property `force`.
When writing, the conversion is done in reverse. For atoms without the atomic property set, a default value of zero force is written (should there be an error / warning?).

Additionally, some smaller changes:
- update Readme to reflect changes from #480.
- fix comparison in an C example.
- run `clang-format` and ICWYU on the TRR related files.


doc/src/properties/atom.toml
```diff
@@ -77,3 +77,8 @@ type = "string"
 TPR = """Describes the atom type specified in the ``[atoms]`` section of the
 topology. The type can be used to reference a specific set of force field
 parameters for a given atom."""
+
+[force]
+type = "vector"
+
+TRR = "Force acting on an atom. In units of 'kJ / (mol * nm)'."
```
Luthaf, line 84: If you do the conversion on input/output; should this be 
```suggestion
TRR = "Force acting on an atom. In units of '(kJ / mol) / A'."
```

src/formats/TRR.cpp
```diff
@@ -82,33 +87,88 @@ TRRFormat::TRRFormat(std::string path, File::Mode mode, File::Compression compre
...
+template <typename T>
+typename std::enable_if<std::is_same<T, float>::value || std::is_same<T, double>::value, void>::type
+read_real_vec(XDRFile& file, std::vector<T>& dx) {
+    if constexpr (sizeof(T) == 4) {
+        file.read_f32(dx);
+    } else {
+        static_assert(sizeof(T) == 8);
+        file.read_f64(dx);
+    }
+}
```
Luthaf, line 114: I think I prefer template specialization for this kind of things:

```suggestion
template <typename T>
void read_real_vec(XDRFile& file, std::vector<T>& dx) {
    throw std::runtime_error("read_real_vec can only be used with float or double")
}

template <>
void read_real_vec<double>(XDRFile& file, std::vector<double>& dx) {
    file.read_f64(dx);
}

template <>
void read_real_vec<float>(XDRFile& file, std::vector<float>& dx) {
    file.read_f32(dx);
}
```

Luthaf (commented):
Thanks a lot!

doc/src/properties/atom.toml
ezavod: I should have learned to never copy and paste units :-(

src/formats/TRR.cpp
ezavod: Thanks for the suggestion.
I've made it a compile-time error instead of a runtime error by using `static_assert` instead of exceptions.

Luthaf: good idea!

Luthaf (approved):

PR #513, opened 2025-05-14
Fix TPR reader for v2025

### LINKED ISSUE #512: Add support for v2025 TPR files
Gromacs version v2025 still has the same TPR generation, i.e. the topology format should have been backwards compatible and readable by chemfiles. Unfortunately, there are [changes in Gromacs](https://gitlab.com/gromacs/gromacs/-/merge_requests/4856) which missed to bump the generation but still change the interaction serialization and brake the current reader implementation.

The TPR reader needs to be adapted to the v2025 version in order to avoid unexpected crashes.

ezavod:
Fixes #512.


Luthaf (approved):
Thanks!

PR #515, opened 2025-05-16
add velocity support to extxyz file format

### LINKED ISSUE #514: Add velocities parsing to extxyz file format
I noticed that currently velocities are not parsed in the extxyz file format. Would it be possible to add parsing for velocities? 

 It seems possible to add this property to each `Atom` in a `Frame`,  which when written out gives a extxyz file with velocities in it. Then to read them back you need to pull the `velo` (typically velocities are `velo:R:3`) property for each `Atom` to get the velocities. 

Would be much nicer to have `frame.velocities()` work for extxyz file formats like they do for other formats. Is it not easy to parse for properties associated with `velo:R:3` and store them as the velocities? Or was this omitted on purpose?

A possible solution would be to add a check in `parse_property_list` for a property named `velo` then store the associated vector as velocity similar to how `pos` is done. The only issue I see (which is why I ask if omitting was purposeful) is the velocities metadata flag being set to true would cause issues when `velo` is not included. This could be mediated by setting the velocity to `0` when `velo` is not found.

Cavenfish:
I've added support for parsing `velo` into atom velocities, which lets users pull the velocities from extxyz files through `frame.velocities()`.  Similarly, if the frame has velocities set then writing will naturally include them in extxyz as `velo:R:3`.

This fixes #514.

I confirmed that all tests passed, but since I rarely work with c++ the code style/format may not be great. Hopefully its still somewhat useful.

### DIFF
--- a/src/formats/XYZ.cpp
+++ b/src/formats/XYZ.cpp
@@ -75,7 +75,7 @@ template<> const FormatMetadata& chemfiles::format_metadata<XYZFormat>() {
     metadata.memory = true;
 
     metadata.positions = true;
-    metadata.velocities = false;
+    metadata.velocities = true;
     metadata.unit_cell = true;
     metadata.atoms = true;
     metadata.bonds = false;
@@ -98,12 +98,22 @@ void XYZFormat::read_next(Frame& frame) {
         auto count = scan(line, name, x, y, z);
         auto atom = Atom(std::move(name));
         read_atomic_properties(properties, line.substr(count), atom);
-        frame.add_atom(std::move(atom), Vector3D(x, y, z));
+
+        // Check for velocity 
+        if (auto velo = atom.get("velo")) {
+            auto v = velo.value().as_vector3d();
+
+            frame.add_atom(std::move(atom), Vector3D(x, y, z), v);
+        } else {
+            frame.add_atom(std::move(atom), Vector3D(x, y, z));
+        }
+        
     }
 }
 
 void XYZFormat::write_next(const Frame& frame) {
     const auto& positions = frame.positions();
+    auto velocities = frame.velocities();
     auto properties = get_atom_properties(frame);
 
     file_.print("{}\n", frame.size());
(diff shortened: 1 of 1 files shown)


src/formats/XYZ.cpp
```diff
@@ -98,12 +98,22 @@ void XYZFormat::read_next(Frame& frame) {
         auto count = scan(line, name, x, y, z);
         auto atom = Atom(std::move(name));
         read_atomic_properties(properties, line.substr(count), atom);
-        frame.add_atom(std::move(atom), Vector3D(x, y, z));
+
+        // Check for velocity 
+        if (auto velo = atom.get("velo")) {
```
Luthaf, line 103: I would rather not store a `velo` property on atoms in this case, but rather store a `frame_has_velocities_` flag in this class, which is set by `read_extended_comment_line`; and used here to change the `scan(line, name, x, y, z)`. 

Although I'm not sure, do we have a guarantee that the `velo` field always comes after `pos`, or could it be anywhere?

Luthaf:
It would be nice to add a test for an xyz file with velocities. Do you have such a file around? If so, can you share it with me, I'll add it to https://github.com/chemfiles/tests-data/, which is then pulled by the tests on this repo.

src/formats/XYZ.cpp
Cavenfish: Typically, it comes after `pos` but in reality it can be anywhere. I tried to leave this free by pulling `velo` from the properties reading function. I do see now that storing `velo` as a property could cause the writing to print the `velo` property twice.

Cavenfish:
> It would be nice to add a test for an xyz file with velocities. Do you have such a file around? If so, can you share it with me, I'll add it to https://github.com/chemfiles/tests-data/, which is then pulled by the tests on this repo.

Does this simple example work? If not I can email you a more robust xyz file.
```
3
Properties=species:S:1:pos:R:3:velo:R:3 Lattice="10 0 0 0 10 0 0 0 10"
H 1 0 0 1 0 0
O 0 0 0 1 0 0
H 0 1 0 1 0 0
```

Cavenfish:
Ok, I've revised the code such that the velocities are not stored in `properties`. This came with a constraint that even when `velo` is not in the extxyz file the velocities are initialized as zero. I wasn't sure how else to allow `velo` to appear anywhere but also still parse it without storing it in properties. In the end this seems (and hopefully is) fine though.

Let me know what you think!

Luthaf:
Tanks a lot @Cavenfish, and sorry for the long delay here

PR #525, opened 2025-11-19
BCIF new format support

ValentinGuillaume:
The goal of this PR is to add our new bcif read/write feature to the main chemfiles repository.

Bcif file format is the new standard for downloading protein structure from the rcsb pdb since july 2024 and we would like to support it. @sguionni wrote an issue a few weeks ago regarding this specific subject 

The implementation should follow the pattern that is being used in the repository and doesn't need any new dependencies. It is tested by the read, write, reread method on a few heterogeneous files found on the rcsb pdb. The files are subject to their own PR on the associated repository.


Luthaf:
Thanks a lot for this PR! I'm bit underwater right now, but I should be able to review by the end of next week.

In the mean time I can restart CI multiple time, until is passes!

ValentinGuillaume:
No worries, we shall take our time with this ;)
In the meantime, I needed to have the test-data repo merged to be fully ready for this PR. Now that you approved it, I updated the hashes from the tests/CMakelists.txt to point toward the right commit and now it should work.

ValentinGuillaume:
I resolved some issues regarding the checksum, file copy path and error message format enforcement.
Can you restart the CI ?

ValentinGuillaume:
I added support for a table that connect some entities together sometimes. Also I removed the ui64 notation that caused you CI to fail compilation. 
Can you restart the CI ?

ValentinGuillaume:
The coverage issue is a big blow for me. The lines that need to be hit for the 83% to be reached are mostly failsafe lines that should occurs when reading a ill-formed or extremely huge bcif file. I have not such file at hand and thus increasing the coverage demands from me to craft those files, which is not as easy as crafting a cif or pdb or whatever human readable format. 
I will need way more time to do it right so for the time being, I will put this PR on hold. 
Do you have a doc/tutorial somewhere on how I can run this coverage test by myself before submitting this PR back ?

Luthaf:
Do not worry too much about the coverage for now, I can also lower the threshold. If you have some time it would be nice to if there are parts of your code are not tested and also not failsafe lines, and add tests for these. Error cases for invalid files are fine to leave as is for now.

ValentinGuillaume:
I made some read speed improvements and tried to improve test coverage. Can you run the CI again ?

Luthaf:
The test faillure makes it look like the code is not pulling the most recent version of the test data

ValentinGuillaume:
I updated the hashes to point toward the last commit that should contains all the needed files.
Can you restart the CI ?

ValentinGuillaume:
I made some change in the architecture in order to increase the code coverage value. Lets see if that's enough to pass that 83% limit

ValentinGuillaume:
Ok I added 8 easy lines of code coverage and we should be ready to merge.

Luthaf:
This is quite a lot of code, so I'll do a review ASAP, hopefully by next week. 

One question that jumps to mind after a first very high level look at the code: why do you have separate `BCIF` and `BCIF_impl` files? Just to be able to test the internal implementation?

Also, if you set any atom/residue/frame properties, please document them here: https://github.com/chemfiles/chemfiles/tree/master/doc/src/properties

ValentinGuillaume:
> Just to be able to test the internal implementation?

Well, yes, partly. I needed to unit test them to avoid forging binary data files and filling the chemfiles-test-data repo. And it was functions that are designed to work with the BCIF reading/writing code specifically, so I didn't want to expose them in the BCIF header because it would send the wrong message. Furthermore, I figured that BCIF.cpp was quite large too. 
All of that made me think it was cleaner to add this file. 
But feel free to move it to another location, rename it, or whatever suits your taste :) 

Regarding the tolm files and the properties : I followed the mmCIF guidelines as the bcif is designed as a binary counterpart of mmcif and thus have a lot in common with it. Did I miss something ?

Luthaf:
> Regarding the tolm files and the properties : I followed the mmCIF guidelines as the bcif is designed as a binary counterpart of mmcif and thus have a lot in common with it. Did I miss something ?

These files are documenting what properties are set/used when reading/writing different format. I guess you can just add a copy of the `mmCIF` line with `BCIF` instead! I'll let you check that this includes everything relevant

ValentinGuillaume:
I couldn't get the doc to work locally as I'm working on windows and the default text-file `open()` method from python uses os encoding, which made the angtrom symbol fail. 
Since the encoding will always be utf-8 I figured it was an appreciable fix, plus I could test it locally thanks to this modification.
Now the doc-build should work.

src/formats/BCIF_impl.cpp
```diff
@@ -0,0 +1,318 @@
...
+void get_int_type_str(const int32_t& type_code, std::string& type_str) {
+    if (type_code == 1) type_str = "Int8";
+    else if (type_code == 2) type_str = "Int16";
+    else if (type_code == 4) type_str = "Int32";
+    else if (type_code == 5) type_str = "Uint8";
+    else if (type_code == 6) type_str = "Uint16";
+    else if (type_code == 8) type_str = "Uint32";
+    else if (type_code == 32) type_str = "Float32";
+    else if (type_code == 33) type_str = "Float64";
+    else type_str = "Int32";  // fallback
```
Luthaf: Why does this need a fallback? Would it be better to throw an error on unknown type code?

src/formats/BCIF.cpp
```diff
@@ -0,0 +1,3767 @@
...
+    struct hash<std::pair<size_t, size_t>> {
+        size_t operator()(const std::pair<size_t, size_t>& p) const {
+            auto h1 = std::hash<size_t>{}(p.first);
+            auto h2 = std::hash<size_t>{}(p.second);
+            // Combine hash values using a simple approach
+            return h1 ^ (h2 << 1);
+        }
+    };
+}
+namespace chemfiles {
```
Luthaf: Why is this in the chemfiles namespace? If it is only a utility class it can be defined inline in this file.

src/formats/BCIF.cpp
```diff
@@ -0,0 +1,3767 @@
...
+    class StringCollection
```
Luthaf, line 98: Could you add some documentation about what this class is and what it does?

src/formats/BCIF.cpp
```diff
@@ -0,0 +1,3767 @@
...
+                    conn.ptnr1.label_comp_id = (i < struct_data.ptnr1.auth_comp_id.size()) ? struct_data.ptnr1.auth_comp_id[i] : "";
+                    conn.ptnr1.label_seq_id = (i < struct_data.ptnr1.auth_seq_id.size()) ? struct_data.ptnr1.auth_seq_id[i] : -1;
+                    conn.ptnr1.label_atom_id = (i < struct_data.ptnr1.label_atom_id.size()) ? struct_data.ptnr1.label_atom_id[i] : "";
+                    conn.ptnr2.label_asym_id = (i < struct_data.ptnr2.auth_asym_id.size()) ? struct_data.ptnr2.auth_asym_id[i] : "";
+                    conn.ptnr2.label_comp_id = (i < struct_data.ptnr2.auth_comp_id.size()) ? struct_data.ptnr2.auth_comp_id[i] : "";
+                    conn.ptnr2.label_seq_id = (i < struct_data.ptnr2.auth_seq_id.size()) ? struct_data.ptnr2.auth_seq_id[i] : -1;
+                    conn.ptnr2.label_atom_id = (i < struct_data.ptnr2.label_atom_id.size()) ? struct_data.ptnr2.label_atom_id[i] : "";
+                    conn.pdbx_value_order = (i < struct_data.pdbx_value_order.size()) ? struct_data.pdbx_value_order[i] : "";
+                    // WIP
```
Luthaf: Why WIP?

src/formats/BCIF.cpp
```diff
@@ -0,0 +1,3767 @@
...
+
+            };
+            inline void generate_structconn(const StructConnData& struct_data, std::vector<BCIFFormat::BCIFData::StructConn>& out, BCIFFormat::BCIFData::StructConnMap& map)
+            {
+                // Build struct_conn entries
+                size_t num_conns = struct_data.conn_type_id.size();
+                out.reserve(num_conns);
+                for (size_t i = 0; i < num_conns; ++i) {
+                    out.push_back({});
+                    BCIFData::StructConn& conn = out.back();
```
Luthaf, line 1207: This is not very easy to read/understand. Could you add some comments that explain what's going on?

src/formats/BCIF.cpp
```diff
@@ -0,0 +1,3767 @@
...
+
+        std::vector<uint8_t> encode_byte_array_float64(const std::vector<double>& data) {
+            // Encode Float64 array to little-endian byte array
+            std::vector<uint8_t> result(data.size() * 8);
+
+            for (size_t i = 0; i < data.size(); ++i) {
+                uint64_t bits;
+                std::memcpy(&bits, &data[i], sizeof(double));
+
+                // Store as little-endian
```
Luthaf, line 2208: does BCIF specify endianess in any way?

src/formats/BCIF.cpp
```diff
@@ -0,0 +1,3767 @@
...
+        // Encode the frame to MessagePack binary
+        auto binary_data = encode(frame);
+
+        // Write to file
+        auto file = TextFile(filename_, File::WRITE, compression_);
+        file.print("{}", std::string(binary_data.begin(), binary_data.end()));
+
+        has_written_ = true;
+    }
+    void BCIFFormat::decode(const char* data // TODO: When upgrading to c++20, use a span instead.
```
Luthaf, line 3731: This function decodes everything and stores the data inside the class, right?

Luthaf (commented):
The code looks very good overall, I left a couple of questions below.

For most other formats in chemfiles, the code that is currently in BCIF_impl (and part of formats/BCIF.cpp) would be handled by a custom `BinaryFile` class, that deals with reading, unpacking/packing data, … Do you have some bandwidth to refactor this code in this way, splitting better the bits that are mostly IO related from the bits that parse an actual chemical structure?

src/formats/BCIF_impl.cpp
ValentinGuillaume: I'll have an educated answer that once I encounter this case-scenario, but for now either way is fine for me. If you want I can make it throw instead.

src/formats/BCIF.cpp
ValentinGuillaume: You right, maybe this is something cleanup-worthy that I overlooked. I'll change that.

src/formats/BCIF_impl.cpp
Luthaf: Yes I'd rather be safe and not try to guess here =)

src/formats/BCIF.cpp
ValentinGuillaume: At some point it was, but I didn't remove the comment, sorry for that.

src/formats/BCIF.cpp
ValentinGuillaume: "specify" is a big word, but from what I could understand from the reference implementation [here](https://github.com/molstar/BinaryCIF), bytes are always little-endian in bcif.

src/formats/BCIF.cpp
ValentinGuillaume: The decode method parse the input buffer and organize data into the BCIFData that is owned by the instance of BCIFFormat for later usage.

ValentinGuillaume:
> For most other formats in chemfiles, the code that is currently in BCIF_impl (and part of formats/BCIF.cpp) would be handled by a custom BinaryFile class, that deals with reading, unpacking/packing data, … Do you have some bandwidth to refactor this code in this way, splitting better the bits that are mostly IO related from the bits that parse an actual chemical structure?

Half of BCIF_impl free functions are unrelated to binary file io operations. So in the context of your question, please keep in mind that the class you propose won't fulfill all free functions responsibilities and won't erase the BCIF_impl.hpp entirely. 

The custom BinaryFile class raises two concerns on my end : 

1. The functions used to parse the bytes are tightly coupled with how the bcif format organizes the data. Sometimes it is rather simple and intrinsically generic (e.g. integers) sometimes it is very specific (e.g. strings). So my concern is the following : will I end up making a generic class that only decode very specific data ?
2. Do you have an usage of this potential BinaryFile class in mind ? If you do, it would help me to understand the relevance of this refactoring and maybe the responsibilities this class would have. I have trouble to figure what the class would do and why we need it to be a class rather than a set of free functions.

ValentinGuillaume:
Following your advices, I commented more and made the get_int_type_str function throw for unknown type. 
(I also removed the include span that I forgot to cleanup and caused gcc9 to fail)

Luthaf:
> The custom BinaryFile class raises two concerns on my end :
> 
>     1. The functions used to parse the bytes are tightly coupled with how the bcif format organizes the data. Sometimes it is rather simple and intrinsically generic (e.g. integers) sometimes it is very specific (e.g. strings). So my concern is the following : will I end up making a generic class that only decode very specific data ?
> 
>     2. Do you have an usage of this potential BinaryFile class in mind ? If you do, it would help me to understand the relevance of this refactoring and maybe the responsibilities this class would have. I have trouble to figure what the class would do and why we need it to be a class rather than a set of free functions.

This would not be a generic class, but a BCIF-specific class. For example, we have a [NetCDF3](https://github.com/chemfiles/chemfiles/blob/master/src/files/Netcdf3File.cpp) binary file implementation, that provides tools for reading/writing files in NetCDF3 format; and is then used by the [AmberNetCDF](https://github.com/chemfiles/chemfiles/blob/master/src/formats/AmberNetCDF.cpp) format to implement this format. 

The file class is only used together with one format class, it's more about separating concerns between "extracting data from a stream of bytes" and "interpreting this data as a chemical structure". 

Is this clearer or not really?

PR #532, opened 2026-06-10
DCD: convert DELTA between AKMA and picoseconds

traviswheeler:
DCD's DELTA is in AKMA time units (CHARMM/NAMD convention), but the reader was storing it directly into timesteps_.dt and propagating it to Frame.time, which every other chemfiles format treats as picoseconds. Cross-format conversions (e.g. DCD -> XTC) silently inflated per-frame times by 1 / 4.888821e-2 ~ 20.45x.

I recently submitted PR #531, which back-patches DCD timestep metadata and correctly preserves NPRIV/NSAVC/DELTA through a DCD round-trip -- but that round-trip is symmetric, so a missing unit conversion on read paired with the same missing conversion on write cancels out and the bug stayed hidden. It only surfaces when reading DCD and writing XTC/TRR/etc.

Convert at the I/O boundary: multiply by AKMA_TO_PS on read, divide on write, in both the CHARMM-format (f32) and X-PLOR-format (f64) branches. timesteps_.dt is now in picoseconds everywhere in memory.

This matches how MDAnalysis handles DCD time units (DCD is AKMA on disk, ps in memory); see package/MDAnalysis/coordinates/DCD.py lines 81-82, 128, 153-154.


Luthaf:
The CI failure seems relevant here!

traviswheeler:
That's embarrassing - all those pretty tests, then I failed to run them ...

This should resolve those issues.

Luthaf:
Could you also update the description of the time property here https://github.com/chemfiles/chemfiles/blob/master/doc/src/properties/frame.toml#L23, to update this webpage: https://chemfiles.org/chemfiles/latest/properties.html#frame-properties

traviswheeler:
This should address the documentation issue. 

I want to be sure that part of the commit message doesn't get lost:

While making the change I noticed that the entries in the
[simulation_step] table (lines 30-36) all look like copy-pastes of the
matching [time] entries: they describe time values, but it looks to me
like simulation_step is a dimensionless integration step index, not a
time in any unit. For DCD [simulation_step] is set in
src/formats/DCD.cpp:181-183 as NSAVC * index + NPRIV; the XTC and TRR
readers do the equivalent, with the source labeling it "actual step
of MD Simulation" (src/formats/XTC.cpp:116, src/formats/TRR.cpp:170).
I'm not changing it in this commit because it would be out of scope
from the DCD changes I've made ... but I believe those lines should
all be something like:
      DCD = "The integration step number of the frame."

traviswheeler:
Checking to see if these documentation changes meet your needs. (A reminder that we're using chemfiles within our mdcompress tool, and this bug is a blocker for correct compression of DCD files)

Luthaf:
The changes are good, I'll fix the mismatched description for step in a separate PR. Thanks for the fix!

